PR #13, opened 2023-02-28
use a disk-based cache

xloem:


whitead:
Thanks @xloem - could you change the cache location to `Path.home() / ".paperqa" / llm_cache.db`? Thanks

xloem:
fixed

whitead:
Thanks @xloem

PR #21, opened 2023-03-05
gpt-3.5-turbo, priced at $0.002 per 1k tokens

scott------:
The current formula for cost calculation is not correct for the current pricing for gpt-3.5-turbo so we need to add a zero here to correct the calculation from .02 to .002 per 1000 tokens.    https://openai.com/pricing


scott------:
Updating pricing to reflect current gpt-3.5-turbo token costs of .002 per 1000 tokens.

PR #27, opened 2023-03-10
Docs: Fix accidental relative URL to be absolute

mcbhenwood:
It's the smallest of changes but it may help somebody. Previously, clicking on the link yielded a 404 on github.com owing to the relative nature of the link.


whitead:
Thanks!

PR #29, opened 2023-03-11
Fixing the format of (Citation1, Citation2, Citation3)

jan-kubica:
This adds support for recognising the citations in the answer in the format of 

(Citation1, Citation2, Citation3)

Currently, only Citation3 would be recognised in this citation.


whitead:
Thanks!

PR #35, opened 2023-03-26
Fix a reader issue

wpv-chan:
Prevent Mac users from reading the .DS_Store file in the directory.


whitead:
@wpv-chan thanks for this! Could you modify it so it excludes all hidden files (files that start with `.`)?

PR #41, opened 2023-04-04
Integrate Zotero

MilesCranmer:
Zotero is perhaps the most popular tool in academia for storing a database of citations/PDFs/notes of research papers, so it seems like it would be very useful if `paper-qa` could directly search a Zotero database.

This PR does the following:

1. Creates the `paperqa.contrib` module, to hold various plugins/interfaces in a separate module.
2. Creates the `paperqa.contrib.ZoteroDB` class to download PDFs and citation keys from a user's Zotero library (or a specific query of that Zotero library), and generate a `Docs` object from them.
3. All PDFs downloaded by this `ZoteroDB` class are by default cached into `~/.paperqa/zotero/{KEY}.pdf`.

---

Here's how this works: if you set `ZOTERO_USER_ID` and `ZOTERO_API_KEY` in your environment[^1], all you have to do is initialize the ZoteroDB object:

```python
from paperqa.contrib import ZoteroDB
zotero = ZoteroDB()
```

and it will now be set up to query from a user's zotero library and download PDFs.

This class inherits from [`pyzotero.zotero.Zotero`](https://github.com/urschrei/pyzotero). This parent class has a bunch of different functionality to query a Zotero library - much of that is inherited. This means you can do things like:

```python
zotero.top(q="Large Language Model", limit=10)
```

to view the top 10 hits from your Zotero library which contain "Large Language Models" in the title. More search syntax given [here](https://pyzotero.readthedocs.io/en/latest/?badge=latest#zotero.Zotero.add_parameters).

You can create a `Docs` object with the same syntax:

```python
docs = Docs()

for paper in zotero.iterate(q="Transformer", qmode="everything", sort="date", direction="desc", limit=20):
    docs.add(paper.pdf)
```

This function will lazily query the Zotero database (which is in the cloud) and find the first 20 papers, sorted by date in descending order, which have "Transformer" anywhere in the title/abstract (or full text, if the user has it) of the paper. It will lazily download 20 PDFs of those papers directly as saved inside the Zotero database (rather than needing to query an external service). If any paper doesn't have a PDF, it will get the next paper, until it reaches 20.

It will then put them all into a `docs` object for further research.

(You could also download the entire zotero library with `zot.gen_paperdb(limit=100000)`).

Let me know what you think! The API was a first draft so I am happy to tweak the names.

[^1]: See https://pyzotero.readthedocs.io/en/latest/?badge=latest#getting-started-short-version for getting these variables.

---

- Edit: Changed to `ZoteroQA` as the class name.
- Edit: Changed to `ZoteroDB`, and `iterate`. Also made `iterate` iterate over papers lazily, rather than download all at once.

### DIFF
--- a/paperqa/__init__.py
+++ b/paperqa/__init__.py
@@ -1,3 +1,2 @@
-from . import contrib
 from .docs import Docs, maybe_is_text, Answer
 from .version import __version__
--- a/paperqa/__init__.py
+++ b/paperqa/__init__.py
@@ -1,2 +1,3 @@
 from .docs import Docs, maybe_is_text, Answer
 from .version import __version__
+from . import contrib
--- a/paperqa/__init__.py
+++ b/paperqa/__init__.py
@@ -1,3 +1,3 @@
+from . import contrib
 from .docs import Docs, maybe_is_text, Answer
 from .version import __version__
-from . import contrib
--- a/paperqa/contrib/__init__.py
+++ b/paperqa/contrib/__init__.py
@@ -1,1 +1,1 @@
-from . import zotero
+from .zotero import QAZotero
--- a/paperqa/contrib/__init__.py
+++ b/paperqa/contrib/__init__.py
@@ -1,1 +1,1 @@
-from .zotero import ZoteroQA
+from .zotero import ZoteroDB
--- a/paperqa/contrib/__init__.py
+++ b/paperqa/contrib/__init__.py
@@ -1,1 +1,1 @@
-from .zotero import QAZotero
+from .zotero import ZoteroQA
--- a/paperqa/contrib/__init__.py
+++ b/paperqa/contrib/__init__.py
@@ -1,0 +1,1 @@
+from . import zotero
--- a/paperqa/contrib/zotero.py
+++ b/paperqa/contrib/zotero.py
@@ -13,7 +13,13 @@ StrPath = Union[str, Path]
 
 class QAZotero(zotero.Zotero):
     def __init__(
-        self, *, library_type: str = "user", library_id=None, api_key=None, storage=None, **kwargs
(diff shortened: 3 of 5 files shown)


MilesCranmer:
Not sure what this error is from. I can’t reproduce it locally. Any idea?

MilesCranmer:
Nevermind, it was a subtle mistake. Fixed!

MilesCranmer:
I wonder if a better API could just be to have `gen_paperdb` return a list of PDF paths. Then you could run with:

```python
docs = Docs()

for (paper, key) in zotero.gen_paperdb():
    docs.add(paper, key=key)
```

thoughts? You could even have zotero.gen_paperdb do the PDF downloads lazily, calling `yield` after each time it downloads one.

whitead:
Hi @MilesCranmer looks so awesome thank you! For your suggestion - a generator sounds like a nice idea tool. I'm working on some features in the `v1` branch to deal with larger amounts of documents - like keyword filters.

MilesCranmer:
Sounds good. Yeah I think the main usecase of ZoteroQA would just be to download PDFs from a Zotero library. Then they are in the `Docs` class and you could do whatever querying/keyword filtering you want. The query functionality here is just inherited from the pyzotero library but I wasn't imagining it to be the main query tool; it's just for choosing what PDFs to add to the local database.

MilesCranmer:
Okay I tweaked the API! Now use it as follows:

```python
from paperqa import Docs
from paperqa.contrib import ZoteroQA

zotero = ZoteroQA()
docs = Docs()

for item in zotero.gen_paperdb(limit=100):
    docs.add(item.pdf, key=item.key)

```

This will download PDFs lazily.

What do you think?

Issue #45, opened 2023-04-07
Cache embeddings and OCR results?


MilesCranmer:
Hey @whitead,

I think it would be really nice if I could spin up a paperqa instance without waiting for all papers to be OCR'd and embedded each time. Would it be okay to cache a few more things? Specifically:

1. The embeddings used for FAISS, and
2. The OCR of a PDF.

I think both of these could be stored in a separate cache file. The [`llmchain.cache.SQLiteCache`](https://github.com/hwchase17/langchain/blob/aa439ac2ff927d00190dd9b390ddb5a45552229b/langchain/cache.py#L60-L99) makes this pretty easy, as there are only two methods: `lookup` for getting a result, and `update` for updating the cache. Both use strings as keys and values, but you could serialize the metadata into a string.

Here's an example for the OCR. This would go into `readers.py` and wrap by `parse_pdf`:

```py
OCR_CACHE = llmchain.cache.SQLiteCache(CACHE_PATH.parent / "ocr_cache.db")
```

Then, `parse_pdf` would have[^1]:

```py
def parse_pdf(...):
    cache_key = dict(prompt=str(pdf_path), llm_string="")
    test_out = OCR_CACHE.lookup(**cache_key)

    out = _parse_pdf(...) if test_out is None else deserialize(test_out)

    if test_out is None: 
        OCR_CACHE.update(**cache_key, serialize(out))

    return out

def _parse_pdf(...):
    # Regular _parse_pdf
```

What do you think? For serialization I would use `json.dumps` and `json.loads`[^2]. One might argue that it's better to use a custom database for this but why not keep it simple as you are already using the llmchain database.

Cheers,
Miles

[^1]: You might want to hash the PDF file itself, or maybe the first N bytes, rather than the filename. But I think for starters the filename is simpler and fine.
[^2]: For posterity, it would be wise to also serialize the `paperqa.__version__` in each cache entry, and, if it's a different version, then ignore the cache and overwrite it.

whitead:
Great! Highly supportive - some of this work is duplicated from chroma, weaviate, pinecone, etc. But yes, I agree some slightly better caching would really help here.

Related to #38 - we should add an md5 hash on content when caching, instead of paths. Thanks for the contributions!

MilesCranmer:
Sounds good. We could also move that file hashing function into the `utils.py` folder for use elsewhere.

[...]

PR #46, opened 2023-04-07
Cache document parsing

MilesCranmer:
Hey @whitead,

This creates a cache for parsed documents, and gets half of #45 done. I just used the SQLite cache from langchain for simplicity, and deserialize/serialize results into the cache via `json.dumps`/`json.loads`. It helps quite a bit with speed of loading an previously-loaded PDF database, as the OCR doesn't need to re-run each time. The cache also supports other input filetypes, as the cache key is the hash of the file. The key also includes the paperqa version, so there shouldn't be compatibility issues with old caches.

I also refactored the paths into `paths.py`.

Let me know what you think!
Cheers,
Miles

---

- Edit 1: Uses the hash of the file as the cache key, rather than the pathname. This is much more robust.

### DIFF
--- a/paperqa/contrib/zotero.py
+++ b/paperqa/contrib/zotero.py
@@ -7,7 +7,7 @@ from collections import namedtuple
 
 from pyzotero import zotero
 
-from ..docs import CACHE_PATH
+from ..paths import CACHE_PATH
 
 StrPath = Union[str, Path]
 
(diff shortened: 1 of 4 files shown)


whitead:
@MilesCranmer - does it support pickling like existing saving/loading API?

MilesCranmer:
Yep! It just caches the function input->outputs instead of recomputing it each time. The Docs object is unchanged.

PR #50, opened 2023-04-10
Add `num_pages` attribute to `ZoteroPaper`

MilesCranmer:
When doing an embedding of my entire zotero library, I realized some of the PDFs are from course notes/textbooks which number in the hundreds of pages. Thus I think it would be useful to add a `num_pages` attribute to the `ZoteroPaper` type which is generated by `ZoteroDB`.

```python
docs = Docs()
zotero = ZoteroDB(library_type="user")

for item in zotero.iterate(limit=20):
    if item.num_pages > 30:
        continue  # skip long papers
    docs.add(item.pdf, key=item.key)
```

This uses the existing `pypdf` import, and takes about 2ms for a 20 page paper. 

Could also be useful to expose this to the paper scraper interface?

---

This also refactors a few things:

1. Moves custom types (`StrPath`) into `types.py`.
2. Cleans up `ZoteroPaper` implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ docs = paperqa.Docs()
 zotero = ZoteroDB(library_type="user")  # "group" if group library
 
 for item in zotero.iterate(limit=20):
+    if item.num_pages > 30:
+        continue  # skip long papers
     docs.add(item.pdf, key=item.key)
 ```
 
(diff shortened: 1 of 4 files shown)


whitead (approved):

whitead:
Thanks!

PR #62, opened 2023-04-12
Run nest_asyncio.apply() automatically inside Jupyter

g-simmons:
## What does this do?
On Docs.__init__(), try to check if we're inside a Jupyter notebook. If so, run `nest_asyncio.apply()`

## Test Status
- [x] Running the code example from the README fails in Jupyter if this is not included*
- [x] Running the code example from the README succeeds in Jupyter with this fix*
- [x] Running the code example from the README succeeds in the terminal*

*only tested up to Docs.add()

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -66,6 +66,25 @@ class Docs:
             embeddings = OpenAIEmbeddings()
         self.embeddings = embeddings
 
+        if self._in_notebook():
+            import nest_asyncio
+            nest_asyncio.apply()
+    
+    def _in_notebook(self) -> bool:
+        # From https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook
+        try:
+            from IPython import get_ipython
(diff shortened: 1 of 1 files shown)


paperqa/docs.py
```diff
@@ -66,6 +66,25 @@ def __init__(
             embeddings = OpenAIEmbeddings()
         self.embeddings = embeddings
 
+        if self._in_notebook():
+            import nest_asyncio
+            nest_asyncio.apply()
+    
+    def _in_notebook(self) -> bool:
```
MilesCranmer, line 73: Perhaps this function could be put in `utils.py` (outside of the class), since it doesn't use `self` in the method?

whitead:
Does this work? I already have code for this [here](https://github.com/whitead/paper-qa/blob/93c47ccc7ea496a1167ac2b89e7fa512dee7be7f/paperqa/docs.py#L365) but it doesn't seem to work.

whitead:
Since jupyter notebooks support async natively now, going to close this

PR #65, opened 2023-04-15
Created an Application that can be run in CLI

malekinho8:
- Added unit tests for new functions
- Created utility functions related to parsing bibtex files from Zotero clipboard data
- Implemented unit tests
- Added new methods for Docs object to add documents from Zotero clipboard data, as well as PDF URL, or specific PDF path on the local machine
- Added comments and other minor additions to functions in the code for readability
- Added Usage documentation for the application in README.md
- Updated dev-requirements.txt
- Moved PDF loading to use fitz (from PyMuPDF) for faster loading times


whitead:
Hi @malekinho8,

thanks for your contributions! There is a lot going on in this PR and I think it won't be possible to merge as is. You'll need to split up your ideas into a few pieces. I've put some quick comments below, but can give more detailed feedback once split up.

Replacing the reader with the one you've chosen has a license incompatible with this package (https://github.com/pymupdf/PyMuPDF#license-and-copyright), so we cannot change to that reader

The CLI idea is great, but I would recommend you use the click library and get the CLI to be part of install, rather than a contained python file.

The zotero/bibtex changes should be their own PR as well. Some of the things you're writing, like converting from bibtex to MLA, are probably best done using a library rather than writing it ourselves.

malekinho8:
Hey @whitead,

Thanks for the feedback, I am admittedly ignorant about licenses and currently learning a lot. I've done a bit of digging and found a repository based on Poppler (another C++ PDF reader similar to MuPDF), which has a fairly well-maintained Python wrapper: `pdftotext` - repo [here](https://github.com/jalan/pdftotext). This repository has an MIT license, but I was confused because Poppler uses a GPLv2/3 license, so I thought all derivative work should also be GPLv2/3. Someone opened an issue on the repo [here](https://github.com/jalan/pdftotext/issues/81) regarding this concern about 2 years ago, and it has been closed now, with the owner saying that his repo is not definitively classified as deriviative work since no legal precedent has been set in the last 30 years to go off of regarding this area. My question for you was if you considered using the `pdftotext` library, or if you are avoiding it for this reason? I just really want to have a question-answering tool that works efficiently, and the `pypdf` library is orders of magnitude slower for reading PDFs compared to `pdftotext` and `fitz`. I've tried several open source MIT licensed packages and for PDFs in the range of 8 pages or so, the performance difference is still large but not ridiculous, however for PDFs with large number of pages, the difference becomes basically exponential in nature (i.e. C++ based PDF readers take ~1 second, while others take >50 seconds). 

image

I will try to split up my PRs shortly and implement some of the suggestions you made (I spent 30 minutes - 1 hours looking for bibtex to mla libraries, but every one I tried threw a bunch of errors or was not generalizable. I think I will make another github repository to store those functionalities).

PR #74, opened 2023-04-17
Added basic CLI functionality based on click module. 

malekinho8:
Added a method to Docs object that allows user to enter 'chat' mode, which is basically just query on an infinite loop. Reflected changes in setup.py to allow for easy installation. Updated README.md to include a Usage example.

I tried to implement changes based on your suggestion, but not 100% sure if I did everything right. For now I put `cli.py` in the `paperqa.contrib` folder which houses the code to enter into chat mode with the Docs object from the CLI. LMK your thoughts/next steps whenever you get a chance!


MilesCranmer:
Good idea. Might be nice to allow an arbitrary number of arguments? Then one could combine with shell globbing, like:

```
paperqa my_pdfs/**/*.pdf
```

paperqa/docs.py
```diff
@@ -448,3 +448,18 @@ async def aquery(
         answer.references = bib_str
         answer.passages = passages
         return answer
+    
+    def chat(self):
```
whitead, line 452: Let's move this to CLI, and not call it `chat`, since there is no memory of the conversation

paperqa/cli.py
```diff
@@ -0,0 +1,76 @@
...
+    click.echo("This is a tool for answering any question you may have about scientific papers!")
+
+    # initialize the answer and file_path variables to None
+    answer = None
+    file_paths = []
+
+    # check if any command-line arguments were provided
+    if sys.argv:
+        # if exactly one argument is provided, assume it is a file path or glob pattern
+        if len(sys.argv) == 2:
```
whitead, line 29: The click module supports glob filepaths I believe, let's use that instead

paperqa/cli.py
```diff
@@ -0,0 +1,76 @@
...
+        answer = inquirer.prompt(
+            [inquirer.List("type", message="How can I help you today?", choices=[
+                PDF_URL, 
+                FILES, 
+                ZOTERO_CLIPBOARD,
+                # TODO: Add more options, such as ability to query a webpage from URL, a github repo, YouTube video, etc.
+                ])])["type"]
+
+    # handle user input
+    if answer == PDF_URL:
```
whitead, line 52: Can we just remove this for now?

whitead (changes requested):

jamesbraza:
Hello @malekinho8 thank you for this PR. Version 5 (just released) ships with a CLI, so I am going to close this out.

We appreciate your help and please feel free to continue opening PRs oriented towards the new CLI

Issue #66, opened 2023-04-16
Directory structure for ZoteroDB?


g-simmons:
I'm on Zotero v6.0.23, on Mac OS, and attachments get stored in the following format:

`/Users/myusername/Zotero/storage/key/papertitle.pdf`

If I initialize a ZoteroDB with `storage=/Users/myusername/Zotero/storage/`, it seems like ZoteroDB will look for and store attachments as 

`/Users/myusername/Zotero/storage/key.pdf`

On my system this leads to some unnecessary redownloading that can be avoided by modifying the behavior of `ZoteroDB.get_pdf()`. 

Am I understanding this correctly? And do you know if Zotero uses the same directory structure across platforms?

MilesCranmer:
I think this should be raised in [pyzotero](https://github.com/urschrei/pyzotero) rather than here, as we are just calling their zotero downloading API

g-simmons:
Thanks for the response. I'll try to make some clarifications for why I think a change in paper-qa might be appropriate. LMK what you think...

### Download location is set within paper-qa
I'm aware that paper-qa is using pyzotero, but it looks like the specific download location for pdfs is set within paper-qa.contrib at:

https://github.com/whitead/paper-qa/blob/93c47ccc7ea496a1167ac2b89e7fa512dee7be7f/paperqa/contrib/zotero.py#L107

This results in files being downloaded to the storage folder, with a name based on their key. For example a paper titled "Attention is all you Need" with key "ABC123" gets downloaded by contrib.ZoteroDB.get_pdf() to:

```
<storage>/ABC123.pdf
```


### Download location in paper-qa doesn't match default Zotero download location

At least on OSX (I only have evidence from my system), Zotero makes a separate subfolder of the storage folder, with subfolder name based on key, and puts pdfs in the subfolders, with names based on the paper titles. For example a paper titled "Attention is all you Need" with key "ABC123" gets downloaded by the Zotero browser extension to:

```
<storage>/ABC123/Attention-is-all-you-need.pdf
```

### Assumptions:
- it's typical for many users to have already downloaded PDFs for much of their library
- there no reason why paper-qa wouldn't want to use those already-downloaded PDFs if possible
- Zotero default file storage is the same cross-platform
- Zotero users don't often change the default file storage conventions

### Suggested change
[...]

PR #82, opened 2023-04-24
copy from Zotero storage

g-simmons:
Copy files from Zotero storage instead of downloading, if possible. Discussed in #66.

### Changes
- import shutil for file copy
- add zotero_storage arg to ZoteroDB init to specify the Zotero storage location, default is "~/Zotero/storage/"
- if pdf for a document doesn't exist yet, first check Zotero storage.
- if one pdf exists in Zotero storage, copy it to the paperqa storage location and return the destination path
- if there is an error attempting to copy the pdf, fall back to downloading
- if zero or more than one pdfs exist, fall back to downloading

### DIFF
--- a/paperqa/contrib/zotero.py
+++ b/paperqa/contrib/zotero.py
@@ -1,5 +1,6 @@
 # This file gets PDF files from the user's Zotero library
 import os
+import shutil
 from typing import Union, Optional
 from pathlib import Path
 import logging
@@ -44,8 +45,22 @@ class ZoteroDB(zotero.Zotero):
         library_id: Optional[str] = None,
         api_key: Optional[str] = None,
         storage: Optional[StrPath] = None,
+        zotero_storage: Optional[Union[StrPath,bool]] = "~/Zotero/storage/",
         **kwargs,
     ):
+        """Initialize the ZoteroDB object.
+
+        Parameters
+        ----------
+        storage: str, optional
+            The path to the directory where PDFs will be stored. Defaults to
+            `~/.paperqa/zotero`.
+        zotero_storage: str, optional
+            The path to storage directory where Zotero stores PDFs. Defaults to
+            `~/Zotero/storage/`. Set this to use previously-downloaded PDFs. Set to `False` to
+            disable this feature.
+        """
+
         self.logger = logging.getLogger("ZoteroDB")
 
         if library_id is None:
@@ -76,6 +91,11 @@ class ZoteroDB(zotero.Zotero):
 
         if storage is None:
             storage = CACHE_PATH.parent / "zotero"
+        
+        if zotero_storage:
+            self.zotero_storage = Path(zotero_storage).expanduser()
+        else:
+            self.zotero_storage = None
 
         self.logger.info(f"Using cache location: {storage}")
         self.storage = storage
@@ -107,6 +127,22 @@ class ZoteroDB(zotero.Zotero):
         pdf_path = self.storage / (pdf_key + ".pdf")
 
         if not pdf_path.exists():
+            if self.zotero_storage:
+                self.logger.info(f"|  Looking for existing PDF for: {self._get_citation_key(item)}")
+                try:
+                    zotero_doc_folder = self.zotero_storage / pdf_key
+
+                    if zotero_doc_folder.exists():
+                        pdf_files = list(zotero_doc_folder.glob("*.pdf"))
+                        if len(pdf_files) == 1:
+                            self.logger.info(f"|  Copying existing PDF for {self._get_citation_key(item)} from Zotero storage.")
+                            zotero_pdf_path = zotero_doc_folder / pdf_files[0]
+                            shutil.copy(zotero_pdf_path, pdf_path)
+                            return pdf_path
+
+                except Exception as e:
+                    self.logger.warning(f"Could not copy file from Zotero storage, redownloading file. Error: {e}")
+
             pdf_path.parent.mkdir(parents=True, exist_ok=True)
             self.logger.info(f"|  Downloading PDF for: {_get_citation_key(item)}")
             self.dump(pdf_key, pdf_path)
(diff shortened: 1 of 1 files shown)


paperqa/contrib/zotero.py
```diff
@@ -44,8 +45,22 @@ def __init__(
         library_id: Optional[str] = None,
         api_key: Optional[str] = None,
         storage: Optional[StrPath] = None,
+        zotero_storage: Optional[Union[StrPath,bool]] = "~/Zotero/storage/",
```
MilesCranmer, line 48:
```suggestion
        zotero_storage: Optional[Union[StrPath,bool]] = None,
```

And initialize it later on, based on what operating system it is.

Also, should use `pathlib.Path.home() / "Zotero" / "storage"` for safety

paperqa/contrib/zotero.py
```diff
@@ -76,6 +91,11 @@ def __init__(
 
         if storage is None:
             storage = CACHE_PATH.parent / "zotero"
+        
+        if zotero_storage:
+            self.zotero_storage = Path(zotero_storage).expanduser()
+        else:
+            self.zotero_storage = None
```
MilesCranmer, line 98: Here is where you can automatically set it, if the OS default exists.

paperqa/contrib/zotero.py
```diff
@@ -107,6 +127,22 @@ def get_pdf(self, item: dict) -> Union[Path, None]:
...
+
+                    if zotero_doc_folder.exists():
+                        pdf_files = list(zotero_doc_folder.glob("*.pdf"))
+                        if len(pdf_files) == 1:
+                            self.logger.info(f"|  Copying existing PDF for {self._get_citation_key(item)} from Zotero storage.")
+                            zotero_pdf_path = zotero_doc_folder / pdf_files[0]
+                            shutil.copy(zotero_pdf_path, pdf_path)
+                            return pdf_path
+
+                except Exception as e:
```
MilesCranmer, line 143: What other exceptions are you thinking of here? Maybe catch them explicitly, and throw the error if something unexpected comes up?

paperqa/contrib/zotero.py
```diff
@@ -107,6 +127,22 @@ def get_pdf(self, item: dict) -> Union[Path, None]:
...
+                try:
+                    zotero_doc_folder = self.zotero_storage / pdf_key
+
+                    if zotero_doc_folder.exists():
+                        pdf_files = list(zotero_doc_folder.glob("*.pdf"))
+                        if len(pdf_files) == 1:
+                            self.logger.info(f"|  Copying existing PDF for {_get_citation_key(item)} from Zotero storage.")
+                            zotero_pdf_path = zotero_doc_folder / pdf_files[0]
+                            shutil.copy(zotero_pdf_path, pdf_path)
+                            return pdf_path
```
MilesCranmer, line 141:
```suggestion
                            return pdf_path
                        else:
                            self.logger.warning("|  Found more than one PDF for {_get_citation_key(item)}, so skipping.")
```

paperqa/contrib/zotero.py
g-simmons: Ok so we could do:

```
Optional[Union[StrPath,bool]] = None

StrPath: try to use the user-provided path, raise FileNotFoundError if it doesn't exist
None: try to use pathlib.Path.home() / "Zotero" / "storage", if it doesn't exist, set zotero_storage to None
False: never use zotero_storage
```

But what about

```
Union[StrPath,bool] = True

StrPath: try to use the user-provided path, raise FileNotFoundError if it doesn't exist
True: try to use pathlib.Path.home() / "Zotero" / "storage", if it doesn't exist, set zotero_storage to None
False: never use zotero_storage
```

I think default True makes it more obvious that the functionality is engaged by default. WYT?

g-simmons: Also out of curiosity what's the diff btw `pathlib.Path.home() / "Zotero" / "storage"` and `Path("~/Zotero/storage/").expanduser()`?

paperqa/contrib/zotero.py
g-simmons: Are you suggesting one of the three things below? Or something else?

catch specific error cases and resolve them
catch, warn and fall back to downloading
catch, wrap with a message, and raise

For unexpected edge cases, do you think raising an error would be better or just warning the user and falling back to downloading?

g-simmons: Didn't have anything specific in mind - could imagine permissions errors at the destination dir.

paperqa/contrib/zotero.py
MilesCranmer: > <Pathlib question>

It's just a more robust practice: On Windows, the "\" key is used instead of "/". So Pathlib's `/` operator will do the correct thing on each operating system. 

Also sometimes the `"~"` is not expanded correctly. Better to rely on Pathlib.

MilesCranmer: Being explicit is often best so I would recommend one parameter for specifying the location, and another for specifying whether you should use it or not. When variables can take on a bunch of different types it can make code confusing. (I edited my original comment with this suggestion)

paperqa/contrib/zotero.py
MilesCranmer: If there was no specific error you had in mind I would just remove the try-except. If you think there could be permissions issues for the `shutil.copy` I would check for the specific permissions error, and warn, then fall back to downloading. If there's an unexpected error then it should be raised.

jamesbraza:
Hi @g-simmons thank you for this PR. We just released v5 today which changes basically the whole repo.

If you are still interested, please rebase this PR atop the current main branch.

g-simmons:
@jamesbraza thanks for the info. I probably won't be able to rebase this any time soon (other time commitments) :/

whitead:
Closing for now - feel free to re-open when applicable!

PR #94, opened 2023-05-02
change callback manager for latest version of langchain

nachovy:


whitead:
Thanks - should be fixed by #92 though

Issue #64, opened 2023-04-14
Using a local model has issues


Kohulan:
Hi @whitead ,

I used the following code block and used a quantized LLM from here: https://huggingface.co/Pi3141

The model I used: https://huggingface.co/Pi3141/gpt4-x-alpaca-native-13B-ggml

```python3
import paperscraper
from paperqa import Docs
from langchain.embeddings import LlamaCppEmbeddings
from langchain.llms import LlamaCpp

llm = LlamaCpp(model_path="ggml-model-q4_1.bin")
embeddings = LlamaCppEmbeddings(model_path="ggml-model-q4_1.bin")
docs = Docs(llm=llm, embeddings=embeddings)

keyword_search = 'bispecific antibody manufacture'
papers = paperscraper.search_papers(keyword_search, limit=2)
for path,data in papers.items():
    try:
        docs.add(path,chunk_chars=500)
    except ValueError as e:
        print('Could not read', path, e)

answer = docs.query("What manufacturing challenges are unique to bispecific antibodies?")
print(answer)
```

- This did run for nearly 1.5 hours and crashed.

Error log:
```stdout
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
Input In [4], in <cell line: 12>()
     10         print('Could not read', path, e)
     11 print("Im here")
---> 12 answer = docs.query("What manufacturing challenges are unique to bispecific antibodies?")
     13 print(answer)
     14 end_time = time.time()

File ~/.local/lib/python3.10/site-packages/paperqa/docs.py:374, in Docs.query(self, query, k, max_sources, length_prompt, marginal_relevance, answer, key_filter)
    372     loop = asyncio.new_event_loop()
    373     asyncio.set_event_loop(loop)
--> 374 return loop.run_until_complete(
    375     self.aquery(
    376         query,
    377         k=k,
    378         max_sources=max_sources,
    379         length_prompt=length_prompt,
    380         marginal_relevance=marginal_relevance,
    381         answer=answer,
    382         key_filter=key_filter,
    383     )
    384 )

File ~/.local/lib/python3.10/site-packages/nest_asyncio.py:89, in _patch_loop.<locals>.run_until_complete(self, future)
     86 if not f.done():
     87     raise RuntimeError(
     88         'Event loop stopped before Future completed.')
---> 89 return f.result()

File /usr/lib/python3.10/asyncio/futures.py:201, in Future.result(self)
    199 self.__log_traceback = False
[...]

PR #96, opened 2023-05-03
fix: update ReadMe on how to use Local models

Kohulan:
@whitead 
Added information to the readme on how to use local models with a new update.
Issue #64 is now fixed with #92


whitead:
@Kohulan - could you double-check latest version of langchai? I believe now just adding `verbose=True` causes LLMs to stream to stdout - no need to set-up handlers.

Kohulan:
> @Kohulan - could you double-check latest version of langchai? I believe now just adding `verbose=True` causes LLMs to stream to stdout - no need to set-up handlers.

Agreed. Modified now.

PR #133, opened 2023-06-05
Download files from urls.

Mason-McGough:
Download URLs to tempfiles before processing.

See https://github.com/whitead/paper-qa/issues/72

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -31,7 +31,14 @@ from .qaprompts import (
 )
 from .readers import read_doc
 from .types import Answer, Context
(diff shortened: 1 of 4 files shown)


Mason-McGough:
Thanks for the suggestions, @whitead. I made a couple of new commits to address them.

whitead:
Thanks so much for your PR! I took the idea and put it into V3, but I'm not able to merge your PR because of how much was refactored. I added you though as a co-author on the v3 release commit and will make sure you're credited in release notes.

Mason-McGough:
Thanks!

PR #147, opened 2023-06-17
Fix the arguments for a function call to add_embeddings

thenatefisher:
Fix the arguments for a function call to add_embeddings in order to make a list of (text,embedding) tuples.


thenatefisher:
Hi there, great project! I noticed after loading a saved pickle that I could not continue to add documents to the Docs object. It seems there may be a problem with this function call to add_embeddings. Not sure if there is a better way or if I was just doing something wrong. Best- Nate

whitead:
This is probably a change in langchain versions - let me double-check the new syntax. Thanks!

PR #167, opened 2023-07-24
Detailed citation bugfix

whitead:


whitead:
Going to focus on #168

PR #172, opened 2023-08-01
Add embed_query_prefix option for embedding models like e5

### LINKED ISSUE #171: Allow prefixing the query before embedding search
Some embedding models (like [e5](https://huggingface.co/intfloat/e5-large-v2)) require prefixing the query with `query: ` and documents with `passage: ` tokens. The later can be handled by manually embedding everything first, but there's currently no way to inject the query prefix just for the embedding search.

zacps:
Some embedding models, like e5, require queries and passages to be prefixed with strings like 'query: ' and 'passage: ' to obtain the best results. This commit adds a parameter to Docs which allows settings the query embedding search prefix, a prefix can be added to passages by manually chunking and embedding them before handing them to paper-qa.

Fixes #171

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -53,6 +53,7 @@ class Docs(BaseModel, arbitrary_types_allowed=True, smart_union=True):
     memory: bool = False
     memory_model: Optional[BaseChatMemory] = None
     jit_texts_index: bool = False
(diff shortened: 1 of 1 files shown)


zacps:
Closing as I think this is better handled by just subclassing the relevant langchain embedding class to add the prefix.

PR #179, opened 2023-08-08
adding support for claude, azure, cohere, replicate

krrishdholakia:
Hi @whitead 

Noticed you're calling OpenAI w/ langchain. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and was wondering if we could be helpful.

Added support for Claude, Cohere, Azure and Llama2 (via Replicate) by setting client in the langchain ChatOpenAI class. The code is pretty similar to the OpenAI class - as litellm follows the same pattern as the openai-python sdk.

Would love to know if this helps.

Happy to add additional tests / update documentation, if the initial PR looks good to you.

### DIFF
--- a/paperqa/chains.py
+++ b/paperqa/chains.py
@@ -6,6 +6,8 @@ from langchain.callbacks.manager import (
     AsyncCallbackManagerForChainRun,
     CallbackManagerForChainRun,
 )
+import litellm
+from litellm import completion
 from langchain.chains import LLMChain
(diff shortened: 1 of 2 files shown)


fabian57fabian:
I was also searching for cohere support.
Adding +1 to the PR
Also, if needed help i could code, but the branch need to be clean and tests passing

jamesbraza:
Implemented by https://github.com/Future-House/paper-qa/pull/362

PR #181, opened 2023-08-11
delete the texts

Dustyposa:
### Update
- also delete the `doc.texts` when delete the doc. 

I can't make sure this function is been designed for this, but I have a case need do so.


Dustyposa:
@whitead Thanks for confirm workflow. Delated the empty line, forgot that.

PR #192, opened 2023-09-01
Pydantic 2 PR

whitead:


whitead:
Determined langchain doesn't support pydantic v2 yet!

PR #193, opened 2023-09-06
Added support for pydantic v2

whitead:


filippo82:
Hi @whitead 👋🏻 

`pydantic` is still pinned to `<2` in `setup.py`. Shouldn't that file also be updated?

filippo82:
Ok, please ignore my previous comment, this PR is indeed just to add support to v2. I see that there is a branch were you are most likely working to make it the default version.

PR #200, opened 2023-10-30
ci: fix langchain version

johannrichard:
fix langchain dependency to below v0.0.324 due to incompatibilities of the current code base with newer `langchain` versions (fixes https://github.com/whitead/paper-qa/issues/199)

```
langchain<0.0.324>=0.0.198
```


whitead:
Thanks @johannrichard! I just adjusted the import to work with recent langchain versions in #201

PR #203, opened 2023-11-06
Added option to strip indirect citations

whitead:
This enables by default removal of citation markers from summaries that are similar to what is generated in the final response, so that indirect citations (citations that are mentioned in summary) are not propagating up to final answers.


paperqa/docs.py
```diff
@@ -505,6 +508,8 @@ async def process(match):
                 raise e
             if "not applicable" in context.lower() or "not relevant" in context.lower():
                 return None
+            if self.strip_citations:
```
jakublala, line 511:
```suggestion
            if self.strip_citations:
            # remove citations that collide with our grounded citations (for the answer LLM)
```

jakublala (approved):
Made a suggestion to add a comment. Otherwise, good to merge. 

This part of the codebase has many sequential parts doing different things, so it'd be great if we comment on specific decisions made in our 'pipeline' design.

PR #252, opened 2024-03-08
Migrating from `setup.py` to `pyproject.toml`

jamesbraza:
- Moves from `setup.py` to `pyproject.toml` to remove a moving part
- Adds `pytest` `testpaths` for convenience


mskarlin (approved):
pulled locally -- everything is works well. 

@jamesbraza would you mind leaving a note in `paperqa/version.py` that version assignment is pulled from the pyproject.toml as a reminder for the next person who increments.

paperqa/version.py
```diff
@@ -1 +1,3 @@
-__version__ = "4.3.0"
+from importlib.metadata import version
+
+__version__ = version("paper-qa")
```
jamesbraza, line 3:
```suggestion
# If you're looking to bump the version, this now lives
# in pyproject.toml's [project] section
__version__ = version("paper-qa")
```

PR #254, opened 2024-03-12
Separate parsing and chunking in `read_doc`

mskarlin:
This allows users to implement their own chunking if they choose, while retaining the same parsing as in standard PQA implementations. 

PR optionally allows users to get metadata about the parsing or chunking, for persistence / experiment tracking.

### DIFF
--- a/paperqa/readers.py
+++ b/paperqa/readers.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from math import ceil
 from pathlib import Path
-from typing import List  # noqa: F401
+from typing import List, Literal, overload  # noqa: F401
 
 import html2text
 import tiktoken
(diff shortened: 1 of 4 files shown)


paperqa/readers.py
```diff
@@ -4,56 +4,70 @@
...
     pages: list[str] = []
     texts: list[Text] = []
-    for i, page in enumerate(pdfReader.pages):
-        split += page.extract_text()
-        pages.append(str(i + 1))
+    split: str = ""
+
+    if not isinstance(parsed_text.content, dict):
+        raise NotImplementedError(
+            "pdf chunking only implemented per-page, ParsedText.content must be a `dict`."
```
jamesbraza, line 65: Can you use an f-string to include the actual type of `parsed_text.content`?

```python
f"ParsedText.content must be a `dict`, not {type(parsed_text.content)}."
```

paperqa/readers.py
```diff
@@ -152,25 +206,80 @@ def parse_code_txt(path: Path, doc: Doc, chunk_chars: int, overlap: int) -> list
...
     path: Path,
     doc: Doc,
     chunk_chars: int = 3000,
     overlap: int = 100,
     force_pypdf: bool = False,
-) -> list[Text]:
-    """Parse a document into chunks."""
+    parsed_text_only: bool = False,
+    include_metadata: bool = False,
+) -> list[Text] | ParsedText | tuple[list[Text], ParsedMetadata]:
```
jamesbraza, line 217: (per discussion in-person)

This complex return type arises from multiple `bool` configuration arguments, which will lead to installers of `paper-qa` (and within `paper-qa` itself) having to add `type: ignore` comments. An immediate (yet verbose) solution is using `typing.overload` to better express this signature.

However, in general, this complex signature is a code smell for a decomposition needing to take place. Maybe in the future we can decompose this better to make the code more fluent without requiring big signatures

jamesbraza (approved):

paperqa/readers.py
mskarlin: Added! Was able to remove the type ignores from `docs.py` after implementing.

jamesbraza: https://youtu.be/E5KfzyO4feo?t=21

@mskarlin is now Tron at 0:21 after making this change

PR #262, opened 2024-03-29
Added batching to embed documents

whitead:


paperqa/llms.py
```diff
@@ -87,17 +87,23 @@ def process_llm_config(
...
-    response = await client.embeddings.create(
-        model=embedding_model, input=texts, encoding_format="float"
-    )
-    return [e.embedding for e in response.data]
+    N = len(texts)
+    embeddings = []
+    for i in range(0, N, batch_size):
+        response = await client.embeddings.create(
+            model=embedding_model,
+            input=texts[i : i + batch_size],
```
jamesbraza, line 102: Aside: Python 3.12 has [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched) to make this easier, but let's just stick with this for now

jamesbraza (approved):
LGTM

paperqa/llms.py
whitead: Good to know - this will good to know in a few years once we get adoption.

PR #263, opened 2024-03-30
`dev` extra and cleaned up CI

jamesbraza:
- Enables `pip install -e .[dev]` and moves CI to use this
- Uses the latest GitHub Actions versions
- Alphabetizes and groups `dev-requirements.txt`


whitead (approved):
Thanks for simplifying the build! Assuming it doesn't need to be installed for building wheels based on your commit - good to know!

jamesbraza:
I tested `pip install .[dev] && python -m build --sdist --wheel --outdir dist/ .` works just now to be sure

PR #264, opened 2024-03-30
Delaying of state mutation in `Docs.aadd_texts`

jamesbraza:
Delaying mutation of `Docs`, `Doc`, and `Texts` in `Docs.aadd_texts` until after `embed_documents` completes

This enables callers to retry in the event of `openai.RateLimitError`


whitead (approved):
Smart idea - LGTM!

PR #270, opened 2024-04-13
More robust JSON parsing

whitead:
* Added more json variants observed from Claude
* Added unit test with pytest features
* Removed older asyncio unit testing syntax
* Incremented version

### DIFF
--- a/paperqa/utils.py
+++ b/paperqa/utils.py
@@ -186,10 +186,8 @@ def llm_read_json(text: str) -> dict:
     """Read LLM output and extract JSON data from it."""
     # fetch from markdown ```json if present
     text = text.strip().split("```json")[-1].split("```")[0]
-    print(text)
     # split anything before the first {
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
-    print(text)
     return json.loads(text)
--- a/paperqa/utils.py
+++ b/paperqa/utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+import json
 import math
 import re
 import string
@@ -179,3 +180,16 @@ def is_coroutine_callable(obj):
     elif callable(obj):  # noqa: RET505
         return inspect.iscoroutinefunction(obj.__call__)
     return False
+
+
+def llm_read_json(text: str) -> dict:
+    """Read LLM output and extract JSON data from it."""
+    # fetch from markdown ```json if present
+    text = text.strip().split("```json")[-1].split("```")[0]
+    print(text)
+    # split anything before the first {
+    text = "{" + text.split("{", 1)[-1]
+    # split anything after the last }
+    text = text.rsplit("}", 1)[0] + "}"
+    print(text)
+    return json.loads(text)
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -523,7 +523,7 @@ async def test_anthropic_chain():
     assert completion.prompt_count > 0
     assert completion.completion_count > 0
     assert str(completion) == "".join(outputs)
-    assert type(completion.text) is str  # noqa: E721
+    assert isinstance(completion.text, str)
 
     completion = await call({"animal": "duck"})  # type: ignore[call-arg]
     assert completion.seconds_to_first_token == 0
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -5,7 +5,6 @@ import pickle
 import tempfile
 from io import BytesIO
 from pathlib import Path
-from unittest import IsolatedAsyncioTestCase
 
 import numpy as np
 import pytest
@@ -39,6 +38,7 @@ from paperqa.llms import (
 from paperqa.readers import read_doc
 from paperqa.utils import (
     get_citenames,
+    llm_read_json,
     maybe_is_html,
     maybe_is_text,
     name_in_text,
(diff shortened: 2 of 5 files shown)


paperqa/utils.py
```diff
@@ -179,3 +180,16 @@ def is_coroutine_callable(obj):
...
+def llm_read_json(text: str) -> dict:
+    """Read LLM output and extract JSON data from it."""
+    # fetch from markdown ```json if present
+    text = text.strip().split("```json")[-1].split("```")[0]
+    print(text)
+    # split anything before the first {
+    text = "{" + text.split("{", 1)[-1]
+    # split anything after the last }
+    text = text.rsplit("}", 1)[0] + "}"
+    print(text)
```
jamesbraza, line 194: Extra `print`s here?

tests/test_paperqa.py
```diff
@@ -409,101 +409,137 @@ def test_extract_score():
...
+    def accum(x):
+        outputs.append(x)
+
+    outputs: list[str] = []
+    completion = await call({"animal": "duck"}, callbacks=[accum])  # type: ignore[call-arg]
+    assert completion.seconds_to_first_token > 0
+    assert completion.prompt_count > 0
+    assert completion.completion_count > 0
+    assert str(completion) == "".join(outputs)
+    assert type(completion.text) is str  # noqa: E721
```
jamesbraza, line 526:
```suggestion
    assert isinstance(completion.text, str)
```

And a few lines below

tests/test_paperqa.py
```diff
@@ -948,126 +984,124 @@ def test_langchain_embeddings():
...
-        index = LangchainVectorStore(cls=FAISS, embedding_model=OpenAIEmbeddings())
-        assert index._store is None
+    ]
+
+    # checks on builder
+    try:
+        index = LangchainVectorStore()
         index.add_texts_and_embeddings(some_texts)
-        assert index._store is not None
+        raise "Failed to check for builder"  # type: ignore[misc]  # noqa: B016
```
jamesbraza, line 1005: `raise "Failed to check for builder"`

Did you forget an `Exception` type here? I don't think this is really valid:

```python
>>> raise "hi"
  File "<stdin>", line 1, in <module>
TypeError: exceptions must derive from BaseException
```

tests/test_paperqa.py
```diff
@@ -948,126 +984,124 @@ def test_langchain_embeddings():
...
+    except ValueError:
+        pass
```
jamesbraza, line 1007: For patterns like this, I like `with pytest.raises(ValueError)`

whitead: Yeah, this PR switches to pytest. Thanks for catching - will switch to pytest.

tests/test_paperqa.py
whitead: Fixed!

tests/test_paperqa.py
```diff
@@ -409,101 +409,137 @@ def test_extract_score():
...
+        client,
+        "The {animal} says",
+        skip_system=True,
+    )
+    outputs = []
+
+    def accum(x):
+        outputs.append(x)
+
+    completion = await call({"animal": "duck"}, callbacks=[accum])  # type: ignore[call-arg]
```
jamesbraza, line 485:
```suggestion
    outputs = []
    completion = await call({"animal": "duck"}, callbacks=accum.append)  # type: ignore[call-arg]
```

tests/test_paperqa.py
```diff
@@ -409,101 +409,137 @@ def test_extract_score():
...
+    assert completion.seconds_to_first_token > 0
+    assert completion.prompt_count > 0
+    assert completion.completion_count > 0
+    assert str(completion) == "".join(outputs)
+    assert isinstance(completion.text, str)
+
+    completion = await call({"animal": "duck"})  # type: ignore[call-arg]
+    assert completion.seconds_to_first_token == 0
+    assert completion.seconds_to_last_token > 0
+    assert type(completion.text) is str  # noqa: E721
```
jamesbraza, line 531:
```suggestion
    assert isinstance(completion.text, str)
```

tests/test_paperqa.py
```diff
@@ -948,126 +984,114 @@ def test_langchain_embeddings():
...
         index.add_texts_and_embeddings(some_texts)
-        assert index._store is not None
 
-        docs = Docs(
-            texts_index=LangchainVectorStore(
-                cls=FAISS, embedding_model=OpenAIEmbeddings()
-            )
-        )
-        assert docs._embedding_client is not None  # from docs_index default
+    with pytest.raises(ValueError):  # noqa: PT011
```
jamesbraza, line 1005: FYI you can do `pytest.raises(ValueError, match="some aspect of message")` for even more precise assertion

tests/test_paperqa.py
```diff
@@ -948,126 +984,114 @@ def test_langchain_embeddings():
...
+    index = LangchainVectorStore(
+        store_builder=lambda x, y: FAISS.from_embeddings(x, OpenAIEmbeddings(), y)
+    )
+    assert index._store is None
+    index.add_texts_and_embeddings(some_texts)
+    assert index._store is not None
+    # check search returns Text obj
+    data, score = await index.similarity_search(None, "test", k=1)  # type: ignore[unreachable]
+    print(data)
+    assert type(data[0]) == Text
```
jamesbraza, line 1021:
```suggestion
    assert isinstance(data[0], Text)
```

jamesbraza (approved):
Added some more comments

PR #272, opened 2024-04-18
Fixed flaky add_texts test

whitead:
* Switched to deterministic embedding - OAI ones keep coming back random
* Removed some stray print statements
* Added a test on embeddings being rank 1 (which was true but I'm paranoid)

### DIFF
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -571,7 +571,6 @@ def test_evidence():
     evidence = docs.get_evidence(
         Answer(question="For which state was Bates a governor?"), k=1, max_sources=1
     )
-    print(evidence.context)
     assert "Missouri" in evidence.context
     assert original_doc.embedding is not None, "Embedding should have remained in-tact"
     assert (
@@ -610,7 +609,6 @@ def test_json_evidence():
     evidence = docs.get_evidence(
         Answer(question="For which state was Bates a governor?"), k=1, max_sources=1
     )
-    print(evidence.context)
     assert "Missouri" in evidence.context
 
     evidence = docs.get_evidence(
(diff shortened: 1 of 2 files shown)


tests/test_paperqa.py
```diff
@@ -779,6 +776,12 @@ def test_sparse_embedding():
         dockey="test",
     )
     assert any(docs.docs["test"].embedding)  # type: ignore[arg-type]
+    assert all(
+        len(np.array(x.embedding).shape) == 1 for x in docs.docs.values()
+    ), "Embeddings should be 1D"
+    assert all(
+        len(np.array(x.embedding).shape) == 1 for x in docs.texts
+    ), "Embeddings should be 1D"
```
jamesbraza, line 784: If you wanna be cheeky:

```python
assert all(
    len(np.array(x.embedding).shape) == 1
    for x in itertools.chain.from_iterable((docs.docs.values(), docs.texts))
), "Embeddings should be 1D"
```

tests/test_paperqa.py
```diff
@@ -1607,27 +1610,46 @@ def disabled_test_memory():
 
 def test_add_texts():
     llm_config = {"temperature": 0.1, "model": "babbage-02"}
-    docs = Docs(llm_model=OpenAILLMModel(config=llm_config))
+    # want to use keyword embedding because
```
jamesbraza, line 1613: Just from my perspective, I am not sure how "keyword embedding" correlates to one wanting to use `SparseEmbeddingModel`

I guess, maybe clarify the comment with info how sparse embedding model is equated with keyword embedding

jamesbraza (approved):
Nice work

tests/test_paperqa.py
whitead: Beyond me :)

tests/test_paperqa.py
whitead: yea - typo. Meant to say sparse. Fixed in https://github.com/whitead/paper-qa/commit/9b2eb78a5fb013e2abda279889bff76567999cda

PR #276, opened 2024-04-29
Add VoyageAI embeddings

mskarlin:
Adding in some objects & tests which allow users to optionally use the VoyageAI embeddings.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,6 +7,7 @@ python-dotenv
 pyzotero
 requests
 sentence_transformers
+voyageai
 
 # Code QA dependencies
 build
--- a/paperqa/llms.py
+++ b/paperqa/llms.py
@@ -109,6 +109,9 @@ async def embed_documents(
 class EmbeddingModel(ABC, BaseModel):
     name: str
 
+    def set_mode(self, mode: str):
+        """Several embedding models have a 'mode' or prompt which affects output."""
+
     @abstractmethod
     async def embed_documents(self, client: Any, texts: list[str]) -> list[list[float]]:
         pass
@@ -155,6 +158,16 @@ class VoyageAIEmbeddingModel(EmbeddingModel):
     name: str = Field(default="voyage-large-2")
     embedding_type: str = Field(default="document")
     batch_size: int = 10
+
+    def set_mode(self, mode: str):
+        if mode == "query":
+            self.embedding_type = "query"
+        elif mode == "document":
+            self.embedding_type = "document"
+        else:
+            raise NotImplementedError(
+                f"Mode {mode} is not implemented for VoyageAI embeddings."
+            )
 
     async def embed_documents(self, client: Any, texts: list[str]) -> list[list[float]]:
         if client is None:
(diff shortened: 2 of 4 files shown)


whitead (approved):
LGTM!

dev-requirements.txt
```diff
@@ -7,6 +7,7 @@ python-dotenv
 pyzotero
 requests
 sentence_transformers
+voyageai
```
jamesbraza, line 10: Can you add an `extra` for this in `pyproject.toml`? That way we can do `pip install paper-qa[voyage-ai]`

paperqa/llms.py
```diff
@@ -109,6 +109,9 @@ async def embed_documents(
 class EmbeddingModel(ABC, BaseModel):
     name: str
 
+    def set_mode(self, mode: str):
```
jamesbraza, line 112:
```suggestion
    def set_mode(self, mode: str) -> None:
```

paperqa/llms.py
```diff
@@ -149,6 +152,40 @@ async def embed_documents(self, client, texts):
...
+
+    def set_mode(self, mode: str):
+        if mode == "query":
+            self.embedding_type = "query"
+        elif mode == "document":
+            self.embedding_type = "document"
+        else:
+            raise NotImplementedError(
+                f"Mode {mode} is not implemented for VoyageAI embeddings."
+            )
```
jamesbraza, line 170: Getting a code smell for an `Enum` with a method

paperqa/llms.py
```diff
@@ -149,6 +152,40 @@ async def embed_documents(self, client, texts):
...
+
+    async def embed_documents(self, client: Any, texts: list[str]) -> list[list[float]]:
+        if client is None:
+            raise ValueError(
+                "Your client is None - did you forget to set it after pickling?"
+            )
```
jamesbraza, line 176: Do we really need this? Imo this is the problem `mypy`/typing is supposed to fix

paperqa/llms.py
mskarlin: Thought that too, but this will be different per embedding model, Voyage uses 'query'/'document', but not some of the other embeddings on the MTEB. I guess we could force a mapping to these two values.

paperqa/llms.py
mskarlin: this is a mutable state thing where the client gets wiped every time you pickle it, and you manually need to reset it.

dev-requirements.txt
mskarlin: For posterity-- this is because the setuptools is configured as being "dynamic" which means it comes from an external file. Seems like setuptools doesn't support using both a dynamic file and having hard-coded dependencies within the pyproject.toml. 

I didn't want to have to build a separate external file just for this lib, so I kept it like this.

paperqa/llms.py
mskarlin: I'll swap this out with an enum actually -- just scope it for this single model.

dev-requirements.txt
jamesbraza: It's really too bad. I made https://github.com/pypa/setuptools/issues/4335 for this.

I guess a fix would be to move the `dev-requirements.txt` into this `pyproject.toml`

jamesbraza (approved):
Thanks for taking a look through my comments

PR #280, opened 2024-05-28
remove intro sentence in citation prompt

rosepolle:


rosepolle:
Adding this removed the "Here is the citation in the MLA format" intro sentence, improving clarity in the referencing (using 'gpt-4o' in my local testing).

whitead (approved):
Thanks!

PR #281, opened 2024-05-30
Fixed `summary_json_system_prompt`

jamesbraza:
- ~Duplicate curly braces~
    - Turns out these were necessary, documented why
- Trailing space


whitead (approved):
Thanks!

PR #282, opened 2024-06-06
Simplify langhcain config

whitead:
Made it so config can be used with LangchainLLMs


paperqa/llms.py
```diff
@@ -675,15 +675,10 @@ async def similarity_search(
         )
 
 
-# All the langchain stuff is below
-# Many confusing woes here because langchain
```
jamesbraza, line 679: Lol

jamesbraza (approved):

PR #284, opened 2024-06-09
Update docs.py

Yusufibin:
Improve error handling for external APIs

This commit improves the robustness and resilience of `paper-qa` by integrating more precise error handling for calls to external APIs, including OpenAI and Voyage AI.

Changes include:

- Introduced custom exceptions for rate limit, authentication, and other API errors.
- Added try-except blocks around API calls to catch and handle errors.
- Identification of the error type by analyzing the error message and re-throwing custom exceptions.

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from typing import Any, BinaryIO, Callable, Coroutine, cast
 from uuid import UUID, uuid4
 
-from openai import AsyncOpenAI
+from openai import AsyncOpenAI, OpenAIError
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 try:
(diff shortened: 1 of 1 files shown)


jamesbraza:
Hello @Yusufibin thank you very much for this PR!

This is outside of your purview, but we are moving this repo to use LiteLLM shortly, which actually does what you're trying to do here, but for all LLM providers: https://docs.litellm.ai/docs/exception_mapping

Given that, we don't want to merge this PR to avoid two back-to-back breaking changes in expected `Exception` types.

If you have interest in starting the LiteLLM migration, please feel free to take a stab at it 👍

jamesbraza:
Thanks again for this PR @Yusufibin. Today's version 5 moves us to use LiteLLM, which LiteLLM handles these sorts of behaviors (effectively we are outsourcing it to another package).

If there is something amiss in v5, please feel free to open a new PR 👌

PR #285, opened 2024-06-14
Support Anyscale OpenAI models

mskarlin:
Anyscale supports open source models via a custom base URL in the OpenAI client. This PR enables you to use them in PaperQA by setting the appropriate environment variables and model names. See the docs for Anyscale here: https://docs.anyscale.com/endpoints/text-generation/query-a-model/.

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -207,8 +207,8 @@ class Docs(BaseModel):
         if client is None and isinstance(self.llm_model, OpenAILLMModel):
             # will be defaults if ANYSCALE_BASE_URL and ANYSCALE_API_KEY are not set
             client = AsyncOpenAI(
-                api_key=os.environ.get("ANYSCALE_BASE_URL"),
-                base_url=os.environ.get("ANYSCALE_API_KEY"),
+                api_key=os.environ.get("ANYSCALE_API_KEY"),
+                base_url=os.environ.get("ANYSCALE_BASE_URL"),
             )
         self._client = client
         if embedding_client is None:
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -205,11 +205,15 @@ class Docs(BaseModel):
         embedding_client: Any | None = None,
     ):
         if client is None and isinstance(self.llm_model, OpenAILLMModel):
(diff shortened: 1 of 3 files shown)


paperqa/docs.py
```diff
@@ -205,7 +205,15 @@ def set_client(
         embedding_client: Any | None = None,
     ):
         if client is None and isinstance(self.llm_model, OpenAILLMModel):
-            client = AsyncOpenAI()
+            if (api_key := os.environ.get("ANYSCALE_API_KEY")) and (
```
jamesbraza, line 208: I think LiteLLM does this for us fwiw. So in the future we can consolidate this away

tests/test_paperqa.py
```diff
@@ -544,6 +558,49 @@ def accum(x):
...
+@pytest.mark.asyncio()
+async def test_anyscale_chain():
+    client = AsyncOpenAI(
+        base_url=os.environ["ANYSCALE_BASE_URL"], api_key=os.environ["ANYSCALE_API_KEY"]
+    )
+    llm = OpenAILLMModel(
+        config={
+            "temperature": 0,
+            "model": "meta-llama/Meta-Llama-3-8B-Instruct",
+            "max_tokens": 56,
```
jamesbraza, line 574: Can you document why 56 with a code comment? Or make this set by a formula?

tests/test_paperqa.py
```diff
@@ -544,6 +558,49 @@ def accum(x):
...
+        client,
+        "The {animal} says",
+        skip_system=True,
+    )
+    outputs = []
+
+    def accum(x):
+        outputs.append(x)
+
+    completion = await call({"animal": "duck"}, callbacks=[accum])  # type: ignore[call-arg]
```
jamesbraza, line 587:
```suggestion
    completion = await call({"animal": "duck"}, callbacks=[outputs.append])  # type: ignore[call-arg]
```

jamesbraza (approved):

tests/test_paperqa.py
mskarlin: it's arbitrary -- other tests use 56, i can do 50 I suppose.

paperqa/docs.py
mskarlin: for sure

PR #289, opened 2024-06-20
Refactor to functional API

whitead:
Large refactor to bring us closer to decoupling Docs, LLMs, and Config.

1. Moves to centralized config that can be loaded from package data or defaults. This changes many function calls so that they pass config as an object or name of a config object.
2. Removed `doc_match` and `doc_index` - these were only sometimes useful and not really part of current usage. We can add back if needed, but just complicated things.
3. Removed the unnecessary complicated `get_callbacks` factories. Now, callbacks can have a kwarg of `name` to get access to name of chain being called.
4. Switched to `contextvars` for setting `answer_id` in LLMResults so that we do not need to have so much back and forth for callbacks.
5. Deferred updates to `Answer` objects until end of functions so that retrying is possible (except token counts)
6. Generally moved all config out of `Docs`, except `LLMModel` which will be fixed in port to litellm


paperqa/agents/search.py
```diff
@@ -358,23 +356,22 @@ async def process_file(
...
-            if "chunk_chars" in docs_kwargs:
-                chunk_chars = int(docs_kwargs.pop("chunk_chars"))
-            if "overlap" in docs_kwargs:
-                overlap = int(docs_kwargs.pop("overlap"))
-
-            tmp_docs = Docs(**docs_kwargs)
+            # We set the LLM/SummaryLLM because
+            # it could be that all metadata fails and it has to peek to first page
+            # TODO: Eventually we will have an LLM object of some kinda passed here
+            # (or coming from settings)
```
jamesbraza, line 362: Is this actually a `TODO`? I don't see anything being "to do" here

Also, typo in `kinda`

paperqa/agents/search.py
```diff
@@ -405,42 +402,42 @@ async def process_file(
             logger.info(f"Complete ({title}).")
 
 
+# TODO: Use settings directly for some of this config
```
jamesbraza, line 405: What do you mean by use settings directly? You mean like move this to a method of `Settings`?

paperqa/configs/debug.json
```diff
@@ -0,0 +1,15 @@
+{
+  "llm": "openai/babbage-002",
```
jamesbraza, line 2: Can you use `json5` and comment why `babbage` is considered a debugging model?

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+        "about 200 words, but can be longer", description="Length of final answer"
+    )
+    max_concurrent_requests: int = Field(
+        default=4, description="Max concurrent requests to LLMs"
+    )
+    answer_filter_extra_background: bool = Field(
+        default=False,
+        description="Whether to cite background information provided by model.",
+    )
+    model_config = ConfigDict(extra="forbid")
```
jamesbraza, line 58: Let's move `model_config` to the top with a newline separating it, to keep Pydantic internals visually separate from the normal fields

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+    structured_citation_prompt: str = Field(
+        default=structured_citation_prompt,
+        description="Prompt that tries to creates a citation in JSON from peeking one page",
+    )
+    disable_doc_valid_check: bool = Field(
+        default=False,
+        description="Whether to disable checking if a document looks like text (was parsed correctly)",
+    )
+    chunking_algorithm: ChunkingOptions = ChunkingOptions.SIMPLE_OVERLAP
+    model_config = ConfigDict(extra="forbid")
```
jamesbraza, line 106: Let's move `model_config` to the top with a newline separating it, to keep Pydantic internals visually separate from the normal fields

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+            self.chunking_algorithm == ChunkingOptions.SIMPLE_OVERLAP
+            or parsing
+            in {  # type: ignore[unreachable]
+                _get_parse_type(p, self) for p in self.chunking_algorithm.valid_parsings
+            }
+        )
+
+
+# Mock a dictionary and store any missing items
+class _FormatDict(dict):
```
jamesbraza, line 135: Let's move this to a docstring

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+    tool_names: set[str] | None = Field(
+        default=None,
+        description=(
+            "Optional override on the tools to provide the agent. Leaving as the"
+            " default of None will use a minimal toolset of the paper search, gather"
+            " evidence, collect cited papers from evidence, and gen answer. If passing tool"
+            " names (non-default route), at least the gen answer tool must be supplied."
+        ),
+    )
+
```
jamesbraza, line 298:
```suggestion
```

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+        "\n- Gather more evidence using a different phrase"
+        "\nIf you search for more papers or collect new papers cited by previous evidence,"
+        " remember to gather evidence again."
+        " Once you have five or more pieces of evidence from multiple sources, or you have tried a few times, "
+        "call {gen_answer_tool_name} tool. The {gen_answer_tool_name} tool output is visible to the user, "
+        "so you do not need to restate the answer and can simply terminate if the answer looks sufficient. "
+        "The current status of evidence/papers/cost is {status}"
+    )
+    paper_directory: str | os.PathLike = Field(
+        default=Path.cwd(),
```
jamesbraza, line 257: Can we add to this field's description explaining why we have a relative path? My understanding is it enables the CLI to work regardless of the starting directory

Cc @mskarlin

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+            # kind of a hack to get list of attributes in answer
+            attrs = set(Answer.model_fields.keys())
+            if not set(get_formatted_variables(v)).issubset(attrs):
+                raise ValueError(f"Post prompt must have input variables: {attrs}")
+        return v
+
+
+class AgentSettings(BaseModel):
+    agent_llm: str = Field(
+        default="gpt-4o-2024-08-06",
```
jamesbraza, line 229: It would be nice to move this repeated variable to a module-level default with a good name:

```python
# Set this to a model that has fast inference, low cost
# and decent performance, for use as a common default
FAST_CHEAP_DECENT_LLM = "gpt-4o-2024-08-06"
```

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+            raise ValueError(
+                f"If using an override, must contain at least the {answer_tool_name}."
+            )
+        return v
+
+
+class Settings(BaseSettings):
+    llm: str = Field(
+        default="gpt-4o-2024-08-06",
+        description="Default LLM for most things, including answers. Should be 'best' LLM",
```
jamesbraza, line 324: If it should be the "best" LLM, shouldn't we use `gpt-4-turbo` here?

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+        default=1.0, description="Lambda for MMR in text index"
+    )
+    index_absolute_directory: bool = Field(
+        default=False,
+        description="Whether to use the absolute directory for the PQA index",
+    )
+    answer: AnswerSettings = AnswerSettings()
+    parsing: ParsingSettings = ParsingSettings()
+    prompts: PromptSettings = PromptSettings()
+    agent: AgentSettings = AgentSettings()
```
jamesbraza, line 350: We need to use `default_factory` for these to avoid mutable default issues

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+        # index name should use an absolute path
+        # this way two different folders where the
+        # user locally uses '.' will make different indexes
+        paper_directory = self.agent.paper_directory
+        if isinstance(paper_directory, Path):
+            paper_directory = str(paper_directory.absolute())
+
+        index_fields = "|".join(
+            [
+                str(paper_directory),  # cast for typing
```
jamesbraza, line 364:
```suggestion
                str(paper_directory),
```

There is no cast

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def get_settings(config_or_name: MaybeSettings = None) -> Settings:
+    if isinstance(config_or_name, Settings):
+        return config_or_name
+    if config_or_name is None:
+        return Settings()
+
+    config_name = config_or_name
+    return Settings.from_name(config_name)
```
jamesbraza, line 413:
```suggestion
    return Settings.from_name(config_name=config_or_name)
```

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+        if json_path:
+            return Settings.model_validate_json(json_path.read_text())
+
+        raise FileNotFoundError(f"No configuration file found for {config_name}")
+
+
+MaybeSettings = Settings | str | None
+
+
+def get_settings(config_or_name: MaybeSettings = None) -> Settings:
```
jamesbraza, line 406: Let's make this a `classmethod` of `Settings`, since that's mostly what this is

paperqa/config.py
```diff
@@ -0,0 +1,413 @@
...
+                f"No configuration file found for {config_name}"
+            ) from e
+
+        if json_path:
+            return Settings.model_validate_json(json_path.read_text())
+
+        raise FileNotFoundError(f"No configuration file found for {config_name}")
+
+
+MaybeSettings = Settings | str | None
```
jamesbraza, line 403: I don't really like this type alias, can we remove it?

paperqa/contrib/zotero.py
```diff
@@ -240,7 +240,7 @@ def iterate(  # noqa: C901, PLR0912
             _pdfs = [self.get_pdf(item) for item in _items]
 
             # Filter:
-            for item, pdf in zip(_items, _pdfs):
+            for item, pdf in zip(_items, _pdfs, strict=False):
```
jamesbraza, line 243: Can you add a comment why we're not `strict` here? To me, it seems we should be `strict`

paperqa/core.py
```diff
@@ -0,0 +1,112 @@
...
+    pattern = r'"(?:[^"\\]|\\.)*"'
+    text = re.sub(pattern, replace_newlines, text)
+
+    return json.loads(text)
+
+
+async def map_fxn_summary(
+    text: Text,
+    question: str,
+    chain: Chain | None,
```
jamesbraza, line 38: Wdyt of removing the `Chain` class? Imo we're using a term from LangChain that is really just a prompt with some pre or post processing

Maybe I am wrong here, but I feel like the term "chain" falls in the same bucket as "text-to-action" -- not really that useful of a term, more of a buzzword

paperqa/core.py
```diff
@@ -0,0 +1,112 @@
...
+                    else get_score(context)
+                )
+                # just in case question was present
+                result_data.pop("question", None)
+                extras = result_data
+            except KeyError:
+                success = False
+    else:
+        context = text.text
+        score = 5
```
jamesbraza, line 95: Can you move this `5` to a module level variable, and explain its existence with a code comment

I feel like this is some tribal knowledge we need to remove, or move to `settings`. Ideally we can actually get rid of it entirely

paperqa/docs.py
```diff
@@ -580,344 +580,222 @@ def delete(
...
-            _k = k * 10  # heuristic - get enough so we can downselect
-        if disable_vector_search:
-            matches = self.texts
+
+        answer = Answer(question=query) if isinstance(query, str) else query
+
+        if len(self.docs) == 0 and len(self.texts_index) == 0:
+            return answer
+
+        _settings = get_settings(settings)
```
jamesbraza, line 630: Why do we have a private local variable name inside of a method? Imo we should just call it `settings`

paperqa/prompts.py
```diff
@@ -23,7 +23,7 @@
     '"I cannot answer."'
     "For each part of your answer, indicate which sources most support "
     "it via citation keys at the end of sentences, "
-    "like (Example2012Example pages 3-4). Only cite from the context "
+    "like {example_citation}. Only cite from the context "
```
jamesbraza, line 26: Can you now rename the variable name `qa_prompt` to `qa_prompt_template`, since that's what it is now

paperqa/types.py
```diff
@@ -39,17 +32,44 @@
...
+        cvar_answer_id.reset(token)
+
 
 class LLMResult(BaseModel):
-    """A class to hold the result of a LLM completion."""
+    """A class to hold the result of a LLM completion.
+
+    To associate a group of LLMResults, you can use the `set_llm_answer_ids` context manager:
+
+    ```
```
jamesbraza, line 58:
```suggestion
    ```python
```

paperqa/utils.py
```diff
@@ -122,6 +121,37 @@ def strip_citations(text: str) -> str:
     return re.sub(citation_regex, "", text, flags=re.MULTILINE)
 
 
+def get_score(text: str) -> int:
```
jamesbraza, line 124: Can you rename this `extract_score` or add a brief docstring

paperqa/utils.py
```diff
@@ -185,7 +215,7 @@ def flatten(iteratble: list) -> list:
 
 def get_loop() -> asyncio.AbstractEventLoop:
     try:
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()  # new in python 3.10
```
jamesbraza, line 218:
```suggestion
        loop = asyncio.get_running_loop()
```

Feel free to ignore this PR comment, but imo this code comment is just distracting

tests/conftest.py
```diff
@@ -56,20 +59,32 @@ def agent_index_dir(agent_home_dir: Path) -> Path:
...
 
 @pytest.fixture
-def stub_data_dir() -> Path:
-    return Path(__file__).parent / "stub_data"
+def agent_test_settings(agent_index_dir: Path, stub_data_dir: Path) -> Settings:
+    settings = Settings()
+    settings.agent.paper_directory = stub_data_dir
+    settings.agent.index_directory = agent_index_dir
+    settings.agent.search_count = 2
+    settings.embedding = "sparse"
```
jamesbraza, line 72: Can we configure this in the instantiation of `Settings(embedding="sparse")`?

tests/test_configs.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_get_settings_missing_file(mocker):
+    mocker.patch("importlib.resources.files", side_effect=FileNotFoundError)
+    with pytest.raises(FileNotFoundError):
+        get_settings("missing_config")
+
+
+def test_settings_default_instantiation():
+    settings = Settings()
+    assert "gpt-4o-2024-08-06" in settings.llm
```
jamesbraza, line 47:
```suggestion
    assert "gpt-4o" in settings.llm
```

paperqa/config.py
```diff
@@ -0,0 +1,427 @@
...
+        answer_tool_name = GenerateAnswerTool.__fields__["name"].default
+        if answer_tool_name not in v:
+            raise ValueError(
+                f"If using an override, must contain at least the {answer_tool_name}."
+            )
+        return v
+
+
+class Settings(BaseSettings):
+    model_config = model_config = SettingsConfigDict(
```
jamesbraza, line 309:
```suggestion
    model_config = SettingsConfigDict(
```

whitead:
Ready for review - the PR is getting too sprawling to keep going.

TODO list [for future PRs]

- [ ] Get CLI tests back
- [ ] Make it possible to build_index from CLI
- [ ] Add litellm logging to configure_cli_logging
- [ ] Fix search query - if you try to search local index, it find the question field missing (CC @mskarlin - not sure if this was working)
- [ ] Remove tiktoken now that we have litellm
- [ ] Track costs using litellm instead of our list (CC @jamesbraza has code somewhere for this)
- [ ] Switch to Aviary env with baseline agent

paperqa/config.py
mskarlin: IMO this is up to the users cost tolerance. So we should also conditions our defaults on being cheaper models.

jamesbraza (approved):

jamesbraza (approved):
Sorry I realized I didn't have my text in the approval text box

I was gonna say, really excellent work here both from Andrew and Mike, LGTM

tests/conftest.py
whitead: Not opposed - but why?

paperqa/prompts.py
whitead: Think prompt and template are synonyms in our code base

paperqa/core.py
whitead: Discussed in person - no better name for closure for executing LLM + prompt + config

paperqa/contrib/zotero.py
whitead: black junk - fixed

paperqa/config.py
whitead: Would rather leave in place - there may be another larger model soon that is too expensive to be in all three places

paperqa/config.py
whitead: yea - just trying to make it reasonable default for people using the tool without config

paperqa/core.py
whitead: Added an explanation

PR #293, opened 2024-06-25
Fixed sonnet json formatting issue

### LINKED ISSUE #292: Parsing JSON with newlines
If the model response looks like:

```
{
  "summary": "The excerpt discusses different approaches to representing protein structures for machine learning, including equivariant methods. It mentions that scalar quantities of proteins like interaction interfaces are intrinsically rotation and translation invariant. The text describes several equivariant approaches:

1. Tensor Field Networks (TFN) using rotation equivariant convolutions based on spherical harmonics
2. Cormorant networks improving on TFN with Clebsch-Gordan non-linearity 
3. Geometric vector perceptron (GVP) using linear operations on vectors respecting rotation equivariance
4. Equivariant graph neural networks

The excerpt notes that TFN architectures have been successful in predicting protein-protein complex quality. However, it does not explicitly state that equivariant layers are definitively better at modeling protein structure compared to other approaches.",
  "relevance_score": 7
}
```

we do not parse it correctly. We should be able to handle that.

whitead:
Fixes #292

### DIFF
--- a/paperqa/utils.py
+++ b/paperqa/utils.py
@@ -184,4 +184,12 @@ def llm_read_json(text: str) -> dict:
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
+
+    # escape new lines within strings
+    def replace_newlines(match):
+        return match.group(0).replace("\n", "\\n")
+
+    pattern = r'"(?:[^"\\]|\\.)*"'
+    text = re.sub(pattern, replace_newlines, text)
+
     return json.loads(text)
--- a/paperqa/utils.py
+++ b/paperqa/utils.py
@@ -186,9 +186,12 @@ def llm_read_json(text: str) -> dict:
     text = text.rsplit("}", 1)[0] + "}"
 
     # escape new lines within strings
-    def replace_newlines(match):
+    def replace_newlines(match: re.Match) -> str:
         return match.group(0).replace("\n", "\\n")
 
+    # Match anything between double quotes
+    # including escaped quotes and other escaped characters.
+    # https://regex101.com/r/VFcDmB/1
     pattern = r'"(?:[^"\\]|\\.)*"'
     text = re.sub(pattern, replace_newlines, text)
 
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import pickle
 import tempfile
+import textwrap
 from io import BytesIO
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


paperqa/utils.py
```diff
@@ -184,4 +184,12 @@ def llm_read_json(text: str) -> dict:
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
+
+    # escape new lines within strings
+    def replace_newlines(match):
+        return match.group(0).replace("\n", "\\n")
+
+    pattern = r'"(?:[^"\\]|\\.)*"'
+    text = re.sub(pattern, replace_newlines, text)
```
jamesbraza, line 193: TIL `repl` arg can be a function

paperqa/utils.py
```diff
@@ -184,4 +184,12 @@ def llm_read_json(text: str) -> dict:
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
+
+    # escape new lines within strings
+    def replace_newlines(match):
+        return match.group(0).replace("\n", "\\n")
+
+    pattern = r'"(?:[^"\\]|\\.)*"'
```
jamesbraza, line 192: Bonus points for regex 101 link

paperqa/utils.py
```diff
@@ -184,4 +184,12 @@ def llm_read_json(text: str) -> dict:
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
+
+    # escape new lines within strings
+    def replace_newlines(match):
+        return match.group(0).replace("\n", "\\n")
```
jamesbraza, line 190: Is there a latent `NoneType` error here?

paperqa/utils.py
```diff
@@ -184,4 +184,12 @@ def llm_read_json(text: str) -> dict:
     text = "{" + text.split("{", 1)[-1]
     # split anything after the last }
     text = text.rsplit("}", 1)[0] + "}"
+
+    # escape new lines within strings
```
jamesbraza, line 188: What do you think of [`re.escape`](https://docs.python.org/3/library/re.html#re.escape)?

tests/test_paperqa.py
```diff
@@ -457,6 +458,22 @@ def test_llm_read_json(example: str):
     assert llm_read_json(example) == {"example": "json"}
 
 
+def test_llm_read_json_newlines():
```
jamesbraza, line 461: Can you add a docstring or comment that newlines are an edge case that broke us

jamesbraza (approved):
LGTM, added some comments

paperqa/utils.py
whitead: Added a type - I don't think sub will call on None matches

tests/test_paperqa.py
whitead: Thanks! Done

paperqa/utils.py
whitead: Don't see the connection

paperqa/utils.py
whitead: Added!

PR #297, opened 2024-06-26
Fixing `pre` missing a space

jamesbraza:
Before (notice missing whitespace after `information:`):

```none
Extra background information:The degron index is a predictive tool used to identify degradation ...
```

After:

```none
Extra background information:
The degron index is a predictive tool used to identify degradation ...
```


paperqa/docs.py
```diff
@@ -808,9 +808,7 @@ async def aquery(  # noqa: C901, PLR0912, PLR0915
             pre.answer_id = answer.id
             await self.llm_result_callback(pre)
             answer.add_tokens(pre)
-            answer.context = (
-                answer.context + "\n\nExtra background information:" + str(pre)
-            )
+            answer.context += f"\n\nExtra background information:\n{pre}"
```
whitead, line 811: There is not a newline because that is how the other contexts are presented. Don't think we want to change that. 

```suggestion
            answer.context += f"\n\nExtra background information: {pre}"
```

whitead (approved):
Use space instead of newline imo - so it matches rest of context.

PR #300, opened 2024-06-27
Enhanced `mypy` and Ruff 0.5

jamesbraza:
More tools


whitead (approved):
Thanks!

paperqa/docs.py
```diff
@@ -525,12 +527,12 @@ async def adoc_match(
...
             return set()
         # this only works for gpt-4 (in my testing)
-        try:
+        with contextlib.suppress(AttributeError):
             if (
                 rerank is None
                 and (
                     isinstance(self.llm_model, OpenAILLMModel)
-                    and self.config["model"].startswith("gpt-4")
+                    and self.llm_model.config["model"].startswith("gpt-4")
```
jamesbraza, line 535: @whitead FYI previously this always hit an `AttributeError`, so we never actually took this branch.

Now we can

PR #301, opened 2024-06-27
Moved to `setuptools_scm` for versioning from `git tag`

jamesbraza:
This PR moves us to use [pypa/setuptools_scm](https://github.com/pypa/setuptools_scm) for version management, and creation of a version.py file


whitead (approved):
LGTM

PR #306, opened 2024-08-12
Add Client (external API) Module For Enhanced Metadata

mskarlin:
This adds a new module: `paperqa.clients` which provides base classes for both providing (`MetadataProvider`) and processing (`MetadataPostProcessor`) `Doc` metadata. Two metadata providers are included: `CrossrefProvider` and `SemanticScholarProvider` and one processor: `JournalQualityPostProcessor` (which adds in a quality score for each extracted journal). `DocMetadataClient` then orchestrates all the usage of these providers, so a user doesn't need to worry about where metadata comes from, they just need to request it. Usage examples make this simpler: 

```python
async with aiohttp.ClientSession() as session:
    client = DocMetadataClient(session)
    doc_details = await client.query(title="PaperQA: Retrieval-Augmented Generative Agent for Scientific Research")
```

The doc_details (a `DocDetails` object) is a subclass of a `Doc` with extra metadata and the ability to be merged with the `__add__` operator. In this way, users can use whichever `MetadataProvider`'s they want and they will get a resultant `DocDetails` object which fills in whatever fields they can get. `DocDetails` have rich metadata like: 

```
print(doc_details.formatted_citation) 
>>>Jakub L'ala, Odhran O'Donoghue, Aleksandar Shtedritski, Sam Cox, Samuel G. Rodriques, and Andrew D. White. Paperqa: retrieval-augmented generative agent for scientific research. ArXiv, Dec 2023. doi:10.48550/arxiv.2312.07559. This article has 21 citations.
```
The `DocMetadataClient.query` method will accept any valid `ClientQuery` supported by the `MetadataProviders` you have selected (by default all). But we've only implemented a `DOIQuery` and a `TitleAuthorQuery` for now. It will return all fields by default, but you can also filter for whatever you'd like: 

```python
async with aiohttp.ClientSession() as session:
        # now get with authors just from one source
        s2_client = DocMetadataClient(session, clients=[SemanticScholarProvider])
        s2_details = await s2_client.query(
            title="Augmenting large language models with chemistry tools",
            fields=["title", "doi", "authors"],
        )
```

The above example will only query SemanticScholar (not Crossref), and it'll only return the title, doi, and authors for that paper. This is useful if you need to a small lookup for DOI or something along those lines. The processors run by default, but if you don't specify them in the `clients` argument they won't run, like in the above example. So to run with `JournalQualityPostProcessor` enabled: 

```python
async with aiohttp.ClientSession() as session:
        # now get with authors just from one source
        s2_client = DocMetadataClient(session, clients=[SemanticScholarProvider, JournalQualityPostProcessor])
        s2_details = await s2_client.query(
            title="Augmenting large language models with chemistry tools",
            fields=["title", "doi", "authors", "journal"],
        )
```

Note we had to add `"journal"` as a field, otherwise the `JournalQualityPostProcessor` would have had no journal to work from. 

The metadata client has been added into `Docs.aadd` method -- so if you set the `use_doc_details` flag to true, your other inputs will be used to get rich metadata using a `DocDetails` object rather than the standard `Doc` object. 

Tests are included which use the `pytest-vcr` module and cassettes to avoid needing network access. Secrets are filtered from the cassettes by default.


tests/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from dotenv import load_dotenv
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _load_env():
+    load_dotenv()
+
+
+@pytest.fixture(scope="module")
```
jamesbraza, line 10: Wdyt of changing to `session`?

tests/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+def vcr_config():
+    return {
+        "filter_headers": ["Crossref-Plus-API-Token", "x-api-key"],
```
jamesbraza, line 13: Can we move these to module level variables?

paperqa/utils.py
```diff
@@ -192,3 +195,171 @@ def replace_newlines(match: re.Match) -> str:
...
+    key: str | None = None,
+    clean: bool = True,
+    missing_replacements: dict[str, str] | None = None,
+) -> str:
+    """Transform bibtex entry into a citation, potentially adding missing fields."""
+    # WOWOW This is hard to use
+    from pybtex.database import Person, parse_string
+    from pybtex.database.input.bibtex import Parser
+    from pybtex.style.formatting import unsrtalpha
+    from pybtex.style.template import FieldIsMissing
```
jamesbraza, line 275: We can move these to module level now that `pybtex` is first class

paperqa/utils.py
```diff
@@ -192,3 +195,171 @@ def replace_newlines(match: re.Match) -> str:
...
+        "Patent": "patent",
+        "Preprint": "article",
+        "Report": "techreport",
+        "Thesis": "phdthesis",
+        "WebPage": "misc",
+        "Plain": "article",
+    }
+    if "@None" in bibtex:
+        return bibtex.replace("@None", "@article")
+    # new format check
```
jamesbraza, line 247: Can we remove this comment?

.pre-commit-config.yaml
```diff
@@ -5,6 +5,7 @@ repos:
     rev: v4.6.0
     hooks:
       - id: check-added-large-files
+        args: ["--maxkb=1500"]
```
jamesbraza, line 8: Jw, why do this? For the cassettes? You can alternately use `exclude`

mskarlin: This is for the journal quality data -- I can exclude it specifically too.

mskarlin: done!

paperqa/clients/crossref.py
```diff
@@ -0,0 +1,331 @@
...
+        # TODO: we keep this for robustness, but likely not needed anymore,
+        # since we now create the bibtex from scratch
+        if query_bibtex:
+            bibtex = await doi_to_bibtex(
+                message["DOI"], session, missing_replacements=fallback_data  # type: ignore[arg-type]
+            )
+            # track the origin of the bibtex entry for debugging
+            bibtex_source = "crossref"
+
+    except DOINotFoundError:
```
jamesbraza, line 149: Can you just add an inline comment saying why this suppression is okay

paperqa/clients/crossref.py
```diff
@@ -0,0 +1,331 @@
...
+    if "published" in message and "date-parts" in message["published"]:
+        date_parts = message["published"]["date-parts"][0]
+        if len(date_parts) >= 3:  # noqa: PLR2004
+            publication_date = datetime(date_parts[0], date_parts[1], date_parts[2])
+        elif len(date_parts) == 2:  # noqa: PLR2004
+            publication_date = datetime(date_parts[0], date_parts[1], 1)
+        elif len(date_parts) == 1:
+            publication_date = datetime(date_parts[0], 1, 1)
+
+    doc_details = DocDetails(  # type: ignore[call-arg]
```
jamesbraza, line 167: What do you think of making a `classmethod` of `DocDetails` for this?

paperqa/clients/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class DOINotFoundError(Exception):
```
jamesbraza, line 1: Any chance we can just inline this somewhere? Imo it's not worth its own module

Tho I can also see why it should go in its own module, so feel free to keep here

paperqa/clients/journal_quality.py
```diff
@@ -0,0 +1,56 @@
...
+    def query_creator(self, doc_details: DocDetails, **kwargs) -> JournalQuery | None:
+        try:
+            return JournalQuery(journal=doc_details.journal, **kwargs)
+        except ValidationError:
+            logger.debug(
+                "Must have a valid journal name to query journal quality data."
+            )
+            return None
+
+    # TODO: refresh script for journal quality data
```
jamesbraza, line 56: Can you just edit this TODO? I am wondering why a "refresh script" would be in the `JournalQualityPostProcessor` class

paperqa/clients/crossref.py
```diff
@@ -0,0 +1,331 @@
...
+    "issue": "issue",
+    "publisher": "publisher",
+    "ISSN": "issn",
+    "page": "pages",
+    "container-title": "journal",
+    "URL": "url",
+    "bibtex": "bibtex",
+    "is-referenced-by-count": "citation_count",
+}
+CROSSREF_CONTENT_TYPE_TO_BIBTEX_MAPPING = {
```
jamesbraza, line 39:
```suggestion
CROSSREF_CONTENT_TYPE_TO_BIBTEX_MAPPING: dict[str, str] = {
```

paperqa/clients/semantic_scholar.py
```diff
@@ -0,0 +1,310 @@
...
+async def parse_s2_to_doc_details(
+    paper_data: dict, session: aiohttp.ClientSession
+) -> DocDetails:
+
+    bibtex_source = "self_generated"
+
+    if "data" in paper_data:
+        paper_data = paper_data["data"][0]
+
+    if "ArXiv" in paper_data["externalIds"]:
```
jamesbraza, line 132: Do you mind moving `"ArXiv"` to a module level variable? Or you like it here

paperqa/clients/semantic_scholar.py
```diff
@@ -0,0 +1,310 @@
...
+    # Note: we expect the number of authors to be possibly different
+    return any(
+        starmap(
+            lambda x, y: x in y or y in x,
+            zip(s2_authors_noinit, authors_noinit, strict=False),
+        )
+    )
+
+
+async def parse_s2_to_doc_details(
```
jamesbraza, line 123: What do you think of making this a `classmethod` of `DocDetails`?

paperqa/clients/client_models.py
```diff
@@ -0,0 +1,127 @@
...
+
+class MetadataProvider(ABC, Generic[ClientQueryType]):
+    """Provide metadata from a query by any means necessary."""
+
+    async def query(self, query: dict) -> DocDetails | None:
+        return await self._query(self.query_transformer(query))
+
+    @abstractmethod
+    async def _query(self, query: ClientQueryType) -> DocDetails | None:
+        pass
```
jamesbraza, line 82:
```python
    async def query(self, query: dict) -> DocDetails | None:
        try:
            return await self._query(self.query_transformer(query))
        except DOINotFoundError:
            logger.exception(...)
            return None

    @abstractmethod
    async def _query(self, query: ClientQueryType) -> DocDetails:
        """
        Some summary.

        Raises:
            DOINotFoundError: This exception type when ....
        """
```

It seems all the subclasses suppress `DOINotFoundError` with a `logger.exception`.

What do you think of doing that in the base class, to simplify children?

Also, can you make a docstring for `query` or `_query` explaining when to return `None` vs `DocDetails`?

paperqa/clients/exceptions.py
mskarlin: yea I'm gonna keep because I think we're going to add more here

paperqa/clients/crossref.py
mskarlin: For this (and the semantic_scholar one) I was kind of thinking that it's the `MetadataProvider`'s job to deal with parsing. This means that to add a new provider, you just make the class and it gives you back a `DocDetails` object, with all of its logic encapsulated. If we move this logic into a `DocDetails` method, I think it implies that it could be accessed outside of the `clients` module which this re-factor is trying to remove as a possibility. LMK what you think on that.

jamesbraza: Okay I gotchu and sounds good, thanks!

paperqa/clients/semantic_scholar.py
mskarlin: I don't think this has much need outside of this parsing, so I think a module level variable could be more confusing.

paperqa/clients/semantic_scholar.py
mskarlin: we dealt w. this in the comment for the crossref one.

paperqa/clients/journal_quality.py
mskarlin: I'll move it, just a reminder for the future

paperqa/clients/client_models.py
mskarlin: This will only apply for the semantic_scholar/crossref type metadata services, so we could put an extra abstract method in `DOIOrTitleBasedProvider` which _query then runs. 

100% on adding some docstrings.

tests/test_clients.py
```diff
@@ -0,0 +1,415 @@
...
+)
+@pytest.mark.asyncio()
+async def test_doi_search(paper_attributes: dict[str, str]):
+    async with aiohttp.ClientSession() as session:
+        client = DocMetadataClient(session)
+        details = await client.query(doi=paper_attributes["doi"])
+        assert set(details.other["client_source"]) == set(  # type: ignore[union-attr]
+            paper_attributes["source"]
+        ), "Should have the correct source"
+        assert details.key == paper_attributes["key"], "Should have the correct key"  # type: ignore[union-attr]
```
jamesbraza, line 179: You could do a `for` loop here over `paper_attributes.keys()` if you want

tests/test_clients.py
```diff
@@ -0,0 +1,415 @@
...
+        assert (
+            details.formatted_citation == paper_attributes["formatted_citation"]  # type: ignore[union-attr]
+        ), "Should have the correct formatted_citation"
+
+
+@pytest.mark.vcr()
+@pytest.mark.asyncio()
+async def test_bulk_doi_search():
+    dois = [
+        "10.1063/1.4938384",
```
jamesbraza, line 199: Bonus points for comments documenting why each of these were chosen

paperqa/utils.py
```diff
@@ -192,3 +199,164 @@ def replace_newlines(match: re.Match) -> str:
...
+    if missing_replacements is None:
+        missing_replacements = {}
+    try:
+        return bibtex.split(field + "={")[1].split("}")[0].split()[0].strip()
+    except IndexError:
+        return missing_replacements.get(field, "")
+
+
+def create_bibtex_key(author: list[str], year: str, title: str) -> str:
+    FORBIDDEN_KEY_CHARACTERS = {"_", " ", "-", "/", "'", "`", ":"}
```
jamesbraza, line 353: Wdyt of moving this to module level?

paperqa/clients/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    before_sleep_log,
+    retry,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_incrementing,
+)
+
+logger = logging.getLogger(__name__)
+
+TITLE_SET_SIMILARITY_THRESHOLD = 0.75
```
jamesbraza, line 21: Can you add a comment explaining units here?

paperqa/clients/utils.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import annotations
```
jamesbraza, line 1: Can we somehow remove this module? We already have `utils.py` one level up, can we just use that one?

paperqa/prompts.py
```diff
@@ -53,6 +53,14 @@
     "Citation:"
 )
 
+structured_citation_prompt = (
+    "Extract the title, authors, and doi as a JSON from this MLA citation. "
+    "If any field can not be found, return them as null. "
+    "Use title, authors, and doi as keys, author's value should be a list of authors. "
+    "{text}\n\n"
```
jamesbraza, line 60: Can you change this template var name from `text`? Maybe `instructions_suffix`

Also a nit is if it's an empty string, there will be extra spaces

paperqa/clients/utils.py
mskarlin: yea, i wanted this to truly be a submodule, but I suppose maybe we'd reuse this.

paperqa/clients/utils.py
mskarlin: i'm just cutting it for hard coding into the defaults in the respective methods, it's also in the model itself

paperqa/utils.py
mskarlin: I think this is specific to bibtex, no?

paperqa/prompts.py
mskarlin: renamed to "citation"

paperqa/prompts.py
```diff
@@ -53,6 +53,14 @@
     "Citation:"
 )
 
+structured_citation_prompt = (
+    "Extract the title, authors, and doi as a JSON from this MLA citation. "
+    "If any field can not be found, return them as null. "
```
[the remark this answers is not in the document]
mskarlin: great catch!

paperqa/docs.py
```diff
@@ -412,7 +438,48 @@ async def aadd(
...
+                    title = citation_title
+                if citation_doi := citation_json.get("doi"):
+                    doi = citation_doi
+                if citation_author := citation_json.get("authors"):
+                    authors = citation_author
+
+        # see if we can upgrade to DocDetails
+        # if not, we can progress with a normal Doc
+        # if "overwrite_fields_from_metadata" is used:
+        # will map "docname" to "key", and "dockey" to "doc_id"
```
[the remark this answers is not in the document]
mskarlin: I don't know if if we want to break backwards compatibility quite yet-- I'm for keeping and having "extra" duplicative fields.

paperqa/docs.py
```diff
@@ -435,7 +502,7 @@ def add_texts(
         loop = get_loop()
         return loop.run_until_complete(self.aadd_texts(texts, doc))
 
-    async def aadd_texts(self, texts: list[Text], doc: Doc) -> bool:
+    async def aadd_texts(self, texts: list[Text], doc: Doc | DocDetails) -> bool:
```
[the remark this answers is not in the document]
mskarlin: Yes it does! I'd originally had some `DocDetails` attribute usage here, but this is vestigial now.

whitead (approved):
So fucking cool - great work

PR #309, opened 2024-08-27
Agentic workflows, locally indexed search, and CLI

mskarlin:
## Description is still a **WIP**

This PR merges agentic workflows (via an OpenAI-based agent and a 'fake' agent) into a new branch to prep for release. The biggest new included feature is a new CLI tool which allows PaperQA to be used in its entirety from the command line. Papers are chosen to be added into the Docs objects via full-text search of indexed directories on your machine. After installing via `pip install paper-qa[agents]` (or if local, `pip install -e .[agents]`), you can navigate to a directory with some PDF/TXT/HTML documents in it and run: 

```
pqa ask "what is pyroptosis?"
```

Which, for the first time you run, an index will be created, but further runs will use that same index. If you change some PaperQA settings which would invalidate your current index (like changing the chunksize or your target directory), PaperQA will automatically create a new index for you. 

New options can be set like the following (with a relative directory): 
```
pqa set 'agent_tools.paper_directory' 'papers'
```

and then be viewed similarly
```
pqa show 'agent_tools.paper_directory'
```
Or to see everything you've assigned: 
```
pqa show all 
```

All the settings available to pqa can be found in the `QueryRequest` object within `agents/models/`, behind the scenes PaperQA will store a configuration yaml in your (by default) `~/.pqa`, folder but this can be changed via setting `PQA_HOME`.

Additional new stuff:
- In using the client library more heavily to query across 100s of papers, I implemented some minor bugfixes, like stripping `https://doi.org/` out of DOIs which often shows up when they are extracted by LLMs
- Optional install for a 'google' environment that includes vertex. It's a crazy large import that takes ~2-3 seconds on its own, so it's best to leave optional.


agent-requirements.txt
```diff
@@ -0,0 +1,9 @@
+anyio
```
jamesbraza, line 1: Can we just inline these into `pyproject.toml` as normal `extra`s? Imo we don't need these extra requirements files, they are just distracting

pyproject.toml
```diff
@@ -215,8 +217,11 @@ max-doc-length = 120  # Match line-length
 # defaults when analyzing docstring sections.
 convention = "google"
 
-[tool.setuptools.dynamic.optional-dependencies.dev]
-file = ["dev-requirements.txt"]
+[tool.setuptools.dynamic.optional-dependencies]
+agents = {file = ["agent-requirements.txt"]}
+all = {file = ["agent-requirements.txt", "dev-requirements.txt", "google-requirements.txt"]}
```
jamesbraza, line 222: I don't really like `all` extras, because one always has to look up what that means. I would rather see our CI specifying `pip install -e .[agents,google,dev]`

pyproject.toml
```diff
@@ -137,7 +140,6 @@ extend-select = [
     "PLR6201",
     "PLW0108",
     "RUF022",
-    "RUF025",
```
jamesbraza, line 140: Can you make this C420

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
         "filter_headers": [CROSSREF_HEADER_KEY, SEMANTIC_SCHOLAR_HEADER_KEY],
     }
+
+
+@pytest.fixture()
+def unique_tmp_subfolder(tmp_path):
+    yield tmp_path
+    # Cleanup after the test
+    if tmp_path.exists():
+        shutil.rmtree(tmp_path, ignore_errors=True)
```
jamesbraza, line 33: By accepting `tmp_path`, this fixture is really just more like a folder cleanup and is unrelated to tempfiles. Can you rename it?

Note there is a `pytest` plugin https://github.com/pytest-dev/pyfakefs for filesystem management, and it works great I used it a lot in the past

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
+def agent_module_dir(unique_tmp_subfolder, monkeypatch):
+    """Set up a unique temporary folder for the agent module."""
+    monkeypatch.setenv("PQA_HOME", str(unique_tmp_subfolder))
+    importlib.reload(paperqa.agents)
+    return unique_tmp_subfolder
+
+
+@pytest.fixture()
+def agent_index_dir(agent_module_dir):
+    return str(agent_module_dir / ".pqa/indexes")
```
jamesbraza, line 46: Can you use `pathlib` to work on other OS's

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
+@pytest.fixture()
+def agent_module_dir(unique_tmp_subfolder, monkeypatch):
+    """Set up a unique temporary folder for the agent module."""
+    monkeypatch.setenv("PQA_HOME", str(unique_tmp_subfolder))
+    importlib.reload(paperqa.agents)
+    return unique_tmp_subfolder
+
+
+@pytest.fixture()
+def agent_index_dir(agent_module_dir):
```
jamesbraza, line 45: Can you type hint the input args

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
+    # Cleanup after the test
+    if tmp_path.exists():
+        shutil.rmtree(tmp_path, ignore_errors=True)
+
+
+@pytest.fixture()
+def agent_module_dir(unique_tmp_subfolder, monkeypatch):
+    """Set up a unique temporary folder for the agent module."""
+    monkeypatch.setenv("PQA_HOME", str(unique_tmp_subfolder))
+    importlib.reload(paperqa.agents)
```
jamesbraza, line 40: Okay reading this it comes off as a bit hacky. I am left wondering things like:
- Why is `agent_module_dir` not just a static filepath to a Python module made by `pathlib.Path(__file__).parent`?
- How does `PQA_HOME` relates to `agent_module_dir`?
- Will `importlib.reload` have implications with multiprocessing or fixture ordering?

Also, are you open to not using `monkeypatch.setenv` here? Imo it's more intuitive if one `yield`s from within a context manager

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
+
+
+@pytest.fixture()
+def agent_index_dir(agent_module_dir):
+    return str(agent_module_dir / ".pqa/indexes")
+
+
+@pytest.fixture(name="agent_test_kit")
+def fixture_agent_test_kit() -> Answer:
+    return Answer(question="What is is a self-explanatory model?")
```
jamesbraza, line 51: I think I used the name "test kit" in the past because it gave a bunch of objects. This one is just an `Answer`, so can you name it like `stub_answer`?

tests/conftest.py
```diff
@@ -15,3 +23,45 @@ def vcr_config():
...
+
+
+@pytest.fixture(name="stub_paper_path_with_details", scope="session")
+def fixture_stub_paper_path_with_details() -> tuple[Path, dict]:
+    # Corresponds with https://www.semanticscholar.org/paper/A-Perspective-on-Explanations-of-Molecular-Models-Wellawatte-Gandhi/1db1bde653658ec9b30858ae14650b8f9c9d438b
+    return (
+        Path(__file__).parent / "paper.pdf",
+        {
```
jamesbraza, line 59: Wdyt of returning `PaperDetail`s object here instead of metadata?

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+try:
+    from paperqa.agents import ask, build_index, clear, search_query, set_setting, show
+    from paperqa.agents.models import AnswerResponse
+    from paperqa.agents.search import SearchIndex
+except ImportError:
+    SKIP_AGENT_TESTS = True
+
+if SKIP_AGENT_TESTS:
+    pytest.skip("agents module is not installed", allow_module_level=True)
```
jamesbraza, line 15:
```suggestion
try:
    from paperqa.agents import ask, build_index, clear, search_query, set_setting, show
    from paperqa.agents.models import AnswerResponse
    from paperqa.agents.search import SearchIndex
except ImportError:
    pytest.skip("agents module is not installed", allow_module_level=True)
```

Can you do this?

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+    assert show("parsing_configuration.ordered_parser_preferences") == [
+        "paperqa_default"
+    ], "List setting not properly set"
+
+
+@pytest.mark.asyncio()
+async def test_cli_show(agent_index_dir):
+
+    # make empty index
+    assert show("indexes") == [], "No indexes should be present"
```
jamesbraza, line 51:
```suggestion
    assert not show("indexes"), "No indexes should be present"
```

.gitignore
```diff
@@ -137,3 +137,5 @@ env
 # Matching pyproject.toml
 paperqa/version.py
 tests/example*
+tests/test_index/*
+.DS_Store
```
jamesbraza, line 141: We should just generate this file:

```
curl -s \
  https://raw.githubusercontent.com/github/gitignore/master/\{Global/Vim,Global/JetBrains,Global/VisualStudioCode,Global/macOS,Python\}.gitignore \
  > .gitignore
```

Then copy over the paper-qa specifics

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+    llm_model_factory,
+)
+
+from ..version import __version__
+from .prompts import STATIC_PROMPTS
+
+logger = logging.getLogger(__name__)
+
+
+class SupportsPickle(Protocol):
```
jamesbraza, line 45: Can you add a docstring for this?

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+                **c.text.model_dump(exclude={"text", "embedding", "doc"}),
+                doc=paperqa.Doc(**c.text.doc.model_dump(exclude={"embedding"})),
+            ),
+        )
+        for c in answer.contexts
+    ]
+    answer.contexts = new_contexts
+
+
+class AgentStatus(str, Enum):
```
jamesbraza, line 68: What is our Python version range? Can we use `enum.StrEnum`?

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+        "\n- Collect papers cited by previous evidence (preferred)"
+        "\n- Gather more evidence using a different phrase"
+        "\nIf you search for more papers or collect new papers cited by previous evidence,"
+        " remember to gather evidence again."
+        " Once you have five or more pieces of evidence from multiple sources, or you have tried a few times, "
+        "call {gen_answer_tool_name} tool. The {gen_answer_tool_name} tool output is visible to the user, "
+        "so you do not need to restate the answer and can simply terminate if the answer looks sufficient. "
+        "The current status of evidence/papers/cost is {status}"
+    )
+    paper_directory: str = Field(
```
jamesbraza, line 107: For the directories, let's always have `str | os.PathLike` to allow for `pathlib` users

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+        default=None,
+        description=(
+            "Optional manifest CSV, containing columns which are attributes for a DocDetails object. "
+            "Only 'file_location','doi', and 'title' will be used when indexing."
+        ),
+    )
+    search_count: int = 8
+    wipe_context_on_answer_failure: bool = True
+    timeout: float = Field(
+        default=500,
```
jamesbraza, line 129:
```suggestion
        default=500.0,
```

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+    )
+    should_pre_search: bool = Field(
+        default=False,
+        description="If set to true, run the search tool before invoking agent.",
+    )
+    agent_config: dict[str, Any] | None = Field(
+        default=None,
+        description="Optional keyword argument configuration for the agent.",
+    )
+    tool_names: set[str] | list[str] | None = Field(
```
jamesbraza, line 143: Can we rethink this type hint? I think we should use `Sequence[str]`

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+class ParsingOptions(str, Enum):
+    PAPERQA_DEFAULT = "paperqa_default"
+
+    def available_for_inference(self) -> list[ParsingOptions]:
+        return [self.PAPERQA_DEFAULT]  # type: ignore[list-item]
+
+    def get_parse_type(self, config: ParsingConfiguration) -> str:
+        if self == ParsingOptions.PAPERQA_DEFAULT:
+            return config.parser_version_string
+        return None  # type: ignore[unreachable]
```
jamesbraza, line 177: Let's use `typing.assert_never` here

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+    def chunk_type(self, chunking_selection: ChunkingOptions | None = None) -> str:
+        """Future chunking implementations (i.e. by section) will get an elif clause here."""
+        if chunking_selection is None:
+            chunking_selection = self.chunking_algorithm
+        if chunking_selection == ChunkingOptions.SIMPLE_OVERLAP:
+            return (
+                f"{self.parser_version_string}|{chunking_selection.value}"
+                f"|tokens={self.chunksize}|overlap={self.overlap}"
+            )
+        return None  # type: ignore[unreachable]
```
jamesbraza, line 216: Let's use `typing.assert_never` here

paperqa/agents/models.py
```diff
@@ -0,0 +1,532 @@
...
+    # Should we filter out "Extra Background Information" citations
+    # which come from pre-step in paper-qa algorithm
+    filter_extra_background: bool = True
+    # provides post-hoc linkage of request to a docs object
+    # NOTE: this isn't a unique field, on the user to keep straight
+    _docs_name: str | None = PrivateAttr(default=None)
+
+    # strict validation for now
+    class Config:
+        extra = "forbid"
```
jamesbraza, line 281: This is Pydantic v1 syntax, can you use Pydantic v2 syntax?

tests/conftest.py
mskarlin: `agent_module_dir` is a Path (added typing), and the str() casting should default to the correct format for whichever OS it runs on.

pyproject.toml
mskarlin: I vote to leave it for convenience, but we can specify in CI

paperqa/agents/models.py
jamesbraza: The only thing there is for something like "fake agent" tools may be ordered

paperqa/agents/chains.py
```diff
@@ -0,0 +1,61 @@
...
+from openai import AsyncOpenAI
+
+from paperqa import OpenAILLMModel
+
+from .tools import get_year
+
+logger = logging.getLogger(__name__)
+
+
+async def search_chain(
```
jamesbraza, line 15: Maybe this is my opinion, but can we not call them "chains"? It's just one LLM prompt, right?

I know that's a LangChain thing, but tbh it's just an API call with some pre/post processing

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+from ..llms import LangchainLLMModel
+from .models import (
+    QueryRequest,
+)
+
+USE_VERTEX = True
+try:
+    from langchain_google_vertexai import ChatVertexAI, HarmBlockThreshold, HarmCategory
+except ImportError:
+    USE_VERTEX = False
```
jamesbraza, line 28:
```suggestion
try:
    from langchain_google_vertexai import ChatVertexAI, HarmBlockThreshold, HarmCategory
    USE_VERTEX = True
except ImportError:
    USE_VERTEX = False
```

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+    # https://docs.anyscale.com/endpoints/text-generation/query-a-model
+    "meta-llama/Meta-Llama-3-8B-Instruct": (0.15 / 10**6, 0.15 / 10**6),
+    "meta-llama/Meta-Llama-3-70B-Instruct": (1.0 / 10**6, 1.0 / 10**6),
+    "mistralai/Mistral-7B-Instruct-v0.1": (0.15 / 10**6, 0.15 / 10**6),
+    "mistralai/Mixtral-8x7B-Instruct-v0.1": (1.0 / 10**6, 1.0 / 10**6),
+    "mistralai/Mixtral-8x22B-Instruct-v0.1": (1.0 / 10**6, 1.0 / 10**6),
+}
+
+
+def get_llm_name(llm: Any) -> str:
```
jamesbraza, line 69: What do you think of deleting this one? I think it's not useful because callers should know the type

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+def get_cost(model: str, tokens: int, is_completion: bool) -> float:
+    if model in costs:  # Prefer our internal costs model
+        model_costs: tuple[float, float] = costs[model]
+    else:
+        logger.warning(f"Model {model} not found in costs.")
+        return 0.0
+    return tokens * model_costs[int(is_completion)]
+
+
+def compute_cost(token_counts: dict[str, list[int]]) -> float:
```
jamesbraza, line 87: The names `get_cost` and `compute_cost` are so similar, but they do different things. Can we rename them to be more intuitive:
- `get_cost` to `compute_cost`
- `compute_cost` to `bulk_compute_cost`

pyproject.toml
jamesbraza: The thing is, it prevents us from just removing the requirements files entirely and directly writing dependencies in `pyproject.toml`. Also, all the requirements bloat the repo imo

paperqa/types.py
```diff
@@ -360,6 +361,7 @@ class DocDetails(Doc):
     doi: str | None = None
     doi_url: str | None = None
     doc_id: str | None = None
+    file_location: str | None = None
```
jamesbraza, line 364: Can you make this `str | os.PathLike`

paperqa/types.py
```diff
@@ -375,6 +377,10 @@ def clean_key(cls, value: str) -> str:
     @staticmethod
     def lowercase_doi_and_populate_doc_id(data: dict[str, Any]) -> dict[str, Any]:
         if doi := data.get("doi"):
+            remove_urls = ["https://doi.org/", "http://dx.doi.org/"]
```
jamesbraza, line 380: Can you move this to a module level variable or class variable, with a descriptive name for these two URLs?

paperqa/types.py
```diff
@@ -426,6 +432,16 @@ def inject_clean_doi_url_into_data(data: dict[str, Any]) -> dict[str, Any]:
 
         return data
 
+    @staticmethod
+    def remove_invalid_authors(data: dict[str, Any]) -> dict[str, Any]:
+        """Capture and cull strange author names."""
+        if authors := data.get("authors"):
+            data["authors"] = [
+                a for a in authors if a.lower() not in {"et al", "et al."}
```
jamesbraza, line 440: Can you move the `et al` to a module level or class variable with a descriptive name:

```python
INVALID_AUTHORS_NAMES = {"et al", "et al."}  # We exact match these
```

paperqa/types.py
```diff
@@ -477,10 +493,13 @@ def populate_bibtex_key_citation(  # noqa: C901, PLR0912
...
-
-                existing_entry = next(
-                    iter(Parser().parse_string(data["bibtex"]).entries.values())
-                )
+                try:
+                    existing_entry = next(
+                        iter(Parser().parse_string(data["bibtex"]).entries.values())
+                    )
+                except PybtexSyntaxError:
+                    logger.warning(f"Failed to parse bibtex for {data['bibtex']}")
```
jamesbraza, line 501:
```suggestion
                    logger.warning(f"Failed to parse bibtex for {data['bibtex']}.")
```

Periods help suss out empty string errors

paperqa/utils.py
```diff
@@ -419,3 +418,13 @@ async def _get_with_retrying(
 
 def union_collections_to_ordered_list(collections: Iterable) -> list:
     return sorted(reduce(lambda x, y: set(x) | set(y), collections))
+
+
+def pqa_directory(name: str) -> str:
```
jamesbraza, line 423: Are you open to returning `Path` instead of `str`? It's much easier to work with `pathlib`, especially in context of opening files or extending the path

tests/test_manifest.csv
jamesbraza, line 1: This is not a unit test, but stub data. Can you call it "stub_manifest"

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    assert not caplog.records, "Indexing should not be triggered again"
+
+    # now we want to change the settings
+    set_setting("embedding", "sparse")
+
+    # running again should now re-trigger an indexing
+    _ = build_index(
+        directory=str(Path(__file__).parent), index_directory=agent_index_dir
```
jamesbraza, line 161: Can we make `build_index` support `os.PathLike` so we don't need these `str` casts

paperqa/agents/chains.py
mskarlin: yea, legacy func for sure

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+    for record in caplog.records:
+        if "Metadata not found" in record.msg:
+            raise AssertionError(
+                "Indexing logs should not be present after search starts"
+            )
+
+    caplog.clear()
+
+    # running again should not trigger any indexing
+    _ = build_index(
```
jamesbraza, line 150:
```suggestion
    build_index(
```

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+def test_cli_index(agent_index_dir, caplog):
+
+    _ = build_index(
+        directory=str(Path(__file__).parent), index_directory=agent_index_dir
+    )
+
+    caplog.clear()
+
+    _ = ask(
```
jamesbraza, line 134:
```suggestion
    ask(
```

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+    answer = answers[0][0]
+    assert isinstance(answer, AnswerResponse), "Answer not properly returned"
+    assert (
+        "I cannot answer" not in answer.answer.answer
+    ), "An answer should be generated."
+    assert len(answer.answer.context) >= 1, "No contexts were found."
+    assert len(show("answers")) == 1, "An answer should be returned"
+
+
+def test_cli_index(agent_index_dir, caplog):
```
jamesbraza, line 126: Can you type hint the input dir arg

tests/test_cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+    _ = build_index(
```
jamesbraza, line 128:
```suggestion
    build_index(
```

paperqa/agents/docs.py
mskarlin: for sure, it's unused anyway

paperqa/docs.py
```diff
@@ -450,7 +451,10 @@ async def aadd(  # noqa: C901, PLR0912, PLR0915
             )
             chain_result = await structured_cite_chain({"citation": citation}, None)
             with contextlib.suppress(json.JSONDecodeError):
-                citation_json = json.loads(chain_result.text)
+                clean_text = chain_result.text.strip("`")
```
jamesbraza, line 454: Do you want `removesuffix`? More just checking as `strip` is applied everywhere

paperqa/llms.py
```diff
@@ -887,7 +887,7 @@ def llm_model_factory(llm: str) -> LLMModel:
     if llm != "default":
         if is_openai_model(llm):
             return OpenAILLMModel(config={"model": llm})
-        elif llm.startswith("langchain"):  # noqa: RET505
+        elif llm.startswith("langchain") or "gemini" in llm:  # noqa: RET505
```
jamesbraza, line 890:
```suggestion
        if llm.startswith("langchain") or "gemini" in llm:
```

tests/test_agents.py
```diff
@@ -0,0 +1,648 @@
...
+from pydantic import ValidationError
+from pytest_subtests import SubTests
+
+import paperqa
+from paperqa.docs import Docs
+from paperqa.llms import LangchainLLMModel
+from paperqa.types import Answer, Context, Doc, PromptCollection, Text
+from paperqa.utils import get_year
+
+SKIP_AGENT_TESTS = False
```
jamesbraza, line 22: Can you move this to within the `try` and just before the `except`

tests/test_agents.py
```diff
@@ -0,0 +1,648 @@
...
+        PaperSearchTool,
+        SharedToolState,
+    )
+except ImportError:
+    SKIP_AGENT_TESTS = True
+
+if SKIP_AGENT_TESTS:
+    pytest.skip("agents module is not installed", allow_module_level=True)
+
+PAPER_DIRECTORY = str(Path(__file__).parent)
```
jamesbraza, line 58:
```suggestion
PAPER_DIRECTORY = Path(__file__).parent
```

Just a few lines below this it seems we use `anyio.Path`. Let's just stick with `Path` the entire time

tests/test_agents.py
```diff
@@ -0,0 +1,648 @@
...
+        index_name="pqa_index_0",
+        index_directory=agent_index_dir,
+    )
+    assert index.fields == [
+        "title",
+        "file_location",
+        "body",
+        "year",
+    ], "Incorrect fields in index"
+    assert len(await index.index_files) == 4, "Incorrect number of index files"
```
jamesbraza, line 74: Maybe it's just me, but why are there 4 index files? Can you expand the assertion string to mention why 4

tests/test_agents.py
```diff
@@ -0,0 +1,648 @@
...
+    results = await index.query(query="who is Frederick Bates?")
+    assert results[0].docs.keys() == {"dab5b86dea3bd4c7ffe05a9f33ae95f7"}
```
jamesbraza, line 76: Can you add a comment where this ID comes from? Why is it not a UUID?

paperqa/agents/docs.py
mskarlin: Went with `compute_total_model_token_cost` and `compute_model_token_cost`

pyproject.toml
```diff
@@ -29,18 +29,63 @@ dependencies = [
...
+    "langchain-community",
+    "langchain-openai",
+    "pymupdf",
+    "requests",
+    "tantivy",
+    "tenacity",
+    "typer",
+    "typing_extensions",
+]
+dev = [
```
jamesbraza, line 57: Should we split out paperqa LLM dependencies (e.g. `anthropic`, `faiss-cpu`) into an `llm` extra?

tests/conftest.py
```diff
@@ -15,3 +24,45 @@ def vcr_config():
...
+@pytest.fixture
+def tmp_path_cleanup(tmp_path: Path) -> Generator[Path, None, None]:
+    yield tmp_path
+    # Cleanup after the test
+    if tmp_path.exists():
+        shutil.rmtree(tmp_path, ignore_errors=True)
+
+
+@pytest.fixture
+def agent_module_dir(tmp_path_cleanup, monkeypatch) -> Path:
```
jamesbraza, line 38: Can you type hint `tmp_path_cleanup`

pyproject.toml
mskarlin: I'm not 100% what the best module split should be for these, we could split those out, but should we include langchain too? I need to test what needs to be in the import path to function.

jamesbraza: Previously we just had them all grouped into one section of `dev-requirements.txt`. We can just have one extra `llm` for now to keep it simple

mskarlin: Yea I am just wondering if it makes sense to extricate langchain into the llm extra too.

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+    from .search import SearchIndex, get_directory_index
+
+except ImportError as e:
+    raise ImportError(
+        '"agents" module is not installed please install it using "pip install paper-qa[agents]"'
+    ) from e
+
+app = typer.Typer()
+
+logging.basicConfig(
```
jamesbraza, line 35: Can we avoid module-level logging configs? Remember we are a package people install/import, so we don't want to clobber their logging set ups on import

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+            show_level=False,
+            console=Console(force_terminal=True),
+        )
+    ],
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_file_times(path: os.PathLike | str) -> dict[str, str]:
```
jamesbraza, line 53:
```suggestion
def get_file_timestamps(path: os.PathLike | str) -> dict[str, str]:
```

Let's be more explicit on what "times" means

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+            except (ValueError, SyntaxError):
+                parsed[key] = value
+        else:
+            parsed = {key: parsed}
+    return parsed
+
+
+def pop_nested_dict_recursive(d: dict[str, Any], path: str) -> tuple[Any, bool]:
+    """
+    Pop a value from a nested dictionary using a period-separated path.
```
jamesbraza, line 88: Can you document this if it in-place modifies the input `dict`, or if it makes a copy?

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+
+    value, should_remove = pop_nested_dict_recursive(d[keys[0]], ".".join(keys[1:]))
+
+    if should_remove:
+        d.pop(keys[0])
+
+    return value, len(d) == 0
+
+
+def get_current_settings(
```
jamesbraza, line 111:
```suggestion
def get_settings(
```

No need for the word "current" imo

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+    settings_path: str | os.PathLike | None = None,
+) -> dict[str, Any]:
+
+    if settings_path is None:
+        settings_path = pqa_directory("settings") / "settings.yaml"
```
jamesbraza, line 116: Can we move this to a module level `pathlib.Path` variable? It seems to be used in several places. Then we can get rid of `None` arg and set the module variable as the default

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+) -> bool:
+    """Set a persistent PaperQA setting."""
+    settings_path = pqa_directory("settings") / "settings.yaml"
+
+    current_settings = get_merged_settings(
+        parse_dot_to_dict(variable, value), settings_path=settings_path
+    )
+
+    try:
+        _ = QueryRequest(**current_settings)
```
jamesbraza, line 186:
```suggestion
        QueryRequest(**current_settings)
```

Can we do this? Or will some linter complain

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,500 @@
...
+        raise ValueError(
+            f"{variable} (with value {value}) is not a valid setting."
+        ) from e
+
+    logger.info(f"{variable} set to {str(value)[:100]}!")
+
+    with open(settings_path, "w") as f:
+        yaml.dump(current_settings, f)
+
+    return True
```
jamesbraza, line 199: Let's not return anything, in validators it's best to:
- Fail and blow up with an `Exception`
- Silently succeed on success

paperqa/agents/__init__.py
mskarlin: I originally designed it this way, but when you move it to be a default it gets baked into the function at import time, and ignores environment changes for things like tests. If you try to use a `reload` on the module, then all the types get doubled and things like type(A) == type(A) fail. 

So this was my compromise, where the env variables are pulled in dynamically with a function.

tests/conftest.py
```diff
@@ -15,3 +24,45 @@ def vcr_config():
     return {
         "filter_headers": [CROSSREF_HEADER_KEY, SEMANTIC_SCHOLAR_HEADER_KEY],
     }
+
+
+@pytest.fixture
+def tmp_path_cleanup(tmp_path: Path) -> Generator[Path, None, None]:
```
jamesbraza, line 30:
```suggestion
def tmp_path_cleanup(tmp_path: str | os.PathLike) -> Iterator[str | os.PathLike]:
```

Let's stay general if we can

paperqa/agents/search.py
```diff
@@ -0,0 +1,486 @@
...
+            return str(obj)
+        if isinstance(obj, set):
+            return list(obj)
+        return json.JSONEncoder.default(self, obj)
+
+
+class SearchDocumentStorage(str, Enum):
+    JSON_MODEL_DUMP = "json_model_dump"
+    PICKLE_COMPRESSED = "pickle_compressed"
+    PICKLE_UNCOMPRESSED = "pickle_uncompressed"
```
jamesbraza, line 56:
```suggestion
    JSON_MODEL_DUMP = auto()
    PICKLE_COMPRESSED = auto()
    PICKLE_UNCOMPRESSED = auto()
```

Let's use `auto` when the value doesn't matter

paperqa/agents/search.py
```diff
@@ -0,0 +1,486 @@
...
+PQA_INDEX_ABSOLUTE_PATHS = (
+    os.environ.get("PQA_INDEX_ABSOLUTE_PATHS", "true").lower() == "true"
+)
+
+
+class AsyncRetryError(Exception):
+    pass
+
+
+class RobustEncoder(json.JSONEncoder):
```
jamesbraza, line 43: Let's add a docstring explaining what "robust" means

paperqa/agents/search.py
```diff
@@ -0,0 +1,486 @@
...
+        if self == SearchDocumentStorage.JSON_MODEL_DUMP:
+            return json.loads(data)
+        if self == SearchDocumentStorage.PICKLE_COMPRESSED:
+            return pickle.loads(zlib.decompress(data))  # type: ignore[arg-type] # noqa: S301
+        return pickle.loads(data)  # type: ignore[arg-type] # noqa: S301
+
+
+class SearchIndex:
+
+    REQUIRED_FIELDS: ClassVar[list[str]] = ["file_location", "body"]
```
jamesbraza, line 84: Can we make this a `set`? I don't think order matters

paperqa/agents/search.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def __init__(
+        self,
+        fields: list[str] | None = None,
+        index_name: str = "pqa_index",
+        index_directory: str | os.PathLike | None = None,
+        storage: SearchDocumentStorage = SearchDocumentStorage.PICKLE_COMPRESSED,
+    ):
+        if fields is None:
+            fields = ["file_location", "body"]
```
jamesbraza, line 94: Can we make this be `self.REQUIRED_FIELDS`?

paperqa/agents/search.py
```diff
@@ -0,0 +1,486 @@
...
+        self._index_directory = index_directory
+        self._schema = None
+        self._index = None
+        self._searcher = None
+        self._index_files: dict[str, str] = {}
+        self.changed = False
+        self.storage = storage
+
+    async def init_directory(self):
+        await anyio.Path(await self.index_directory).mkdir(parents=True, exist_ok=True)
```
jamesbraza, line 112: Why are we using `anyio.Path`? Can we just use `pathlib.Path`?

paperqa/agents/docs.py
```diff
@@ -0,0 +1,250 @@
...
+    names = set()
+    uniq_docs = []
+    for c in answer.contexts:
+        if c.text.doc.docname not in names:
+            uniq_docs.append(c.text.doc)
+            names.add(c.text.doc.docname)
+    return answer
+
+
+async def stream_cost(
```
jamesbraza, line 155: If possible, can we get rid of all of these stream functions entirely? They are just added indirection imo

tests/conftest.py
mskarlin: ok cutting the reload; it's not needed anymore anyway.

-  The name is bad, I'm changing this to `agent_home_dir` -- this is where the indices and project data are generated. It was named that way before since it's the directory to be used by the module to store data.
- not married to using monkeypatch, I can use patch w a yield.

paperqa/agents/docs.py
```diff
@@ -0,0 +1,250 @@
...
+    "mistralai/Mixtral-8x7B-Instruct-v0.1": (1.0 / 10**6, 1.0 / 10**6),
+    "mistralai/Mixtral-8x22B-Instruct-v0.1": (1.0 / 10**6, 1.0 / 10**6),
+}
+
+
+def compute_model_token_cost(model: str, tokens: int, is_completion: bool) -> float:
+    if model in costs:  # Prefer our internal costs model
+        model_costs: tuple[float, float] = costs[model]
+    else:
+        logger.warning(f"Model {model} not found in costs.")
```
jamesbraza, line 73: What makes you not want to fall back on `litellm` costs? Is it because we don't have `litellm` as a dependency? I think we will add it soon to get rid of our LLM special casing

```python
        try:
            cost_map = litellm.get_model_cost_map(
                url="https://raw.githubusercontent.com/BerriAI/litellm/main/litellm/model_prices_and_context_window_backup.json"
            )[model]
            model_costs = (
                cost_map["input_cost_per_token"],
                cost_map["output_cost_per_token"],
            )
        except KeyError:
            logger.warning(f"Model {model} not found in costs.")
            return 0.0
```

paperqa/agents/helpers.py
```diff
@@ -0,0 +1,92 @@
...
+            "{count}" in template and "{question}" in template and "{date}" in template
+        ):
+            logger.warning(
+                "Template does not contain {count}, {question} and {date} variables. Ignoring template."
+            )
+            template = None
+
+        else:
+            # partial formatting
+            search_prompt = template.replace(r"{date}", get_year())
```
jamesbraza, line 36:
```suggestion
            search_prompt = template.replace("{date}", get_year())
```

Do we need the `r` here? I feel like we just don't put `f` if it's a template

paperqa/agents/helpers.py
```diff
@@ -0,0 +1,92 @@
...
+        raise ValueError(
+            f"Invalid llm: {llm}, note a GPT model must be used for the fake agent search."
+        )
+    client = AsyncOpenAI()
+    model = OpenAILLMModel(config={"model": llm, "temperature": temperature})
+    chain = model.make_chain(client, prompt=search_prompt, skip_system=True)
+    result = await chain({"question": question, "count": count})  # type: ignore[call-arg]
+    search_query = result.text
+    queries = [s for s in search_query.split("\n") if len(s) > 3]  # noqa: PLR2004
+    # remove 2., 3. from queries
```
jamesbraza, line 61: What is `2., 3.`? I am not quite following this comment

paperqa/agents/main.py
```diff
@@ -0,0 +1,383 @@
...
+        fields=["file_location", "body", "question"],
+        index_name="answers",
+        index_directory=index_directory,
+        storage=SearchDocumentStorage.JSON_MODEL_DUMP,
+    )
+
+    response = await run_agent(docs, query, agent_type)
+
+    agent_logger.debug(f"agent_response: {response}")
+    truncation_max = 1_000_000 if verbosity > 1 else 1500 * (verbosity + 1)
```
jamesbraza, line 120: What are the units here? Wdyt of naming `truncation_chars`? I think "max" is implied with truncation

tests/conftest.py
mskarlin: I was keeping the details for posterity, but it's not even used anymore, I'm just going to cut it

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+            client = AsyncOpenAI()
+    elif isinstance(doc.llm_model, AnthropicLLMModel):
+        client = AsyncAnthropic()
+    elif isinstance(doc.llm_model, LangchainLLMModel) and USE_VERTEX:
+        # we have to convert system to human because system is unsupported
+        # Also we do get blocked content, so adjust thresholds
+        client = ChatVertexAI(
+            model=request.llm,
+            safety_settings={
+                HarmCategory.HARM_CATEGORY_DANGEROUS_CONTENT: HarmBlockThreshold.BLOCK_ONLY_HIGH,
```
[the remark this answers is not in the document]
mskarlin: No, because then Vertex has to be installed to get that enum. It's a horribly slow package to have if you want to use a CLI.

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+            client = AsyncOpenAI(
+                base_url=os.environ.get("ANYSCALE_BASE_URL"),
+                api_key=os.environ.get("ANYSCALE_API_KEY"),
+            )
+            logger.info(f"Using Anyscale (via OpenAI client) for {request.llm}")
+        else:
+            client = AsyncOpenAI()
+    elif isinstance(doc.llm_model, AnthropicLLMModel):
+        client = AsyncAnthropic()
+    elif isinstance(doc.llm_model, LangchainLLMModel) and USE_VERTEX:
```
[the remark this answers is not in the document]
mskarlin: sure, I didn't realize this would be the only place they are used.

paperqa/agents/models.py
mskarlin: Ahh actually we only require 3.8 -- interesting. I haven't run the test suite with 3.8 but I'm dubious.

.gitignore
mskarlin: hh that's that this file already was, previously it just didn't include the VS code section. I added it back.

paperqa/agents/docs.py
```diff
@@ -0,0 +1,282 @@
...
+
+    if request.llm.startswith("gemini"):
+        doc.llm_model = LangchainLLMModel(name=request.llm)
+        doc.summary_llm_model = LangchainLLMModel(name=request.summary_llm)
+    else:
+        doc.llm_model = llm_model_factory(request.llm)
+        doc.summary_llm_model = llm_model_factory(request.summary_llm)
+
+    # set temperatures
+    doc.llm_model.config["temperature"] = request.temperature  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
mskarlin: Yea I think that's a good idea, added!

tests/conftest.py
mskarlin: the `.exists` method only exists on a Path object.

paperqa/agents/search.py
mskarlin: Talked about this live, but anyio's Path is async so reads/writes can occur in worker threads.

paperqa/agents/main.py
mskarlin: characters

paperqa/agents/helpers.py
mskarlin: it's literal removing "2., 3." for example from a query i added a quote

mskarlin: https://regex101.com/r/W2f7F1/1

paperqa/agents/docs.py
mskarlin: Absolutely, I want to add litellm, but I want to integrate litellm in a second PR, it'll change too much in this one.

.gitignore
```diff
@@ -1,3 +1,138 @@
...
+*.vsix
+# General
+.DS_Store
+.AppleDouble
+.LSOverride
+
+# Icon must end with two \r
+Icon
+
+
```
jamesbraza, line 117:
```suggestion
# Icon must end with two \r
Icon[\r]

```

The spacing may not play out right here, but you get the idea

pyproject.toml
```diff
@@ -29,18 +29,65 @@ dependencies = [
...
+    "mypy",
+    "pre-commit",
+    "pytest",
+    "pytest-asyncio",
+    "pytest-subtests",
+    "pytest-sugar",
+    "pytest-timer",
+    "pytest-vcr",
+    "python-dotenv",
+    "pyzotero",
```
jamesbraza, line 68: Should `pyzotero ` live in `dev`? Maybe it should live in `llms`

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,550 @@
...
+        rich_tracebacks=True,
+        markup=True,
+        show_path=False,
+        show_level=False,
+        console=Console(force_terminal=True),
+    )
+
+    rich_handler.setFormatter(logging.Formatter("%(message)s", datefmt="[%X]"))
+
+    root_logger = logging.getLogger()
```
jamesbraza, line 74: Why are we manipulating the root logger? Can we just use the `paperqa` loggers (and its descendents)? So that would be `logging.getLogger("paperqa")`

paperqa/agents/__init__.py
```diff
@@ -0,0 +1,550 @@
...
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(
+            verbosity_map.get(min(verbosity, 2), {}).get(logger_name, default_level)
+        )
+        # fallback to the rich hangler
+        logger.handlers.clear()
+        logger.propagate = True
+
+
+logger = logging.getLogger(__name__)
```
jamesbraza, line 89: Let's move this to the top of the module, before all functions. That's usually the convention with `logging`

paperqa/agents/__init__.py
jamesbraza: Now that we no longer have `reload`, is there any chance we can make a module-level variable for this?

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+
+class SupportsPickle(Protocol):
+    """Type protocol for typing any object that supports pickling."""
+
+    def __reduce__(self) -> str | tuple[Any, ...]: ...
+    def __getstate__(self) -> object: ...
+    def __setstate__(self, state: object) -> None: ...
+
+
+def strip_answer(answer: Answer):
```
jamesbraza, line 55: Can we move this to a method of the `Answer` object?

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+            " names (non-default route), at least the gen answer tool must be supplied."
+        ),
+    )
+
+    @field_validator("tool_names")
+    @classmethod
+    def validate_tool_names(cls, v: set[str] | None) -> set[str] | None:
+        if v is None:
+            return None
+        from .main import GenerateAnswerTool
```
jamesbraza, line 162: Can you add a comment about avoiding circular imports

paperqa/agents/docs.py
mskarlin: ok -- the whole file could actually be refactored out, much better here now.

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+    def get_parse_type(self, config: ParsingConfiguration) -> str:
+        if self == ParsingOptions.PAPERQA_DEFAULT:
+            return config.parser_version_string
+        assert_never()
+
+
+class ChunkingOptions(str, Enum):
+    SIMPLE_OVERLAP = "simple_overlap"
+
+    # Note that SIMPLE_OVERLAP must be valid for all by default
```
jamesbraza, line 187: This comment seems misplaced, can you move it next to `SIMPLE_OVERLAP`?

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+        return [self.PAPERQA_DEFAULT]  # type: ignore[list-item]
+
+    def get_parse_type(self, config: ParsingConfiguration) -> str:
+        if self == ParsingOptions.PAPERQA_DEFAULT:
+            return config.parser_version_string
+        assert_never()
+
+
+class ChunkingOptions(str, Enum):
+    SIMPLE_OVERLAP = "simple_overlap"
```
jamesbraza, line 185: Can we use `auto` for the value? I don't know if the value matters

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+
+    # Note that SIMPLE_OVERLAP must be valid for all by default
+    @property
+    def valid_parsings(self) -> list[ParsingOptions]:
+        # TODO: implement for future parsing options
+        valid_parsing_dict: dict[str, list[ParsingOptions]] = {}
+        return valid_parsing_dict.get(self.value, [])
```
jamesbraza, line 192: Won't this always return an empty list?

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+class MismatchedModelsError(Exception):
+    """Error to throw when model clients clash ."""
+
+    LOG_METHOD_NAME: ClassVar[str] = "warning"
+
+
+class QueryRequest(BaseModel):
+    query: str = ""
+    # will be propagated to Answer
+    id: UUID = Field(default_factory=uuid4)
```
jamesbraza, line 246: Can you move this comment to a `Field.description`

paperqa/agents/models.py
```diff
@@ -0,0 +1,535 @@
...
+                str(paper_directory),
+                embedding,
+                str(parsing_configuration.chunksize),
+                str(parsing_configuration.overlap),
+                parsing_configuration.chunking_algorithm,
+            ]
+        )
+
+        return (
+            f"pqa_index_{hashlib.md5(index_fields.encode()).hexdigest()}"  # noqa: S324
```
jamesbraza, line 343: Can you move this encode functionality to a free function in `utils` or the module level here?

paperqa/agents/models.py
mskarlin: yea this is a stub to keep in place for NEW parsers which will be added in a f/u PR.

pyproject.toml
mskarlin: It's not an llm thing -- it's really an extension of its own. I'm leaving in dev for backwards compatibility.

paperqa/agents/models.py
mskarlin: should go in the func below this actually.

paperqa/agents/models.py
mskarlin: no; we'll need the values in downstream usage, grobid will be implemented here.

paperqa/agents/models.py
mskarlin: Gonna come back to this in a later PR — we might need to bump up our minimum python version when we do the merge to main

paperqa/agents/helpers.py
```diff
@@ -0,0 +1,238 @@
...
+        search_prompt = (
+            "We want to answer the following question: {question} \n"
+            "Provide {count} unique keyword searches (one search per line) and year ranges "
+            "that will find papers to help answer the question. "
+            "Do not use boolean operators. "
+            "Make sure not to repeat searches without changing the keywords or year ranges. "
+            "Make some searches broad and some narrow. "
+            "Use this format: [keyword search], [start year]-[end year]. "
+            "where end year is optional. "
+            f"The current year is {get_year()}\n\n"
```
jamesbraza, line 63: Should we have trailing `\n\n` here? Usually callers join together prompts as they see fit

Disclaimer: I know this is how it's done internally, but maybe it's not a good pattern

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+        "\n- Gather more evidence using a different phrase"
+        "\nIf you search for more papers or collect new papers cited by previous evidence,"
+        " remember to gather evidence again."
+        " Once you have five or more pieces of evidence from multiple sources, or you have tried a few times, "
+        "call {gen_answer_tool_name} tool. The {gen_answer_tool_name} tool output is visible to the user, "
+        "so you do not need to restate the answer and can simply terminate if the answer looks sufficient. "
+        "The current status of evidence/papers/cost is {status}"
+    )
+    paper_directory: str | os.PathLike = Field(
+        default=Path("."),
```
jamesbraza, line 91: What does `"."` mean? Does it mean `paperqa/agents`, or the caller's directory?

Do you mind making the default be not a relative path?

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+    LOG_METHOD_NAME: ClassVar[str] = "warning"
+
+
+class QueryRequest(BaseModel):
+    query: str = ""
+    id: UUID = Field(
+        default_factory=uuid4,
+        description="Identifier which will be propagated to the Answer object.",
+    )
+    llm: str = "gpt-4o-2024-08-06"
```
jamesbraza, line 229: Can you add a description what this LLM is used for? There's `agent_llm`, `summary_llm`, what do you think of giving this one a more verbose prefix?

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+    named_prompt: str | None = None
+    # if you change this to something other than default
+    # modify code below in update_prompts
+    prompts: PromptCollection = Field(
+        default=STATIC_PROMPTS["default"], validate_default=True
+    )
+    agent_tools: AgentPromptCollection = Field(default_factory=AgentPromptCollection)
+    texts_index_mmr_lambda: float = 1.0
+    texts_index_embedding_config: dict[str, Any] | None = None
+    docs_index_mmr_lambda: float = 0.5
```
jamesbraza, line 248: Should we default this to 1?

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+        )
+        result = await chain({"question": self.answer.question, "answer": self.answer.answer})  # type: ignore[call-arg]
+        return result.text.strip()
+
+
+class TimerData(BaseModel):
+    # NOTE: use lambda because asyncio.get_running_loop can throw: RuntimeError: no running event loop
+    start_time: float = Field(
+        default_factory=lambda: asyncio.get_running_loop().time()  # noqa: FURB111
+    )
```
jamesbraza, line 371: What do you think of having this `start_time` not come from `asyncio`, but normal `time.time`?

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+        for name, durations in self.timers.items():
+            low = min(durations)
+            mean = sum(durations) / len(durations)
+            max_ = max(durations)
+            result[name] = {
+                "low": low,
+                "mean": mean,
+                "max": max_,
+                "total": sum(durations),
+            }
```
jamesbraza, line 432:
```suggestion
            mean = sum(durations) / len(durations)
            result[name] = {
                "low": min(durations),
                "mean": mean,
                "max": max(durations),
                "total": sum(durations),
            }
```

Imo sometimes less locals is more readable

paperqa/agents/models.py
```diff
@@ -0,0 +1,520 @@
...
+                "messages": messages_to_dict(messages[0]),
+                # SEE: https://platform.openai.com/docs/api-reference/chat/create#chat-create-functions
+                "functions": kwargs["invocation_params"]["functions"],
+                "tool_history": self.tool_invocations,
+            },
+            "model": kwargs["invocation_params"]["model"],
+            "date": datetime.now().isoformat(),
+        }
+
+    async def on_llm_end(self, response: LLMResult, **kwargs) -> None:
```
jamesbraza, line 503: Does this do anything? It seems like we have some checks but don't actually fill out the result

Maybe add a comment why we have the checks, but leave it up to subclassers to extend from there

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+
+        try:
+            await _add_document_with_retry()
+        except RetryError:
+            logger.exception(
+                f"Failed to add document after {max_retries} attempts: {index_doc['file_location']}"
+            )
+            raise
+
+        # Success
```
jamesbraza, line 234: What is this comment for? Can we either expand it, or remove it?

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+        filehash = None
+        if body:
+            filehash = self.filehash(body)
+        index_files = await self.index_files
+        return bool(
+            index_files.get(filename)
+            and (filehash is None or index_files[filename] == filehash)
+        )
+
+    async def add_document(
```
jamesbraza, line 192: Can you type hint the return type

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+        # Success
+
+    @staticmethod
+    @retry(
+        stop=stop_after_attempt(1000),
+        wait=wait_random_exponential(multiplier=0.25, max=60),
+        retry=retry_if_exception_type(AsyncRetryError),
+        reraise=True,
+    )
+    def delete_document(index: Index, file_location: str):
```
jamesbraza, line 243:
```suggestion
    def delete_document(index: Index, file_location: str) -> None:
```

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+        try:
+            writer = index.writer()
+            writer.delete_documents("file_location", file_location)
+            writer.commit()
+        except ValueError as e:
+            if "Failed to acquire Lockfile: LockBusy." in str(e):
+                raise AsyncRetryError("Failed to acquire lock") from e
+            raise
+
+    async def remove_from_index(self, file_location: str):
```
jamesbraza, line 253:
```suggestion
    async def remove_from_index(self, file_location: str) -> None:
```

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+                docs_index_dir / f"{filehash}.{self.storage.extension()}", "rb"
+            ) as f:
+                content = await f.read()
+                if keep_filenames:
+                    return self.storage.read_from_string(content), file_location
+                return self.storage.read_from_string(content)
+        return None
+
+    def clean_query(self, query: str):
+        for replace in ["*", "[", "]"]:
```
jamesbraza, line 290:
```suggestion
    def clean_query(self, query: str) -> str:
        for replace in {"*", "[", "]"}:
```

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+            docs_index_dir = await self.docs_index_directory
+            # TODO: since the directory is part of the filehash these
+            # are always missing. Unsure of how to get around this.
+            await (docs_index_dir / f"{filehash}.{self.storage.extension()}").unlink(
+                missing_ok=True
+            )
+
+            self.changed = True
+
+    async def save_index(self):
```
jamesbraza, line 268:
```suggestion
    async def save_index(self) -> None:
```

paperqa/agents/search.py
```diff
@@ -0,0 +1,487 @@
...
+        docs_kwargs: Keyword arguments for the Docs object.
+        max_concurrency: maximum number of files to be simultaneously indexed.
+    """
+    semaphore = anyio.Semaphore(max_concurrency)
+
+    if use_absolute_directory_path:
+        directory = await directory.absolute()
+
+    search_index = SearchIndex(
+        fields=["title", "file_location", "body", "year"],
```
jamesbraza, line 440: Can we somehow use `SearchIndex.REQUIRED_FIELDS` here?

jamesbraza (approved):
Huge battle @mskarlin here wow. Amazing work

paperqa/agents/helpers.py
mskarlin: Sure we can cut this

paperqa/agents/models.py
mskarlin: Callers current directory -- it gets turned into an absolute path for you in the function. This is a little complicated because as a CLI, it needs to be whatever directory the user is in OR an absolute path if specified. This way it supports both, and it reflects the default behavior, i.e. you move to a directory and then it runs there.

jamesbraza: Can you add a comment or description just documenting this then?

paperqa/agents/models.py
mskarlin: good call, i'm not sure why it defaults to this anyway.

paperqa/agents/models.py
mskarlin: actually can just use `Path.cwd()`

paperqa/agents/models.py
mskarlin: I agree with this, but since this is used in a lot of places (even outside of paper-qa), I want to make that change in another PR.

PR #310, opened 2024-08-30
Add new unpaywall provider

mskarlin:
Adds a new metadata provider, unpaywall. (https://unpaywall.org/) Also associated tests to get open access status of papers. 

Additional minor bookeeping updates: 
- Refreshes all the cassettes for having extra data from unpaywall, also fixes a bug caused by the prior PR which removed facet filtering from crossref on doi queries (it's not supported anyway) 
- Renamed ALL_CLIENTS -> DEFAULT_CLIENTS since that's really what it is
- Adds `anthropic` into the agents module since it was needed in the import path

### DIFF
--- a/paperqa/clients/unpaywall.py
+++ b/paperqa/clients/unpaywall.py
@@ -107,11 +107,6 @@ class UnpaywallProvider(DOIOrTitleBasedProvider):
 
         return self._create_doc_details(results)
 
-    @staticmethod
-    def clean_query(query: str) -> str:
-        """Toss out any word in the query that has a forbidden character."""
-        return query
-
     async def search_by_title(
         self,
         query: str,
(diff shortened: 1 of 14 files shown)


paperqa/clients/unpaywall.py
```diff
@@ -0,0 +1,206 @@
...
+            raise DOINotFoundError(
+                f"Unpaywall results returned with a bad schema for DOI {doi!r}."
+            ) from e
+
+        return self._create_doc_details(results)
+
+    @staticmethod
+    def clean_query(query: str) -> str:
+        """Toss out any word in the query that has a forbidden character."""
+        return query
```
whitead, line 113: ?

whitead (approved):

paperqa/clients/unpaywall.py
mskarlin: Whoops

PR #312, opened 2024-08-30
Recursive directory indexing; catch indexing errors

sidnarayanan:
Opening this PR into `test-md5` to make the diff easier to read. Will rebase + point to main once #311 is merged.

- Added `get_directory_index(recursive: bool=True)` to recursively index a directory
- If S2 doesn't have journal data, it can either skip the field or return None. Extended to handling both cases
- Raise an `ImpossibleParsingError` if no text is parsed from PDF

Also lots of whitespace changes from ruff and black - they're conflicting with each other in some places.


paperqa/agents/models.py
```diff
@@ -354,6 +346,7 @@ def strip_answer(
     async def get_summary(self, llm_model="gpt-4-turbo") -> str:
         sys_prompt = (
             "Revise the answer to a question to be a concise SMS message. "
+            # Force two lines to keep ruff and black happy
```
jamesbraza, line 349: Why do you have this?

paperqa/agents/search.py
```diff
@@ -443,9 +445,13 @@ async def get_directory_index(
 
     valid_files = [
         file
-        async for file in directory.iterdir()
+        async for file in (directory.rglob("*") if recursive else directory.iterdir())
         if file.suffix in {".txt", ".pdf", ".html"}
     ]
+    if len(valid_files) > 999:  # noqa: PLR2004
```
jamesbraza, line 451: Why `999` and can you move it to a module level or class variable?

```python
VALID_FILE_LENGTH = 999
```

paperqa/readers.py
```diff
@@ -65,6 +66,10 @@ def chunk_pdf(
             f"ParsedText.content must be a `dict`, not {type(parsed_text.content)}."
         )
 
+    if not parsed_text.content:
+        # did not parse any text - an empty or (more likely) corrupted document
+        raise ImpossibleParsingError("No text was parsed from the document.")
```
jamesbraza, line 71: Can you combine this comment into the `ImpossibleParsingError`'s message?

paperqa/agents/models.py
sidnarayanan: Black thinks the line length is too long, so breaks it up into two strings on separate lines. 

Ruff thinks the line lengths are too short, so puts the two strings on one line. Then Ruff complains that we're doing string concatenation in one line. 

I think there's incompatible configurations.

jamesbraza: We don't have `ruff-format` here, so I am not sure why Ruff would be reshaping strings. Lemme check this out and play with it

jamesbraza: Deleting this line everything seems fine (?)

I think you can just delete this comment and everything passes

paperqa/agents/search.py
sidnarayanan: Made a module-level variable. And it's an arbitrary cutoff

paperqa/agents/models.py
sidnarayanan: Seems the ruff VSCode extension was interfering. Resolved!

jamesbraza (approved):

whitead:
Is this relevant @sidnarayanan ? The base PR is dead

sidnarayanan:
Oh if #311 is merged, this should be good to go. I can rebase off main. @mskarlin what do you think?

sidnarayanan:
Oh boy, this has really drifted from main. I'm going to close this PR and open a new one

PR #315, opened 2024-09-06
LiteLLM integration

mskarlin:
First pass at integrating LiteLLM into PaperQA to replace our code which tries to set up clients for any user's LLM request. This is a draft, as I still have some failing tests to fix, but I think that's in the test code rather than our module. I'm opening this PR into the open refactor branch so that we don't drift again. If that gets merged first then I can just rebase. 

One minor complication is that instead of client objects being unpickleable, now the entire LLMModel object is unpickleable without removing the LiteLLM "Router" object as an attribute. It's not a big deal, but could cause unexpected behavior if someone set up some complex Router configuration and passed it into our LiteLLMModel, expecting it to still be there after a serialization cycle. 

After removing the other LLM clients, I based all of the LLM/embedding model configuration on the new `Settings` object rather than the prior `Docs` object attributes. The `Docs` object is just a holder with some convenience methods.


paperqa/docs.py
```diff
@@ -209,7 +209,15 @@ def set_client(
...
-        if client is None and isinstance(self.llm_model, OpenAILLMModel):
+        """Sets an API client for non LiteLLM compatibility.
+
+        If using the LiteLLMModel or LiteLLMEmbeddingModel, no client is needed.
+
+        """
+        if isinstance(self.llm_model, LiteLLMModel):
+            client = None
+        # below logic only for backwards compatibility
+        elif client is None and isinstance(self.llm_model, OpenAILLMModel):
```
jamesbraza, line 220: Let's break backwards compatibility and remove all this extra stuff

paperqa/llms.py
```diff
@@ -34,36 +36,24 @@
...
-#     model_arr = list(model_union)
-
-#     complete_model_types = get_type_hints(
-#         openai.types.completion_create_params.CompletionCreateParamsBase
-#     )["model"]
-#     complete_model_union = get_args(get_args(complete_model_types)[1])
-#     complete_model_arr = list(complete_model_union)
-
-#     return model_name in model_arr or model_name in complete_model_arr
+litellm.vertex_ai_safety_settings = [
```
jamesbraza, line 39: We should not be specify this at the package level, given others install us. We need to let callers specify global behaviors like this.

What if someone installs `paper-qa`, but doesn't realize we removed safeties?

Let's move this to either a helper function or into our tests?

paperqa/docs.py
mskarlin: @whitead what do you think?

paperqa/llms.py
mskarlin: Ofc for sure

paperqa/docs.py
whitead: Yes, break it all

paperqa/llms.py
whitead: TBH - I think this is required for basic functionality. Their safety stuff trips up on all kinds of benign scientific words. Like a paper about breast cancer will give extremely opaque errors in paper-qa. I can't imagine a use-case where you would want to filter this strongly. Note that they apply other safety filtering (e.g., just turning this setting down won't let you make meth or bombs).

mskarlin: makes sense @whitead  -- I compromised by making this a default setting that the user can override, but moved it out of module scope.

paperqa/docs.py
mskarlin: it is broken

paperqa/llms.py
jamesbraza: I agree with both of you. People won't always use VertexAI, but when they are using it, it should be specified. I like Mike's idea of placing it to settings where it's obviously configured, instead of unexpectedly/"silently" at `import` time

pyproject.toml
```diff
@@ -19,6 +19,7 @@ classifiers = [
 dependencies = [
     "PyCryptodome",
     "html2text",
+    "litellm",
     "numpy",
     "openai>=1",
```
mskarlin, line 24: I think we want to drop this as a primary dependency now too

jamesbraza: Yeah I like the idea. Note that LiteLLM depends on `openai`, so it's not gonna lighten the load overall, but I agree it postures _our_ code to use LiteLLM for everything

pyproject.toml
```diff
@@ -19,6 +19,7 @@ classifiers = [
 dependencies = [
     "PyCryptodome",
     "html2text",
+    "litellm",
```
jamesbraza, line 22: Can you add:

```toml
    "litellm",  # SEE: https://github.com/BerriAI/litellm/issues/825
```

To the `mypy` overrides in the config?

whitead (commented):
Awesome! Couple quick q's:


1. Can we drop tiktoken now - couldn't tell from pr easily if it's possible
2. Can we put cost into the `LLMResult` now, since litellm has that in responses?

Looks so much better! We can add some sweet tenacity retries too now that all exceptions map correctly

paperqa/agents/__init__.py
```diff
@@ -67,13 +67,16 @@ def configure_cli_logging(verbosity: int = 0) -> None:
 
     rich_handler.setFormatter(logging.Formatter("%(message)s", datefmt="[%X]"))
 
+    module_logger = logging.getLogger("paperqa")
```
jamesbraza, line 70: Can we import this somehow?

paperqa/agents/tools.py
```diff
@@ -46,6 +47,9 @@ class SharedToolState(BaseModel):
     docs: Docs
     settings: Settings
     profiler: SimpleProfiler = Field(default_factory=SimpleProfiler)
+    _llm_model: LiteLLMModel | None = None
```
jamesbraza, line 50: Can we make a sentinel `POPULATE_FROM_SETTINGS: ClassVar[None] = None` and use that as the default?

It would be more readable

paperqa/config.py
```diff
@@ -432,6 +449,43 @@ def from_name(
 
         raise FileNotFoundError(f"No configuration file found for {config_name}")
 
+    def get_llm(self) -> LiteLLMModel:
+        return LiteLLMModel(
+            name=self.llm,
+            config=self.llm_config
+            or {
+                "model_list": [
```
jamesbraza, line 457: Can you add a brief inline comment what this structure is supposed to match?

I also see it's repeated, should we make a private helper function for this?

jamesbraza (approved):
Really nice work! 🥳 So much cleaner

paperqa/agents/tools.py
mskarlin: great idea

paperqa/config.py
mskarlin: yea good call -- added

paperqa/agents/__init__.py
mskarlin: yea I think we can just split __name__ actually

mskarlin:
> Awesome! Couple quick q's:
> 
> 1. Can we drop tiktoken now - couldn't tell from pr easily if it's possible
> 2. Can we put cost into the `LLMResult` now, since litellm has that in responses?
> 
> Looks so much better! We can add some sweet tenacity retries too now that all exceptions map correctly

We use titkoken in few places for parsing and our sparse encoding, since litellm just wraps it I think maybe we leave that for a later PR. 

looking into adding costs to the response

mskarlin:
@whitead the cost change is definitely possible, but we'll need to change up our `LLMModel` contracts because we expect strings returned from their completions methods. We're actually just tossing the rest of the response away at the moment so this would be a good change. I don't want to do bunch it all into this PR right now since it's already such a large change.

PR #320, opened 2024-09-07
Check if a journal name starts with  `the` 

### LINKED ISSUE #317: Check if journal name starts with `the`

geemi725:
Found exceptions where a journal name can start with `the` in the JournalQualityDB

Now checks the journal name with and without `the`

Closes #317


paperqa/clients/journal_quality.py
```diff
@@ -44,7 +44,12 @@ async def _process(
         # remember, if both have docnames (i.e. key) they are
         # wiped and re-generated with resultant data
         return doc_details + DocDetails(  # type: ignore[call-arg]
-            source_quality=self.data.get(query.journal.casefold(), -1)  # type: ignore[union-attr]
+            source_quality=max(
+                [
+                    self.data.get(query.journal.casefold(), -1),  # type: ignore[union-attr]
```
jamesbraza, line 49:
```suggestion
                    self.data.get(query.journal.casefold(), DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
```

paperqa/clients/journal_quality.py
```diff
@@ -44,7 +44,12 @@ async def _process(
         # remember, if both have docnames (i.e. key) they are
         # wiped and re-generated with resultant data
         return doc_details + DocDetails(  # type: ignore[call-arg]
-            source_quality=self.data.get(query.journal.casefold(), -1)  # type: ignore[union-attr]
```
jamesbraza, line 47: Should we be using `casefold`?

geemi725: I think so? just to be more robust with journal names. @mskarlin  thoughts?

whitead: yup that's what we've been using on title comparisons!

whitead:
Thanks @geemi725! Can you port over the unit test too?

tests/test_clients.py
```diff
@@ -299,7 +299,7 @@ async def test_s2_only_fields_filtering():
         assert not s2_details.source_quality, "No source quality data should exist"  # type: ignore[union-attr]
 
 
-@pytest.mark.vcr
+@pytest.mark.vcr(record_mode="new_episodes")
```
whitead, line 302: ~Don't think you need this~ - I was wrong

```suggestion
@pytest.mark.vcr
```

whitead (approved):
LGTM

tests/test_clients.py
jamesbraza: I think we should let the VCR config in `conftest.py` be the one source of truth. So I vote we do actually commit this change

PR #321, opened 2024-09-08
Changed to rely on litellm for computing cost

whitead:
See title


paperqa/llms.py
```diff
@@ -129,9 +129,9 @@ class LLMModel(ABC, BaseModel):
         " LLMResult (different than callbacks that execute on each chunk)",
         exclude=True,
     )
-    config: dict = Field(default={})
+    config: dict = Field(default_factory=dict)
 
-    async def acomplete(self, prompt: str) -> str:
+    async def acomplete(self, prompt: str) -> tuple[str, tuple[int, int]]:
```
jamesbraza, line 134: Since this is sort of an `abstractmethod`, can you somehow document what the return type is?

Can make a `NamedTuple` that is sort of self-documenting, or a docstring

paperqa/llms.py
```diff
@@ -205,16 +207,17 @@ async def execute(
...
+                    output, usage = await self.achat(messages)
                 else:
                     sync_callbacks = [
                         f for f in callbacks if not is_coroutine_callable(f)
                     ]
                     async_callbacks = [f for f in callbacks if is_coroutine_callable(f)]
                     completion = self.achat_iter(messages)
                     text_result = []
-                    async for chunk in completion:  # type: ignore[attr-defined]
+                    async for chunk, _usage in completion:  # type: ignore[attr-defined]
```
jamesbraza, line 220: Why is this `_usage` if it's used below? Also, imo `usage` is a bit vague out of context.

Do you mind naming it something like `chunk_usage`

jamesbraza (approved):

mskarlin (approved):

paperqa/llms.py
whitead: OK added a new class and revised code to use it

jamesbraza: Nice, thanks

PR #325, opened 2024-09-09
Removing `make_chain` in favor of `run_prompt`

jamesbraza:
We seemed to always `make_chain` and then immediately call it. This PR just removes the `make_chain` closure in favor of a normal stateless free function `run_prompt`

This PR also cleans up `make_chain` by:
- Decomposing it into two private methods
- Simplifies its implementation where possible


whitead:
What's the difference between `PromptRunner` and `make_chain`?

whitead (approved):
We're free from the chains that bind us

jamesbraza:
> What's the difference between `PromptRunner` and `make_chain`?

So in general:
- `make_chain` was a method creating a closure to bind arguments, `run_prompt` is a stateless (no pre-binding) method
- When talking about `map_fxn_summary`, it was designed around a partially bound function. Its `Chain` was a type alias for the output closure from `make_chain`

This PR could take two stances:
1. Invent a `functools.partial`'d version of `run_prompt` that's compatible with the old `make_chain`. Correspondingly, `PromptRunner` is a type alias equivalent to `Chain` (it basically renames `Chain`)
2. Pass an additional argument to `map_fxn_summary` which handles the additional arguments for a non-closure

Anyways, we discussed this in person, going to merge this PR

PR #336, opened 2024-09-09
Add configs for contracrow + wikicrow

mskarlin:
Two jsons for modes of interest for the paper.


jamesbraza (approved):

whitead (commented):
We might want to have a complete specification, in case the defaults later change. Like system prompt is missing on contracrow

PR #345, opened 2024-09-09
Test speed ups

mskarlin:
First I'm tying DNS caching to see if it's a DNS issue.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,6 +40,12 @@ jobs:
         python-version: ["3.12"] # Our min and max supported Python versions
     steps:
       - uses: actions/checkout@v4
+      - name: Setup DNS caching
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y dnsmasq
+          echo "nameserver 127.0.0.1" | sudo tee /etc/resolv.conf
+          sudo systemctl start dnsmasq
       - name: Set up uv
         run: |-
           curl -LsSf https://astral.sh/uv/install.sh | sh


whitead:
An improvement - but it's not using the cassette (?)

whitead:
Hey @mskarlin - tried a few variants. 21d703fce6acbd4113e7959535e090d6172b3c58 uses python from a uv venv, same poor performance (canceled after 10 minutes) and 0c7c64fca4c2a6f56828c856caad17e7f320b8a4 uses actions/setup-python and gets same poor performance. Only other python to try I guess would be system python, but it's 3.10 on our runners. Not sure what else to try?

mskarlin:
Yea we can just build the requirements.txt via `uv export` then use the system python & upgrade it like: ```      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v5
        with:
          python-version: ${{ matrix.python-version }}
          cache: pip
      - run: pip install .``` 
      
Then the system python should be what we use. We can return to using uv for tests after we figure this out.

whitead:
> Yea we can just build the requirements.txt via `uv export` then use the system python & upgrade it like:
> 
> ```
>       - name: Set up Python ${{ matrix.python-version }}
>         uses: actions/setup-python@v5
>         with:
>           python-version: ${{ matrix.python-version }}
>           cache: pip
>       - run: pip install .
> ```
> 
> Then the system python would be what we use. We can return to using uv for tests after we figure this out.

That's pretty much https://github.com/Future-House/paper-qa/commit/0c7c64fca4c2a6f56828c856caad17e7f320b8a4 - except uv did the pip install. 

I tried to not use `uv` for doing the pip install, but that's not compatible with our `pyproject.toml`. I think it's pretty unlikely that installing system packages with `uv` would impact performance - but we can try locally committing a requirements I guess

whitead:
Ok @mskarlin - I completely purged uv so it's not used for parsing toml file, for pip install, or for python management. Completely clean in b2c1e33ab727bd1ac17dd60ad516c8cc511cbd4b. We'll see how long the tests take

PR #349, opened 2024-09-10
Adding `pylint` 

jamesbraza:
Introduces `pylint` into CI, suppresses a few more errors, and manually fixes all errors that came up

### DIFF
--- a/paperqa/agents/__init__.py
+++ b/paperqa/agents/__init__.py
@@ -71,16 +71,16 @@ def configure_cli_logging(verbosity: int = 0) -> None:
 
     rich_handler.setFormatter(logging.Formatter("%(message)s", datefmt="[%X]"))
 
-    module_logger = logging.getLogger(__name__.split(".")[0])
+    module_logger = logging.getLogger(__name__.split(".", maxsplit=1)[0])
 
     if not any(isinstance(h, RichHandler) for h in module_logger.handlers):
         module_logger.addHandler(rich_handler)
 
-    for logger_name, logger in logging.Logger.manager.loggerDict.items():
-        if isinstance(logger, logging.Logger) and (
+    for logger_name, logger_ in logging.Logger.manager.loggerDict.items():
+        if isinstance(logger_, logging.Logger) and (
             log_level := verbosity_map.get(min(verbosity, 2), {}).get(logger_name)
         ):
-            logger.setLevel(log_level)
+            logger_.setLevel(log_level)
 
     if verbosity > 0:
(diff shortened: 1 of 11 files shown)


paperqa/settings.py
```diff
@@ -66,7 +66,7 @@ def available_for_inference(self) -> list["ParsingOptions"]:
 def _get_parse_type(opt: ParsingOptions, config: "ParsingSettings") -> str:
     if opt == ParsingOptions.PAPERQA_DEFAULT:
         return config.parser_version_string
-    assert_never()
+    assert_never(opt)
```
jamesbraza, line 69: @mskarlin fyi it looks like you didn't properly roll out `assert_never`

paperqa/agents/__init__.py
```diff
@@ -79,16 +79,16 @@ def configure_cli_logging(verbosity: int = 0) -> None:
 
     rich_handler.setFormatter(logging.Formatter("%(message)s", datefmt="[%X]"))
 
-    module_logger = logging.getLogger(__name__.split(".")[0])
+    module_logger = logging.getLogger(__name__.split(".", maxsplit=1)[0])
```
mskarlin, line 82: I don't think we want maxsplit here -- like `paperqa.agents.search` -- we want two splits there.

paperqa/clients/__init__.py
```diff
@@ -19,19 +19,14 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_CLIENTS: (
```
mskarlin, line 22: We're losing some semantics here it takes either a collection OR a sequence of collections -- this is removing the second possibility.

jamesbraza: So the place this gets used still has both. Imo it's fine for the default to only have one of the logic branches

paperqa/agents/__init__.py
jamesbraza: We throw away everything beyond `[0]` so I think `maxsplit` would be a slight improvement

paperqa/clients/__init__.py
mskarlin: Ahh i see what you're saying I was thinking this was in a doc signature -- makes sense.

jamesbraza:
This is sort of a draft. Some of this is going to go away with `aviary` refactor removing some of the code

jamesbraza:
Okay, it's ready for review

whitead (approved):
Nice work!

mskarlin (approved):

PR #350, opened 2024-09-10
Renaming fitz to pymupdf

mskarlin:
See note here: https://pymupdf.readthedocs.io/en/latest/tutorial.html#note-on-the-name-fitz


jamesbraza (approved):
LGTM

PR #360, opened 2024-09-11
Added account for cost info

whitead:
Waiting on https://github.com/Future-House/aviary/pull/28

Checks messages from agent for usage info and if present, will attach to current answer objects.

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -152,7 +152,7 @@ class Environment(_Environment[EnvironmentState]):
         self, action: ToolRequestMessage
     ) -> tuple[list[Message], float, bool, bool]:
 
-        # add usage for actions that have usage
+        # add usage for action if it has usage
         info = action.info
         if info and "usage" in info and "model" in info:
             r = LLMResult(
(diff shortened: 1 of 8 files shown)


mskarlin (approved):
Super clean!

paperqa/agents/env.py
```diff
@@ -151,21 +151,22 @@ def export_frame(self) -> Frame:
     async def step(
         self, action: ToolRequestMessage
     ) -> tuple[list[Message], float, bool, bool]:
+
+        # add usage for actions that have usage
```
jamesbraza, line 155:
```suggestion
        # add usage if the action has usage
```

There's just one action

paperqa/settings.py
```diff
@@ -239,6 +239,12 @@ class AgentSettings(BaseModel):
         default="gpt-4o-2024-08-06",
         description="Model to use for agent",
     )
+
+    agent_llm_config: dict | None = Field(
```
jamesbraza, line 243: A few lines below this there is an `agent_config` that is basically kwargs for the `Agent`/`ToolSelector`

Do you mind adjusting their names and/or descriptions so it's clear what is the difference between `agent_llm_config` and `agent_config`? They're similar enough in naming right now I think we should fix

whitead: 👍 Tried to make the descriptions more clear in distinction

jamesbraza: Thanks for your efforts! Being honest, it's still not quite intuitive enough for my tastes, but for now it's good.

I think we can make this clearer by moving `to_aviary_tool_selector` and `to_ldp_agent` to be methods of `AgentSettings`, since _all_ their information is derived from there.

jamesbraza (approved):
Nice work

PR #363, opened 2024-09-11
Fixed extra noise in logs from litellm routers

whitead:
Finally figured it out!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "fhaviary[llm]>=0.6",  # For info on Message
     "html2text",  # TODO: evaluate moving to an opt-in dependency
     "httpx",
-    "litellm",
+    "litellm>=1.44",  # to prevent sys.stdout on router creation
     "numpy",
     "pybtex",
     "pydantic-settings",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "fhaviary[llm]>=0.6",  # For info on Message
     "html2text",  # TODO: evaluate moving to an opt-in dependency
     "httpx",
-    "litellm",
+    "litellm>=1.44",  # to prevent sys.stdout on router creation
```
jamesbraza, line 26: Our internal envs pins an incompatible range with this, fyi. Maybe we should move the internal one to use LiteLLM router and/or not Anthropic in CI

mskarlin (approved):

pyproject.toml
whitead: Ok - I tried to see if there was a compatible version with both - but there's none.

jamesbraza: I wonder if we can pin `litellm` this way in just dev dependencies.

Do you think the verbose logs you're hitting impact all users (so we want to pin in core deps), or just your machine for development?

PR #364, opened 2024-09-11
More digits on cost

whitead:
Increased resolution on cost. As models have gotten cheaper, it's actually hard to see changes sometimes.


jamesbraza (approved):
Just check the regex 101 for parsing `status` will still work with this

whitead:
Hey @jamesbraza - we couldn't find any regex relying on cost format. If you can recall where that is, yea I can update. Thanks for flagging!

jamesbraza:
> Hey @jamesbraza - we couldn't find any regex relying on cost format. If you can recall where that is, yea I can update. Thanks for flagging!

Yeah sorry I sent that comment from my mobile last night while unable to sleep.

I realized today that actually cost is not part of the regex: https://github.com/Future-House/paper-qa/blob/e1a8dadb9cfd2c7a5193734c9f3a9fda7c09e7b7/paperqa/agents/tools.py#L30-L32

So we are good 👌

PR #365, opened 2024-09-11
Add high-quality settings

mskarlin:
Wanted one with a deeper evidence_k and JSON summaries on.


jamesbraza (approved):
Can you add a `README` table somewhere detailing each of these from a high level?

PR #375, opened 2024-09-11
Testing `MemoryAgent` and timeouts on ldp agents

jamesbraza:
I also took the time to clean up `agents.main` a bit

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -87,7 +87,7 @@ async def test_agent_types(
     agent_test_settings.llm = "gpt-4o-mini"
     agent_test_settings.summary_llm = "gpt-4o-mini"
     agent_test_settings.agent.agent_prompt += (
-        "\n\n Call each tool once in appropriate order and "
+        "\n\nCall each tool once in appropriate order and"
         " accept the answer for now, as we're in debug mode."
     )
     request = QueryRequest(query=question, settings=agent_test_settings)
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -110,7 +110,7 @@ async def test_agent_types(
 
 
 @pytest.mark.asyncio
-async def test_successful_memory_agent() -> None:
+async def test_successful_memory_agent(agent_test_settings: Settings) -> None:
     tic = time.perf_counter()
     memory_id = "call_Wtmv95JbNcQ2nRQCZBoOfcJy"  # Stub value
     memory = Memory(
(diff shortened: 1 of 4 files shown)


tests/test_agents.py
```diff
@@ -104,15 +109,97 @@ async def test_agent_types(
...
+        query=(
+            "Use the tools to answer the question: Q: Acinetobacter lwoffii has"
+            " been evolved in the lab to be resistant to which of these"
+            " antibiotics?\n\nOptions:\nA) gentamicin\nB) Insufficient information"
+            " to answer this question\nC) meropenem\nD) ciproflaxin\nE)"
+            " ampicillin\n\nThe gen_answer tool output is visible to the user, so"
+            " you do not need to restate the answer and can simply terminate if"
+            " the answer looks sufficient. The current status of"
+            " evidence/papers/cost is Status: Paper Count=0 | Relevant Papers=0 |"
+            " Current Evidence=0 | Current Cost=$0.00\n\nTool request message ''"
```
whitead, line 125: $0.0000

whitead: Maybe there is a way to get this from `get_status` method directly? Rather than try to format

jamesbraza: Just added the two extra zeros

> Maybe there is a way to get this from `get_status` method directly? Rather than try to format

I made a `make_status` free function and used it to simplify our test stubs 👌

tests/test_agents.py
```diff
@@ -103,16 +109,84 @@ async def test_agent_types(
         assert response.answer.cost > 0, "Expected nonzero cost"
 
 
+@pytest.mark.flaky(reruns=2, only_rerun=["AssertionError"])
 @pytest.mark.asyncio
-async def test_timeout(agent_test_settings: Settings) -> None:
+async def test_successful_memory_agent() -> None:
```
whitead, line 114:
```suggestion
async def test_successful_memory_agent(agent_test_settings: Settings) -> None:
```

jamesbraza: Thanks for this, appreciated 👍

paperqa/readers.py
```diff
@@ -48,7 +48,7 @@ def chunk_pdf(
 
     if not parsed_text.content:
         raise ImpossibleParsingError(
-            "No text was parsed from the document: either empty or corrupted."
+            f"No text was parsed from the document {doc}, either empty or corrupted."
```
whitead, line 51: Dup?

whitead (approved):
LGTM - think you have a conflict with #389

paperqa/readers.py
jamesbraza: Oh wow nice catch. You are on a roll tonight. I will revert this change

jamesbraza: Man actually bad rebase, this like is a once in a 3 month mistake I make. Nice catch here

jamesbraza: Alright, resolved this

PR #379, opened 2024-09-12
Fix examples in README

taabishm2:
In the README examples
* Fix loop for adding documents to the `Docs` object
* In the async example, wrap await in a function which otherwise throws a syntax error


jamesbraza (approved):
Nice work, thanks

jamesbraza:
https://github.com/Future-House/paper-qa/pull/384 removed the google auth that was messing up CI here. Sorry about that @taabishm2 and thanks again!

PR #386, opened 2024-09-12
LitQA2 downloading and question creation functionality

jamesbraza:
This PR creates functions for:

- Downloading LitQA2 from HuggingFace Hub, facilitated by a `datasets` extra
    - (and CI caching of them)
- Making LitQA2 formatted questions, with a test

### DIFF
--- a/paperqa/litqa.py
+++ b/paperqa/litqa.py
@@ -1,0 +1,207 @@
+"""Utilities for working with LitQA v1 and v2."""
+
+from __future__ import annotations
+
+import random
+import re
+from ast import literal_eval
+from collections.abc import Awaitable, Callable, Sequence
+from enum import IntEnum
+from typing import TYPE_CHECKING
+
+try:
+    from ldp.alg.algorithms import discounted_returns
+except ImportError:
+    discounted_returns = None  # type: ignore[assignment]
+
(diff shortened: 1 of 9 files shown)


paperqa/settings.py
```diff
@@ -482,35 +494,24 @@ def from_name(
...
-                    "litellm_params": {"model": llm, "temperature": self.temperature},
-                }
-            ]
-        }
-
     def get_llm(self) -> LiteLLMModel:
         return LiteLLMModel(
             name=self.llm,
-            config=self.llm_config or self._default_litellm_router_settings(self.llm),
+            config=self.llm_config or make_default_litellm_router_settings(self.llm),
```
mskarlin, line 500: can you put self.temperature as an input into these now that it's a free function?

jamesbraza: Alright, just pushed this fix, thanks again

paperqa/litqa.py
```diff
@@ -0,0 +1,195 @@
+"""Utilities for working with LitQA v1 and v2."""
```
mskarlin, line 1: nice -- it's cool these are centralized in paper-qa now!

mskarlin (approved):

PR #389, opened 2024-09-13
Fixing crash in `chunk_text` for empty file

jamesbraza:
It looks like `chunk_pdf` handles empty PDFs by raising `ImpossibleParsingError`, but `chunk_text` didn't have the same logic. So this PR ports the logic, and expands tests to cover this edge case

### DIFF
--- a/paperqa/readers.py
+++ b/paperqa/readers.py
@@ -48,7 +48,8 @@ def chunk_pdf(
 
     if not parsed_text.content:
         raise ImpossibleParsingError(
-            f"No text was parsed from the document {doc}, either empty or corrupted."
+            f"No text was parsed from the document named {doc.docname!r}, either empty"
+            " or corrupted."
         )
 
     for page_num, page_text in parsed_text.content.items():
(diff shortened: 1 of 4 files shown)


tests/test_agents.py
```diff
@@ -38,8 +38,8 @@ async def test_get_directory_index(agent_test_settings: Settings) -> None:
         "title",
         "year",
     ], "Incorrect fields in index"
-    # paper.pdf + flag_day.html + bates.txt + obama.txt
-    assert len(await index.index_files) == 4, "Incorrect number of index files"
+    # paper.pdf + empty.txt + flag_day.html + bates.txt + obama.txt
+    assert len(await index.index_files) == 5, "Incorrect number of index files"
```
whitead, line 42: Shouldn't the empty not result in an index file?

mskarlin: I think it does but tracks it as importable for future reference.

paperqa/readers.py
```diff
@@ -48,7 +48,7 @@ def chunk_pdf(
 
     if not parsed_text.content:
         raise ImpossibleParsingError(
-            "No text was parsed from the document: either empty or corrupted."
+            f"No text was parsed from the document {doc}, either empty or corrupted."
```
mskarlin, line 51: I think the repr for doc can be quite large, we should limit this

tests/test_agents.py
whitead: interesting 👍

paperqa/readers.py
jamesbraza: Good point, I went with just `doc.docname`

whitead (approved):

tests/test_agents.py
jamesbraza: Alright @whitead I adjusted the test to account for the return of `index_files`, which is a `dict` with a value of `"ERROR"` for `empty.txt`. So now hopefully that's more intuitive

PR #394, opened 2024-09-13
Fixing flaky `test_crossref_retraction_status`

jamesbraza:
- Adds a VCR for the flaky `test_crossref_retraction_status`
- Adds assertions to resolve `union-attr` errors

### DIFF
--- a/tests/test_clients.py
+++ b/tests/test_clients.py
@@ -575,6 +575,7 @@ async def test_ensure_sequential_run_early_stop(
         assert record_indices["early_stop"] != -1, "We should stop early."
 
 
+@pytest.mark.vcr
 @pytest.mark.asyncio
 async def test_crossref_retraction_status() -> None:
     async with aiohttp.ClientSession() as session:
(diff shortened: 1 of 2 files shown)


tests/test_clients.py
```diff
@@ -589,7 +592,9 @@ async def test_crossref_retraction_status():
             title="The Dilemma and Countermeasures of Music Education under the Background of Big Data",
             fields=["title", "doi", "authors", "journal"],
         )
-
-        assert "**RETRACTED ARTICLE** Citation: Jiaye Han." in crossref_details.formatted_citation  # type: ignore[union-attr]
```
jamesbraza, line 593: @geemi725 these `union-attr` errors come from `crossref_client.query` giving a possible `None` value. You can fix by asserting that the query's return is not `None`. Just fyi

jamesbraza:
Realized the test `test_crossref_retraction_status` doesn't check for the presence of `retractions.csv` on local file system, so the cassette is incorrect. Going to let @geemi725 handle this one

PR #401, opened 2024-09-13
Creating `LitQAv2TaskDataset` for agent training/evaluation

jamesbraza:
This PR continues https://github.com/Future-House/paper-qa/pull/386 to make a dataset:

- Extends a `GradablePaperQAEnvironment`, that is able to give a meaningful reward
- Creates a `LitQATaskDataset` base class for making one `GradablePaperQAEnvironment`/question in a LitQA-like dataset
    - Makes up a stub dataset `StubLitQADataset` in testing to confirm it works
- Creates a `LitQAv2TaskDataset` for LitQA v2 from Hugging Face

### DIFF
--- a/paperqa/agents/task.py
+++ b/paperqa/agents/task.py
@@ -1,12 +1,18 @@
+__all__ = [
+    "ENV_NAME",
+    "TASK_DATASET_NAME",
+    "GradablePaperQAEnvironment",
+    "LitQATaskDataset",
+    "LitQAv2TaskDataset",
+    "LitQAv2TaskSplit",
+]
+
 from abc import ABC
 from collections.abc import Awaitable, Callable, Sequence
 from enum import StrEnum
 from typing import TYPE_CHECKING, assert_never
 
-from aviary.env import ENV_REGISTRY, TASK_DATASET_REGISTRY, Frame
-from aviary.env import (
-    TaskDataset as _TaskDataset,
-)
+from aviary.env import ENV_REGISTRY, TASK_DATASET_REGISTRY, Frame, TaskDataset
 from aviary.message import Message
 from aviary.tools import ToolRequestMessage, ToolResponseMessage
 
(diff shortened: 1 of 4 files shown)


paperqa/agents/task.py
```diff
@@ -0,0 +1,196 @@
...
+    def export_frame(self) -> Frame:
+        raise NotImplementedError("Didn't yet need to export a frame.")
+
+
+ENV_NAME = "paperqa-local"
+ENV_REGISTRY[ENV_NAME] = GradableEnvironment.__module__, GradableEnvironment.__name__
+
+
+class LitQATaskDataset(
+    _TaskDataset[GradableEnvironment], ComputeTrajectoryMetricsMixin, ABC
```
mskarlin, line 103: why is this an ABC?

paperqa/agents/task.py
```diff
@@ -0,0 +1,196 @@
...
+    DEFAULT_LABBENCH_HF_HUB_NAME,
+    DEFAULT_REWARD_DISTRIBUTION,
+    LitQAEvaluation,
+    read_litqa_v2_from_hub,
+)
+from paperqa.llms import EmbeddingModel, LiteLLMModel, LLMModel
+from paperqa.types import Answer
+
+from .env import POPULATE_FROM_SETTINGS
+from .env import Environment as _Environment
```
whitead, line 33: Can we rename pqa Environment to be PQAEnvironment or something to remove this conflict

paperqa/agents/task.py
sidnarayanan: I *think* this is an ABC because it doesn't implement `get_new_env()` or `get_new_env_by_idx()`, right?

sidnarayanan (approved):
LGTM!

paperqa/agents/task.py
jamesbraza: It's an ABC because it's not specific to a particular LitQA. It could be v1, v2, or the test stub LitQA I made

I will add a comment for this

jamesbraza: Added this explanation to the class docstring just now

paperqa/agents/task.py
jamesbraza: Nice suggestion, implemented this just now

jamesbraza:
The test failure is a flaky and unrelated test, merging

PR #403, opened 2024-09-13
Add support to read docx files

taabishm2:
Changes:
    - Implement `parse_docx_to_text` function to handle `.doc` and `.docx` files.
    - Use text chunking logic for splitting up docx text after parsing
    
    Bug-fixes:
    - Convert file names in `paperqa/readers.py` to lowercase before checking extensions to ensure case-insensitive matching.


pyproject.toml
```diff
@@ -18,19 +18,25 @@ classifiers = [
 ]
 dependencies = [
     "PyCryptodome",
-    "aiohttp",  # TODO: remove in favor of httpx
+    "aiohttp",
+    # TODO: remove in favor of httpx
```
jamesbraza, line 22: Can you undo the comment wrapping here?

pyproject.toml
```diff
@@ -18,19 +18,25 @@ classifiers = [
...
     "httpx",
-    "litellm>=1.44",  # to prevent sys.stdout on router creation
+    "litellm>=1.44",
+    # to prevent sys.stdout on router creation
     "numpy",
     "pybtex",
     "pydantic-settings",
     "pydantic~=2.0",
     "pymupdf",
+    "python-docx>=1.1.2",
```
jamesbraza, line 36: We are trying to keep dependencies minimal, can you move this to an extra (see `project.optional-dependencies` below), and do lazy imports within `readers.py`

paperqa/readers.py
```diff
@@ -34,6 +35,18 @@ def parse_pdf_to_pages(path: Path) -> ParsedText:
     return ParsedText(content=pages, metadata=metadata)
 
 
+def parse_docx_to_text(path: Path) -> ParsedText:
```
jamesbraza, line 38: Can you allow `path: str | os.PathLike` as the input arg

pyproject.toml
jamesbraza, line 36: Do you know if this package is typed? If yes, can you add it to `mypy` config in `pre-commit`? And if not, can you add it to `mypy` disables in `pyproject.toml`

jamesbraza (commented):
Looking good. Added some comments, and can you make a unit test?

paperqa/readers.py
```diff
@@ -272,12 +285,13 @@ def read_doc(
     parsed_text = None
 
     # start with parsing -- users may want to store this separately
-    if str_path.endswith(".pdf"):
+    if str_path.lower().endswith(".pdf"):
```
jamesbraza, line 288: Let's do this `.lower()` once and add a comment why we're doing it

taabishm2:
> Looking good. Added some comments, and can you make a unit test?

@jamesbraza Good points. Pushed some changes, please take a look

paperqa/readers.py
```diff
@@ -34,6 +35,25 @@ def parse_pdf_to_pages(path: Path) -> ParsedText:
     return ParsedText(content=pages, metadata=metadata)
 
 
+def parse_docx_to_text(path: str | os.PathLike) -> ParsedText:
+    try:
+        import docx
+    except ImportError as e:
+        raise ImportError(
+            "python-docx is required for reading docx files. Please install using:"
+            " `pip3 install paper-qa[python-docx]`."
```
jamesbraza, line 44:
```suggestion
            "Reading docx files requires 'python-docx'. Please install using:"
            " `pip install paper-qa[docx]`."
```

pyproject.toml
```diff
@@ -138,6 +141,7 @@ ignore_missing_imports = true
 # Per-module configuration options
 module = [
     "datasets",  # SEE: https://github.com/huggingface/datasets/issues/3841
+    "docx",  # SEE: https://github.com/python-openxml/python-docx/issues/1432
```
jamesbraza, line 144: Actually, it looks like they are typed: https://github.com/python-openxml/python-docx/blob/master/src/docx/py.typed

Let's add to `pre-commit` `mypy`'s `additional_requirements`, and remove this change here

paperqa/readers.py
```diff
@@ -34,6 +35,25 @@ def parse_pdf_to_pages(path: Path) -> ParsedText:
...
+        ) from e
+    doc = docx.Document(path)
```
jamesbraza, line 46: I think you can fix the `mypy` error here with `Document(str(path))`

tests/test_paperqa.py
```diff
@@ -1018,3 +1018,44 @@ def test_external_doc_index(stub_data_dir: Path) -> None:
...
+    # we limit to only crossref since s2 is too flaky
+    docs.add(
+        doc_path,
+        "Wellawatte et al, A Perspective on Explanations of Molecular Prediction"
+        " Models, XAI Review, 2023",
+        use_doc_details=True,
+        clients={CrossrefProvider},
+        fields=["author", "journal"],
+    )
+    doc_details = next(iter(docs.docs.values()))
```
jamesbraza, line 1045:
```suggestion
    doc_details = next(iter(docs.docs.values()))
    assert isinstance(doc_details, DocDetails)
```

This will inform `mypy` of the correct type, and removes the need for some of the `type: ignore`a

pyproject.toml
```diff
@@ -411,6 +415,7 @@ dev-dependencies = [
     "pytest-timer[colorama]",
     "pytest-xdist",
     "pytest>=8",  # Pin to keep recent
+    "python-docx",
```
jamesbraza, line 418: A couple lines up do this: `paper-qa[datasets,docx,ldp,typing]`

tests/stub_data/paper.pdf
jamesbraza, line 1: Can you make this an additive change only? Let's not delete `paper.pdf`

whitead:
Thanks for your contribution! This is now stale - please open a new PR if you're still interested.

PR #406, opened 2024-09-14
Removing `new_episodes` on `test_crossref_journalquality_fields_filtering`

jamesbraza:
This flag was leading local `pytest -n auto` to remake the fixture a bunch, when we don't want that


jamesbraza:
Turns out this breaks the test

PR #411, opened 2024-09-14
Broken title search ut

nadolskit:
**Summary of Problem**
The expected vs asserted citation count was not matching, causing this UT to specifically fail. 

I noticed if I switched the order of sources from `semantic_scholar, crossref` to `crossref, semantic_scholar` the citation count would be even more dramatically different (similar to the results James was noticing earlier)

More importantly, I found the papers on [semantic scholar](https://www.semanticscholar.org/paper/PaperQA%3A-Retrieval-Augmented-Generative-Agent-for-L'ala-O'Donoghue/7e55d8701785818776323b4147cb13354c820469) (25); which doesn't match our [expected value](https://github.com/Future-House/paper-qa/blob/main/tests/test_clients.py#L69) (23); so that failure makes sense.
The following assertion is failing for the same reason: [196](https://www.semanticscholar.org/paper/Augmenting-large-language-models-with-chemistry-Bran-Cox/354dcdebf3f8b5feeed5c62090e0bc1f0c28db06) assertion vs [191](https://github.com/Future-House/paper-qa/blob/main/tests/test_clients.py#L93) expected.

Vibe checked the results of this change via a CI run: https://github.com/Future-House/paper-qa/actions/runs/10859898582/job/30139826477

This does not fix the issue that James found earlier this evening in test_agents.py.


tests/utils/paper_helpers.py
```diff
@@ -0,0 +1,18 @@
+import re
+
+def compare_formatted_citations(expected: str, actual: str) -> bool:
+    """
+    Compares two formatted citation strings; ignoring the citation_count value.
+
+    :param expected: The expected formatted citation string.
```
jamesbraza, line 7: We use Google Style docstrings, you'll perhaps want to update your IDE

tests/utils/paper_helpers.py
```diff
@@ -0,0 +1,18 @@
...
+    :param actual: The actual formatted citation string.
+    :return: True if the citations match except for the citation count, False otherwise.
+    """
+    # https://regex101.com/r/lCN8ET/1
+    citation_pattern = r"(This article has )\d+( citations?)"
+
+    # between group 1 and 2, replace with the character "n"
+    expected_cleaned = re.sub(citation_pattern, r"\1n\2", expected).strip()
+    actual_cleaned = re.sub(citation_pattern, r"\1n\2", actual).strip()
```
jamesbraza, line 16: I think you got the capturing groups wrong, you want `r"This article has (\d+) citations?"`

We don't care to extract the text, we want to extract the number

Then you can just directly compare the `int`, no need for the weird `r"\1n\2"`

tests/utils/paper_helpers.py
```diff
@@ -0,0 +1,18 @@
...
+def compare_formatted_citations(expected: str, actual: str) -> bool:
+    """
+    Compares two formatted citation strings; ignoring the citation_count value.
+
+    :param expected: The expected formatted citation string.
+    :param actual: The actual formatted citation string.
+    :return: True if the citations match except for the citation count, False otherwise.
+    """
+    # https://regex101.com/r/lCN8ET/1
+    citation_pattern = r"(This article has )\d+( citations?)"
```
jamesbraza, line 12: This regex is closely coupled to `DocDetails`. Can you move this regex to a `ClassVar[str]` on `DocDetails`?

tests/utils/paper_helpers.py
nadolskit: That's fair. 
This original approach using `\1n\2` was because it's focusing on comparing the structure of the citation text while ignoring the citation count itself; just ignoring whatever number is there.

But you make a valid point. I'll get this updated!

jamesbraza: Fwiw (I am not sure your experience levels with regex) going from `r"(This article has )\d+( citations?)"` to `r"This article has (\d+) citations?"` will still match the surrounding text too. A capturing group `()` is here to make it easy to capture and extract a value

And plz point out if we're on the same page already haha. Hope you have a good night

tests/utils/paper_helpers.py
nadolskit: done!

tests/utils/paper_helpers.py
nadolskit: I believe I've addressed this accordingly. Thx James!

tests/utils/paper_helpers.py
nadolskit: No longer required

mskarlin:
The minor variations, i.e. 196 vs 191 is expected, because the papers will get more citations over time. The bigger question is why the cassettes aren’t being used because this should come back with a recorded request each time, I’m in favor of root causing that before we merge this. Maybe we regenerated these cassettes but didn’t update the test?

tests/test_clients.py
```diff
@@ -112,11 +114,44 @@ async def test_title_search(paper_attributes: dict[str, str]) -> None:
...
+        actual_citation_format = re.search(
+            DocDetails.CITATION_COUNT_REGEX_PATTERN,
+            details.formatted_citation,  # type: ignore[union-attr]
+        )
+        assert (
+            expected_citation_format is not None
+        ), "Expected string should match the citation pattern"
+        assert (
+            actual_citation_format is not None
+        ), "Actual string should match the citation pattern"
```
jamesbraza, line 133:
```suggestion
        assert (
            expected_citation_format
        ), "Expected string should match the citation pattern"
        assert (
            actual_citation_format
        ), "Actual string should match the citation pattern"
```

This may resolve the `union-attr` ignore a few lines below

tests/test_clients.py
```diff
@@ -112,11 +114,44 @@ async def test_title_search(paper_attributes: dict[str, str]) -> None:
...
+
+        actual_remaining = (
+            details.formatted_citation[: actual_citation_format.start()]  # type: ignore[union-attr]
+            + details.formatted_citation[actual_citation_format.end() :]  # type: ignore[union-attr]
+        )
+
+        # Assert that the parts of the strings outside the citation count are identical
+        assert (
+            expected_remaining == actual_remaining
+        ), "Formatted citation text should match except for citation count"
```
jamesbraza, line 148: So we are doing a regex search to find a region, then removing that region, and lastly doing an equality check on the front/back end.

I think it would be simpler to do one regex removal of the region, and directly equality compare the remanants.

I also think this is what you originally had 😅 😓  hahaha sorry! I actually realized I misunderstood your original logic last night, my bad here.

What do you think?

tests/test_clients.py
```diff
@@ -112,11 +114,44 @@ async def test_title_search(paper_attributes: dict[str, str]) -> None:
...
+
         assert set(details.other["client_source"]) == set(  # type: ignore[union-attr]
             paper_attributes["source"]
         ), "Should have the correct source"
         for key, value in paper_attributes.items():
-            if key not in {"is_oa", "source"}:
+            if key not in {"is_oa", "source", "formatted_citation"}:
```
jamesbraza, line 154:
```suggestion
            # Equality check all attributes but the ones in the below set
            if key not in {"is_oa", "source", "formatted_citation"}:
```

Just making this logic easier to understand

tests/test_clients.py
nadolskit: definitely! no worries, communicating through PRs can be hard sometimes. Thank you!! adjusting now.

before I merge this I'm also going to look into why the cassette file aren't (hasn't) been updated

paperqa/types.py
```diff
@@ -650,13 +650,19 @@ def __add__(self, other: DocDetails | int) -> DocDetails:
...
 
-            elif (
-                field in {"citation_count", "year", "publication_date"}
-                and self_value is not None
-                and other_value is not None
-            ):
+            elif field in {"citation_count", "year", "publication_date"}:
                 # get the latest data
-                merged_data[field] = max(self_value, other_value)
+                if self_value is None or other_value is None:
```
jamesbraza, line 655: Can you just add a comment explaining the race condition-y thing you encountered that this fixes, either on this line or just above this `if` clause

paperqa/types.py
```diff
@@ -650,13 +650,23 @@ def __add__(self, other: DocDetails | int) -> DocDetails:
...
+                # that value is 0
+                if self_value is None or other_value is None:
+                    merged_data[field] = (
+                        # if self_value is 0, it's evaluated as falsy and will fallback,
+                        # to other_value even if other_value is None
+                        # 0 is a valid value here
+                        self_value
+                        if self_value is not None
+                        else other_value
+                    )
```
jamesbraza, line 667:
```suggestion
                    merged_data[field] = (
                        self_value
                        if self_value is not None  # Dance around 0
                        else other_value
                    )
```

jamesbraza (approved):
Not sure why some requests were missing from fixtures, but otherwise LGTM and thanks for doing this

PR #414, opened 2024-09-14
`DocMetadataClient` can now take instantiated providers and processors

geemi725:
- [x]  updated clients to work with instantiated providers and processors
- [x] retraction test now looks at a stub csv file instead of downloading
- [x] update RetrationDataPostProcessor -> Retra**c**tionDataPostProcessor


tests/test_clients.py
```diff
@@ -572,16 +572,18 @@ async def test_ensure_sequential_run_early_stop(
         assert record_indices["early_stop"] != -1, "We should stop early."
 
 
+@pytest.mark.vcr(record_mode="once")
 @pytest.mark.asyncio
 async def test_crossref_retraction_status():
     async with aiohttp.ClientSession() as session:
+        retractprocessor = RetractionDataPostProcessor("stub_data/stub_retractions.csv")
```
jamesbraza, line 579:
```suggestion
        retract_processor = RetractionDataPostProcessor("stub_data/stub_retractions.csv")
```

tests/test_clients.py
```diff
@@ -572,16 +572,18 @@ async def test_ensure_sequential_run_early_stop(
         assert record_indices["early_stop"] != -1, "We should stop early."
 
 
+@pytest.mark.vcr(record_mode="once")
```
jamesbraza, line 575:
```suggestion
@pytest.mark.vcr
```

tests/test_clients.py
```diff
@@ -101,7 +101,7 @@
 async def test_title_search(paper_attributes: dict[str, str]) -> None:
     async with aiohttp.ClientSession() as session:
         client_list = list(ALL_CLIENTS)
-        client_list.remove(RetrationDataPostProcessor)
+        client_list.remove(RetractionDataPostProcessor)
```
jamesbraza, line 104: Can you add an inline comment why we remove this processor here in the tests?

Alternately, what do you of something like this at the module level?

```python
# Exclude RetrationDataPostProcessor because abc123
TEST_CLIENTS = {ct for ct in ALL_CLIENTS if ct not in {RetrationDataPostProcessor}}
```

jamesbraza (commented):
Nice, thanks for doing this!

jamesbraza (approved):

tests/test_clients.py
```diff
@@ -572,16 +574,20 @@ async def test_ensure_sequential_run_early_stop(
         assert record_indices["early_stop"] != -1, "We should stop early."
 
 
+@pytest.mark.vcr
 @pytest.mark.asyncio
 async def test_crossref_retraction_status():
     async with aiohttp.ClientSession() as session:
+        retract_processor = RetractionDataPostProcessor(
+            "stub_data/stub_retractions.csv"
```
jamesbraza, line 582: Okay running tests locally after this PR, we are now hitting issues as this is a relative path ungrounded in the system.

Can you move this test to intake the fixture `stub_data_dir`:

```python
async def test_crossref_retraction_status(stub_data_dir: Path):
    async with aiohttp.ClientSession() as session:
        retract_processor = RetractionDataPostProcessor(
            stub_data_dir / "stub_retractions.csv"
        )
        ...
```

Also, it looks like this test creates a 163 MB cassette. That is too big. Why is this test making a fixture still, when we are passing in a retractions CSV?

PR #420, opened 2024-09-16
Add stub_data_dir fixture to retraction test

geemi725:


jamesbraza (approved):
Nice work, thanks

tests/cassettes/test_crossref_retraction_status.yaml
```diff
@@ -0,0 +1,53 @@
+interactions:
+  - request:
+      body: null
+      headers: {}
+      method: GET
+      uri: https://api.crossref.org/works?mailto=example@papercrow.ai&query.title=The+Dilemma+and+Countermeasures+of+Music+Education+under+the+Background+of+Big+Data&rows=1&select=DOI,author,container-title,title
+    response:
+      body:
+        string:
+          '{"status":"ok","message-type":"work-list","message-version":"1.0.0","message":{"facets":{},"total-results":4250430,"items":[{"DOI":"10.1155\/2022\/8341966","author":[{"ORCID":"http:\/\/orcid.org\/0000-0003-3902-2613","authenticated-orcid":true,"given":"Jiaye","family":"Han","sequence":"first","affiliation":[{"name":"Pingdingshan
```
jamesbraza, line 10: Just to be clear, there still seems to be a fixture made. This fixture corresponds to the `retractions.csv` already existing on the local filesystem (which makes sense)

One question is, why do we have a request being made at all if the CSV already exists?

geemi725: Oh I think that's for just crossref query not for retractions

PR #424, opened 2024-09-17
Pinning min `pymupdf` version

jamesbraza:
Apparently `import pymupdf` won't work with version 1.23. Would appreciate a second set of eyes here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ classifiers = [
 ]
 dependencies = [
     "PyCryptodome",
+    "PyMuPDF>=1.24.3",  # python -c "import pymupdf" won't run before version 1.24.3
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -18,6 +18,7 @@ classifiers = [
 ]
 dependencies = [
     "PyCryptodome",
```
whitead, line 20: Thought we cut PyCryptodome? Anyway LGTM

whitead (approved):

pyproject.toml
jamesbraza: Let's cut it, but just in a different PR since CI takes too long (lol)

jamesbraza: See https://github.com/Future-House/paper-qa/pull/426

PR #429, opened 2024-09-17
Exposing `"default"` as an import

jamesbraza:
Making client's life's a bit easier

### DIFF
--- a/paperqa/agents/__init__.py
+++ b/paperqa/agents/__init__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import argparse
 import logging
 import os
-from typing import Any
+from typing import Any, Final
 
 from pydantic_settings import CliSettingsSource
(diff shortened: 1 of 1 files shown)


whitead (commented):
Is this really more legible than "default"? Why not just use `None` if you don't like `"default"`?

jamesbraza:
I was trying to avoid redefining magic strings everywhere (both here and in client code), without introducing a breaking change

Changing it to `None` imo would be a breaking change. Are you cool with a breaking change here? I agree `None` is a more sensible value

whitead (approved):
But a caller will now need to import a magic variable name - `AUTOGENERATE_INDEX_NAME` - instead of using the string `"default"`. 

Anyway - it's ok. Just my opinion.

jamesbraza:
I decided lame is not our m/o

PR #430, opened 2024-09-17
`build_index` with defaults

jamesbraza:
Making it easier to use

### DIFF
--- a/paperqa/agents/__init__.py
+++ b/paperqa/agents/__init__.py
@@ -123,7 +123,7 @@ def build_index(
     """Build a PaperQA search index, this will also happen automatically upon using `ask`."""
     settings = get_settings(settings)
     if index_name == "default":
-        index_name = settings.get_index_name()
+        index_name = None
     configure_cli_logging(verbosity=settings.verbosity)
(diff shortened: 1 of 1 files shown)


paperqa/agents/__init__.py
```diff
@@ -116,15 +116,17 @@ def search_query(
...
-    index_name: str,
-    directory: str | os.PathLike,
-    settings: Settings,
+    index_name: str | None = None,
+    directory: str | os.PathLike | None = None,
+    settings: Settings | None = None,
 ) -> SearchIndex:
     """Build a PaperQA search index, this will also happen automatically upon using `ask`."""
+    settings = get_settings(settings)
     if index_name == "default":
```
whitead, line 125: should this check for `None` too?

jamesbraza: I was letting `None` be handled by the wrapped function `get_directory_index`. I can do this better to make it clear

jamesbraza: Just pushed a change to DRY it up with `get_directory_index`

PR #431, opened 2024-09-17
Less common `CROSSREF_XYZ` warnings

jamesbraza:
My logs were getting blown up with:

```none
CROSSREF_MAILTO environment variable not set. Crossref API rate limits may apply.
```

This PR decreases the log occurrences


whitead (approved):
Protect the users

Issue #402, opened 2024-09-13
CLI functionality in Python module


ivanhigueram:
I am loving this library! The README is very through showing the library functions on the CLI. For instance, it shows clearly how to name an index and save it using `pqa -i nanomaterials index`, then we can execute full-text search using `pqa -i nanomaterials search thermoelectrics`. But is not clear to me how to execute the same functionality with the Python module. Can I recreate this by doing something like: 

```python
from paperqa import Settings, ask, search_query

search_query(query="test query", index_name="pqa_index_3cd017b0934549ec2602837942e27025",  index_directory="~/.pqa/indexes/", settings=Settings())
```

I would like some hints on how to name the index and how to execute a search over the index. I would really appreciate your help.

ivanhigueram:
@dosubot this is good, but how can I use a previous `ask` object index to execute a search? that is the part that is not clear in the documentation. 

In the example you provide, I'd have to add each document individually, but I just want to use the ones used in a previous query (i.e. using `agent_query`).

ivanhigueram:
I guess I can get that from the documentation, but again, this is using the CLI, I want to execute this search in Python.

mskarlin:
hey @ivanhigueram -- you should be able to find all your indexes in the `PQA_HOME` directory (defaults to `~/.pqa/indexes/`), you can see the names for each of them there. All answers (no matter from where the originated) will go into the `answers` index, which you can also search via the `search_query `, it'll return `Answer` objects with relevant details for each answer. 

If you have several paper indexes, and you're trying to work out which goes with your settings of interest, just make a new `Settings` object with the paper directory and settings that you previously used and then do: `settings.get_index_name()` -- it'll tell you the name of the index.

jamesbraza:
Hi @ivanhigueram, we just documented this: https://github.com/Future-House/paper-qa/tree/v5.0.5#reusing-index

If you have any other questions, feel free to open a new issue. Thanks for pointing out areas for improvement to us it's appreciated

Issue #413, opened 2024-09-14
Eval paper-qa 2 on lit qa 2


amanpreet692:
Hello,
Thanks for the great work!
I was looking for a way to eval paper-qa 2 on lit-qa 2 and was wondering if you can help me with a couple of queries:
i) Assuming I am able to get the pdfs from the dois in the dataset into a local folder, how do I go about at least running paper qa-2 end to end specifically on the lit-qa 2 multiple choice questions?
ii) Are the answers from paper-qa2 and other systems in the paper on lit-qa 2 questions available publicly?
iii) Lastly, the human and non RAG results on lit-qa 2 in the [lab bench paper](https://arxiv.org/pdf/2407.10362) seem different that in paper-qa2 ... I am not sure if I am missing something, but can you please say a bit more on this?

Thanks!

jamesbraza:
Hello @amanpreet692 thanks for the questions.

> i) Assuming I am able to get the pdfs from the dois in the dataset into a local folder, how do I go about at least running paper qa-2 end to end specifically on the lit-qa 2 multiple choice questions?

We have a bunch if tickets here on making some docs for end-to-end usage of the code. For now, you can look at the unit tests for direct access to the code.

> ii) Are the answers from paper-qa2 and other systems in the paper on lit-qa 2 questions available publicly?

Yeah they are from the LAB-Bench paper, whose data is here: https://github.com/Future-House/lab-bench and https://huggingface.co/datasets/futurehouse/lab-bench/viewer/LitQA2

Let's have part of resolving this ticket be to document this better in the `README`.

> iii) Lastly, the human and non RAG results on lit-qa 2 in the [lab bench paper](https://arxiv.org/pdf/2407.10362) seem different that in paper-qa2 ... I am not sure if I am missing something, but can you please say a bit more on this?

Can you clarify what differences you're observing?

amanpreet692:
Thanks for the prompt reply!
Some follow ups please - 

> We have a bunch if tickets here on making some docs for end-to-end usage of the code. For now, you can look at the unit tests for direct access to the code.

Could you kindly point me to a relevant test case? 

> Yeah they are from the LAB-Bench paper, whose data is here: https://github.com/Future-House/lab-bench and https://huggingface.co/datasets/futurehouse/lab-bench/viewer/LitQA2

Thanks! However, the data set only consists of the ground truth answers if I am not mistaken? 
[...]

Issue #378, opened 2024-09-12
Open Source LLM Server


DrachmaDroysen:
Congrats on the recent release!

I was hoping to test PaperQA2 out with an open source LLM server, but I was struggling to get PaperQA2 to work with one. I have used ollama previously and I was able to set up llamafile for this, but I'm not sure what else is needed to get PaperQA2 to work with llamafile. Could someone please provide some guidance on this?

> PaperQA2 uses an LLM to operate, so you'll need to either set an appropriate [API key environment variable](https://docs.litellm.ai/docs/providers) (i.e. export OPENAI_API_KEY=sk-...) or set up an open source LLM server (i.e. using [llamafile](https://github.com/Mozilla-Ocho/llamafile). **Any LiteLLM compatible model can be configured to use with PaperQA2.**

CrispStrobe:
The documentation already shows how to do it in python, but maybe you mean per cli? I just tried it with:
`pqa --llm "ollama/cas/llama3.1-8b-spaetzle-v74" --llm_config '{"model_list": [{"model_name": "ollama/cas/llama3.1-8b-spaetzle-v74", "litellm_params": {"model": "ollama/cas/llama3.1-8b-spaetzle-v74", "api_base": "http://localhost:11434"}}]}' --summary_llm "ollama/cas/llama3.1-8b-spaetzle-v74" --summary_llm_config '{"model_list": [{"model_name": "ollama/cas/llama3.1-8b-spaetzle-v74", "litellm_params": {"model": "ollama/cas/llama3.1-8b-spaetzle-v74", "api_base": "http://localhost:11434"}}]}' ask 'What is...?'  
`

maykcaldas:
Check [this discussion](https://github.com/Future-House/paper-qa/discussions/753). `ollama` is supported on PaperQA through `litellm`. It can be set using the `Settings` class. 

If you prefer using the `cli`, @CrispStrobe 's answer is correct. In the [discussion](https://github.com/Future-House/paper-qa/discussions/753) I showed how to save this config so you don't need to type this huge command every time

PR #438, opened 2024-09-18
Many documentation improvements

jamesbraza:
- https://github.com/Future-House/paper-qa/issues/378 reveals that our docs [here](https://github.com/Future-House/paper-qa/tree/v5.0.4?tab=readme-ov-file#locally-hosted) were incorrect
- Our docs were getting big so I added a Table of Contents
- Closes https://github.com/Future-House/paper-qa/issues/402
- Adds LitQA v2 docs for https://github.com/Future-House/paper-qa/issues/413
- More logically groups sections

There is still a ton of room for growth here, but this is a good start


whitead (approved):
Thank you so much!!! Approving so we can just get this done and we can iterate more in future PRs

PR #440, opened 2024-09-19
Added some docs and adjusted field names for ease of use in DocDetails

whitead:
Just promoted some fields I thought were useful: `license` and `pdf_url`. These can be provided by S2 client at some point, but for now they come from unpaywall.

Also added to README: https://github.com/Future-House/paper-qa/tree/doc-details-docs?tab=readme-ov-file#using-clients-directly

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ question answering, summarization, and contradiction detection.
   - [Using External DB/Vector DB and Caching](#using-external-dbvector-db-and-caching)
   - [Reusing Index](#reusing-index)
   - [Running on LitQA v2](#running-on-litqa-v2)
+  - [Using Clients Separately](#using-clients-separately)
 - [Where do I get papers?](#where-do-i-get-papers)
   - [Zotero](#zotero)
   - [Paper Scraper](#paper-scraper)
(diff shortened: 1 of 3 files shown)


paperqa/types.py
```diff
@@ -334,6 +334,12 @@ class DocDetails(Doc):
     doi_url: str | None = None
     doc_id: str | None = None
     file_location: str | os.PathLike | None = None
+    license: str | None = Field(
+        default=None,
+        description="string indicating license."
+        " Should refer specifically to pdf_url (since that could be preprint). None means unknown/unset.",
+    )
+    pdf_url: str | None = None
```
jamesbraza, line 342: What do you think of calling this `source_url` over `pdf_url`, as we'll someday generalize beyond PDFs

paperqa/types.py
```diff
@@ -570,17 +576,19 @@ def formatted_citation(self) -> str:
...
+            return (
+                f"{self.citation} This article has {self.citation_count} citations and is"
+                f" from a {self.source_quality_str}."
+            )
+        return f"{self.citation} This article has {self.citation_count} citations."
 
-        if quality is None:
-            return f"{self.citation} This article has {self.citation_count} citations."
+    @property
+    def source_quality_str(self) -> str:
```
jamesbraza, line 587:
```suggestion
    def source_quality_message(self) -> str:
```

Let's not restate the return type in the name, it's redundant

paperqa/types.py
```diff
@@ -570,17 +576,19 @@ def formatted_citation(self) -> str:
...
             )
 
-        quality = (
-            SOURCE_QUALITY_MESSAGES[self.source_quality]
-            if self.source_quality >= 0
-            else None
-        )
+        if self.source_quality_str:
+            return (
+                f"{self.citation} This article has {self.citation_count} citations and is"
```
jamesbraza, line 581: Does `self.citation` have a newline at the end of it?

jamesbraza (approved):

README.md
```diff
@@ -603,6 +604,47 @@ async def evaluate(folder_of_litqa_v2_papers: str | os.PathLike) -> None:
...
+
+client = DocMetadataClient(clients=ALL_CLIENTS)
+details = await client.query(title="Augmenting language models with chemistry tools")
+
+print(details.formatted_citation)
+# Andres M. Bran, Sam Cox, Oliver Schilter, Carlo Baldassari, Andrew D. White, and Philippe Schwaller.
+#  Augmenting large language models with chemistry tools. Nature Machine Intelligence,
+# 6:525-535, May 2024. URL: https://doi.org/10.1038/s42256-024-00832-8,
+# doi:10.1038/s42256-024-00832-8.
+# This article has 243 citations and is from a domain leading peer-reviewed journal.
```
mskarlin, line 622: This should align with the citation count below— maybe we had a regression w recent changes. Or is it a copy-paste issue?

whitead: I copy pasted because README's aren't rendered with wordwrap in code

paperqa/types.py
whitead: should not have that

README.md
whitead: Oh I misunderstood your question! No I manually edited it. This was a timeout issue - don't remember which but one of the clients timed out and the other sent back 20 so I fetched from google scholar

whitead: fixed

paperqa/types.py
whitead: I think `doi_url` should cover the HTML/WEB url and `pdf_url` should directly go to PDF. We may later add `source_url` and then I think `doi_url` would be depcretaed rather than `pdf_url`

jamesbraza: Sounds good

PR #445, opened 2024-09-20
Handling `AttributeError` on structured citation prompt failure

jamesbraza:
Our `ParsingSettings.structured_citation_prompt` extracts a JSON that is supposed to contain the paper's title, DOI, and authors. However, last night I hit a new crash possibility:
1. The paper `Casein Kinase 1&delta; Triggers Giant Ankyrin Expression` had its references on the first page.
2. This led to `ParsingSettings.structured_citation_prompt` extracting a list of authors dicts (one dict per reference).
3. `list[dict]` was not accounted for in `Docs.aadd`, it leads to an `AttributeError` and crashes the indexing process.

This PR handles the `AttributeError` possibility, and logs a warning if it happens

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-import contextlib
 import json
+import logging
 import os
 import re
 import tempfile
@@ -51,6 +51,8 @@ from paperqa.utils import (
     md5sum,
     name_in_text,
 )
+
+logger = logging.getLogger(__name__)
 
 
 # this is just to reduce None checks/type checks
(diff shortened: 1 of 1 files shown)


paperqa/docs.py
```diff
@@ -306,22 +308,32 @@ async def aadd(  # noqa: PLR0912
...
                 if citation_title := citation_json.get("title"):
                     title = citation_title
                 if citation_doi := citation_json.get("doi"):
                     doi = citation_doi
                 if citation_author := citation_json.get("authors"):
                     authors = citation_author
+            except (json.JSONDecodeError, AttributeError):
+                # json.JSONDecodeError: clean_text was not actually JSON
+                # AttributeError: citation_json was not a dict (e.g. a list)
+                logger.exception(
```
mskarlin, line 331: I think we need to consider some of the other modes here that aren't really failures. I think it will be really common (a majority of the cases) that only title and authors can be extracted. In those cases, I don't think we want to show this warning. 

I think we should only show this if it fails on both title and doi, otherwise, if either is present, we should continue without warning the user.

jamesbraza: So if we hit a:
- JSON decode error: we are guaranteed to miss all of title, authors, and DOI
- AttributeError (due to `list` confusion): we are guaranteed to miss all of title, authors, and DOI

Partial failures (as you're trying to protect) can only happen when we successfully decoded JSON and its output was actually a `dict`

jamesbraza: What I did was both add a comment, and adjust the logged warning to reflect this design detail

nadolskit (approved):

mskarlin (approved):
thanks for addressing the comment

PR #479, opened 2024-09-24
Retrying on known Crossref flaky SSL error in `doi_to_bibtex`

jamesbraza:
https://github.com/Future-House/paper-qa/pull/444 added retrying of `Cannot connect to host api.crossref.org:443 ssl:default` to `get_doc_details_from_crossref`. This PR now also adds it to `doi_to_bibtex` as this error just came up there too.


whitead (approved):
nice

PR #483, opened 2024-09-25
Cleaning up and testing `get_directory_index`

jamesbraza:
I found this code somewhat confusing, so I:
1. Added more tests
2. Improved variable names, removed locals, and added type hints

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -343,7 +343,7 @@ async def maybe_get_manifest(
             async with await anyio.open_file(filename, mode="r") as file:
                 content = await file.read()
             records = [DocDetails(**row) for row in csv.DictReader(StringIO(content))]
-            logger.info(
+            logger.debug(
                 f"Found manifest file at {filename} and read {len(records)} records"
                 " from it."
             )
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -433,8 +433,8 @@ async def get_directory_index(
     Args:
         index_name: Override on the name of the index. If unspecified, the default
             behavior is to generate the name from the input settings.
-        sync_index_w_directory: Sync the index with the directory (i.e. delete index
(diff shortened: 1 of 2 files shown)


paperqa/agents/search.py
```diff
@@ -343,7 +343,7 @@ async def maybe_get_manifest(
             async with await anyio.open_file(filename, mode="r") as file:
                 content = await file.read()
             records = [DocDetails(**row) for row in csv.DictReader(StringIO(content))]
-            logger.info(
+            logger.debug(
```
jamesbraza, line 346: This was a bit noisy, so downgraded its level

tests/test_agents.py
```diff
@@ -38,22 +41,54 @@
...
-    # but empty.txt fails to be added
-    path_to_id = await index.index_files
-    assert (
-        sum(id_ != FAILED_DOCUMENT_ADD_ID for id_ in path_to_id.values()) == 4
-    ), "Incorrect number of parsed index files"
-    results = await index.query(query="who is Frederick Bates?")
-    paper_dir = cast(Path, agent_test_settings.paper_directory)
-    assert results[0].docs.keys() == {md5sum((paper_dir / "bates.txt").absolute())}
+    # Use a tempdir here so we can delete files without affecting concurrent tests
+    with tempfile.TemporaryDirectory() as tempdir:
```
mskarlin, line 45: I would have thought this fixture would handle that -- did you see deleting files affecting other tests?

mskarlin (approved):
👍

tests/test_agents.py
jamesbraza: Yeah good question. I just expanded the explanatory comment to address this:

```python
    # Since agent_test_settings is used by other tests, and thus uses the same
    # paper_directory as other tests, we use a tempdir so we can delete files
    # without affecting concurrent tests
```

PR #489, opened 2024-09-26
Refactor to expose `agents.RichHandler`

jamesbraza:
- Refactors out `LOG_VERBOSITY_MAP` for easier to read code
- Decomposes `is_running_under_cli` and `set_up_rich_handler` utilities

### DIFF
--- a/paperqa/agents/__init__.py
+++ b/paperqa/agents/__init__.py
@@ -19,47 +19,44 @@ from .search import SearchIndex, get_directory_index
 
 logger = logging.getLogger(__name__)
 
+LOG_VERBOSITY_MAP = {
+    0: {
+        "paperqa.agents": logging.INFO,
+        "paperqa.agents.helpers": logging.WARNING,
+        "paperqa.agents.main": logging.WARNING,
+        "paperqa.agents.main.agent_callers": logging.INFO,
+        "paperqa.agents.models": logging.WARNING,
+        "paperqa.agents.search": logging.INFO,
+        "anthropic": logging.WARNING,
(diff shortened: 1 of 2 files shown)


paperqa/agents/__init__.py
```diff
@@ -19,66 +19,71 @@
...
-        module_logger.addHandler(rich_handler)
+def configure_cli_logging(verbosity: int = 0) -> None:
+    """Suppress loquacious loggers according to verbosity level."""
+    setup_default_logs()
+    set_up_rich_handler()
 
     for logger_name, logger_ in logging.Logger.manager.loggerDict.items():
         if isinstance(logger_, logging.Logger) and (
-            log_level := verbosity_map.get(min(verbosity, 2), {}).get(logger_name)
+            log_level := LOG_VERBOSITY_MAP.get(min(verbosity, 2), {}).get(logger_name)
```
whitead, line 86: Think this is a bug:

```suggestion
            log_level := LOG_VERBOSITY_MAP.get(min(verbosity, 3), {}).get(logger_name)
```

or maybe use `max(list(LOG_VERBOSITY_MAP.keys())` instead of `3`

whitead (approved):
please fix bug from comment, but other wise LGTM

paperqa/agents/__init__.py
jamesbraza: Oh wow nice catch, let's use `max` to be future proof

jamesbraza: @whitead amazing comment! Great spot

PR #497, opened 2024-09-27
Broken reader ut

nadolskit:
Screenshot 2024-09-27 at 9 04 54 AM

seems consistent now

### DIFF
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -813,10 +813,11 @@ def test_pdf_reader_w_no_match_doc_details(stub_data_dir: Path) -> None:
         next(iter(docs.docs.values())).citation == "Wellawatte et al, XAI Review, 2023"
     )
 
+
 # default: ['method', 'scheme', 'host', 'port', 'path', 'query']
 # this is the default list + body
 # this ensures vcr distinguishes requests with a different body
-@pytest.mark.vcr(match_on=['method', 'scheme', 'host', 'port', 'path', 'query', 'body'])
+@pytest.mark.vcr(match_on=["method", "scheme", "host", "port", "path", "query", "body"])
 def test_pdf_reader_match_doc_details(stub_data_dir: Path) -> None:
     docs = Docs()
     docs.add(
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -814,10 +814,14 @@ def test_pdf_reader_w_no_match_doc_details(stub_data_dir: Path) -> None:
     )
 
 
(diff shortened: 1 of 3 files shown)


tests/test_paperqa.py
```diff
@@ -814,6 +814,10 @@ def test_pdf_reader_w_no_match_doc_details(stub_data_dir: Path) -> None:
     )
 
 
+# default: ['method', 'scheme', 'host', 'port', 'path', 'query']
```
jamesbraza, line 817:
```suggestion
# vcr.VCR default: ('method', 'scheme', 'host', 'port', 'path', 'query')
```

tests/test_paperqa.py
```diff
@@ -814,6 +814,10 @@ def test_pdf_reader_w_no_match_doc_details(stub_data_dir: Path) -> None:
...
+# this is the default list + body
+# this ensures vcr distinguishes requests with a different body
```
jamesbraza, line 819: I am tripping up on "distinguishes". Do you mean this cassette will:
- Generate a new one if the body changes
- Not generate a new one if the body changes

nadolskit: Maybe I can word this better, vcr uses these properties to distinguish how requests are different.
If you look through the cassette, many of the stored requests are identical with the exception of the body (identical method, URI, headers, etc)

for example the calls to https://api.openai.com/v1/embeddings and /v1/chat/completion.

Adding body won't create new cassettes; it just helps vcr correctly match the right request with its response.

tests/test_paperqa.py
```diff
@@ -814,6 +814,10 @@ def test_pdf_reader_w_no_match_doc_details(stub_data_dir: Path) -> None:
     )
 
 
+# vcr.VCR default: ('method', 'scheme', 'host', 'port', 'path', 'query')
+# this is the default list + body
+# Adding body won't create new cassettes; it just helps vcr correctly match the right request with its response.
+@pytest.mark.vcr(match_on=["method", "scheme", "host", "port", "path", "query", "body"])
```
jamesbraza, line 820:
```suggestion
# SEE: https://github.com/kevin1024/vcrpy/blob/v6.0.1/vcr/config.py#L43
VCR_DEFAULT_MATCH_ON = ('method', 'scheme', 'host', 'port', 'path', 'query')


# Adding body won't create new cassettes; it just helps vcr correctly match the right request with its response.
@pytest.mark.vcr(match_on=[*VCR_DEFAULT_MATCH_ON, "body"])
```

jamesbraza (approved):
Nice work here, thank you so much!

PR #510, opened 2024-10-01
Refactor to create `settings.agent.index` grouping

jamesbraza:
This PR groups all index settings into an `AgentSettings` subset called `IndexSettings`.

To aid in this transition it:
- Starts a deprecation cycle for old versions
- Marks fields as `frozen` to flag a read-only type error in client code for `mypy` users

It also removes some defaulted fields from the configs

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -6,11 +6,12 @@ import logging
 import os
 import pathlib
 import pickle
+import warnings
 import zlib
 from collections.abc import Sequence
 from enum import StrEnum, auto
 from io import StringIO
-from typing import Any, ClassVar, cast
+from typing import TYPE_CHECKING, Any, ClassVar, cast
 from uuid import UUID
 
 import anyio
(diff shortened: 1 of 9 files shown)


paperqa/agents/search.py
```diff
@@ -31,12 +32,15 @@
...
 
 from paperqa.docs import Docs
-from paperqa.settings import MaybeSettings, Settings, get_settings
+from paperqa.settings import IndexSettings, get_settings
 from paperqa.types import DocDetails
 from paperqa.utils import ImpossibleParsingError, hexdigest
 
 from .models import SupportsPickle
 
+if TYPE_CHECKING:
```
nadolskit, line 41: if `TYPE_CHECKING` is an import, won't it always be truthy?
(might be a silly question)

jamesbraza: No it's only `True` for type checkers (e.g. `mypy`), and is `False` at runtime

Read here: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

nadolskit (approved):

Issue #499, opened 2024-09-27
Log `This may take a few minutes` less


jamesbraza:
This log line shows up a lot at inference time:

```
2024-09-27 11:44:57,184 - paperqa.agents.search - WARNING - Indexing 5497 files. This may take a few minutes.
```

We need to make this happen less. https://github.com/Future-House/paper-qa/issues/488 will probably alleviate this too

Issue #488, opened 2024-09-26
Refactor to split up `get_directory_index` so every `paper_search` doesn't have to re-index


jamesbraza:
Our current workflow as of `paper-qa==5.0.8` is to re-index every `paper_search`

A more performant and perhaps modular workflow is:
1. Index at the start of the agent
2. Not re-index during each `paper_search`

PR #512, opened 2024-10-01
Not rebuilding `SearchIndex` every `paper_search`

jamesbraza:
- Closes https://github.com/Future-House/paper-qa/issues/488
- Closes https://github.com/Future-House/paper-qa/issues/499

### DIFF
--- a/paperqa/agents/main.py
+++ b/paperqa/agents/main.py
@@ -27,7 +27,7 @@ from paperqa.types import Answer
 from .env import PaperQAEnvironment
 from .helpers import litellm_get_search_query, table_formatter
 from .models import AgentStatus, AnswerResponse, QueryRequest, SimpleProfiler
-from .search import SearchDocumentStorage, SearchIndex
+from .search import SearchDocumentStorage, SearchIndex, get_directory_index
 from .tools import EnvironmentState, GatherEvidence, GenerateAnswer, PaperSearch
 
 if TYPE_CHECKING:
@@ -106,6 +106,8 @@ async def run_agent(
         f" query {query.model_dump()}."
     )
 
+    # Build the index once here, and then all tools won't need to rebuild it
+    await get_directory_index(settings=query.settings)
     if isinstance(agent_type, str) and agent_type.lower() == FAKE_AGENT_TYPE:
         answer, agent_status = await run_fake_agent(query, docs, **runner_kwargs)
     elif tool_selector_or_none := query.settings.make_aviary_tool_selector(agent_type):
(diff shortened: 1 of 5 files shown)


nadolskit (approved):

paperqa/agents/main.py
```diff
@@ -51,7 +51,7 @@ async def agent_query(
     if docs is None:
         docs = Docs()
 
-    search_index = SearchIndex(
+    answers_index = SearchIndex(
```
sidnarayanan, line 54: What's the distinction between a search and answer index? You have `search_or_answers_index` below too

sidnarayanan (approved):
Just one question

paperqa/agents/main.py
jamesbraza: Yeah I actually made this change because I had a similar question in my head. It turns out we instantiate two types of `SearchIndex`:

- "answer" index: we hardcode `index_name="answers"` and store only `Answer.answer`s inside
- "search" index: we don't hardcode the `index_name`, and use this for normal documents

Before this PR, everything was named `search_index` to match the type `SearchIndex`, but this name lost the semantics of the index (e.g. one is for searching, one is for housing answers)

jamesbraza: So, do you think I should do something different here? I have `_index` to match the type `SearchIndex`, and the prefix `answer` or `search` to align with its semantics

sidnarayanan: Ah ok, I missed `index_name="answers"`. Thanks for explaining. I think this makes sense, though a comment by `search_or_answers_index` might help.

jamesbraza: I agree that `search_or_answers_index` leaves one wonder a bit. In the case of `search_or_answers_index`, it's being made to `query` and then immediately discard. So I renamed it to `index_to_query`

paperqa/agents/search.py
```diff
@@ -436,15 +437,14 @@ async def get_directory_index(
     Args:
         index_name: Deprecated override on the name of the index. If unspecified,
             the default behavior is to generate the name from the input settings.
-        sync_index_w_directory: Sync the index (add or delete index files) with the
-            source paper directory.
+        sync_index_w_directory: Opt-out flag to sync the index (add or delete index
+            files) with the source paper directory.
         settings: Application settings.
+        build: Opt-out flag (default is True) to read the contents of the source paper
```
mskarlin, line 443: IMO this logic is a bit complicated for users— what do you think about a “from_directory” class method on the SearchIndex class that you can use when you don’t need to build one? Then you can call that instead of adding an extra flag to this function. Logic looks good otherwise though.

jamesbraza: Looks like you were approximately a few seconds too slow on the PR comment haha.

I didn't decompose `get_directory_index` in this PR because I was trying to be backwards compatible. I will make a downstream PR that does this then

PR #515, opened 2024-10-02
Aviary agent `max_timesteps` and fixed `test_gather_evidence_rejects_empty_docs`

jamesbraza:
Since its addition in https://github.com/Future-House/paper-qa/pull/309, `test_gather_evidence_rejects_empty_docs` was failing for the wrong reason. It was supposed to fail due to empty docs, but it was failing due to a bad `patch`:

```none
ERROR    paperqa.agents.main:main.py:237 Agent <aviary.tools.utils.ToolSelector object at 0x10d5a73b0> failed.
  File "/path/to/paper-qa/paperqa/agents/main.py", line 194, in run_aviary_agent
    obs, tools = await env.reset()
                 ^^^^^^^^^^^^^^^^^
  File "/path/to/paper-qa/paperqa/agents/env.py", line 136, in reset
    self.state, self.tools = self.make_initial_state_and_tools()
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/paper-qa/paperqa/agents/env.py", line 123, in make_initial_state_and_tools
    self.tools = settings_to_tools(
                 ^^^^^^^^^^^^^^^^^^
  File "/path/to/paper-qa/paperqa/agents/env.py", line 71, in settings_to_tools
    tool = Tool.from_function(
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/.venv/lib/python3.12/site-packages/aviary/tools/base.py", line 339, in from_function
    raise ValueError(f"Missing description for parameter {pname}.")
ValueError: Missing description for parameter args.
```

This PR:
- Properly propagates the `__doc__` onto the patched `gen_answer`, so `Tool.from_function` succeeds and the test fails as expected
- Implements `max_timesteps` so the test doesn't cycle forever until it times out
- Moves from `AgentStatus.TIMEOUT` to a more general `AgentStatus.TRUNCATED` to match aviary's general terminology

### DIFF
--- a/paperqa/agents/main.py
+++ b/paperqa/agents/main.py
@@ -140,6 +140,11 @@ async def run_fake_agent(
     ) = None,
     **env_kwargs,
 ) -> tuple[Answer, AgentStatus]:
+    if query.settings.agent.max_timesteps is not None:
+        logger.warning(
+            f"Max timesteps (configured {query.settings.agent.max_timesteps}) is not"
+            " applicable with the fake agent, ignoring it."
+        )
     env = PaperQAEnvironment(query, docs, **env_kwargs)
     _, tools = await env.reset()
     if on_env_reset_callback:
@@ -209,7 +214,14 @@ async def run_aviary_agent(
                 tools=tools,
             )
 
+            timestep, max_timesteps = 0, query.settings.agent.max_timesteps
             while not done:
+                if max_timesteps is not None and timestep >= max_timesteps:
+                    logger.warning(
+                        f"Agent didn't finish within {max_timesteps} timesteps, just answering."
+                    )
+                    await tools[-1]._tool_fn(question=query.query, state=env.state)
+                    return env.state.answer, AgentStatus.FAIL
                 agent_state.messages += obs
                 for attempt in Retrying(
                     stop=stop_after_attempt(5),
@@ -226,6 +238,7 @@ async def run_aviary_agent(
                 obs, reward, done, truncated = await env.step(action)
                 if on_env_step_callback:
                     await on_env_step_callback(obs, reward, done, truncated)
+                timestep += 1
             status = AgentStatus.SUCCESS
     except TimeoutError:
         logger.warning(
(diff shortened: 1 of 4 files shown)


paperqa/agents/main.py
```diff
@@ -121,7 +121,10 @@ async def run_agent(
     else:
         raise NotImplementedError(f"Didn't yet handle agent type {agent_type}.")
 
-    if "cannot answer" in answer.answer.lower() and agent_status != AgentStatus.TIMEOUT:
+    if (
+        "cannot answer" in answer.answer.lower()
```
mskarlin, line 125: not related to this PR but we should really turn this into a function or module level thing

paperqa/agents/main.py
```diff
@@ -140,6 +143,11 @@ async def run_fake_agent(
     ) = None,
     **env_kwargs,
 ) -> tuple[Answer, AgentStatus]:
+    if query.settings.agent.max_timesteps is not None:
```
mskarlin, line 146: jc -- why are we calling the agent steps timesteps? I think of timesteps from physical models (i.e. molecular dynamics) where each iteration is a unit of time, like 4 picoseconds or something. This is more like actionsteps in my mind.

mskarlin (approved):
looks good to me -- couple q's on naming

paperqa/agents/main.py
jamesbraza: I hear what you're saying.

I think one can also encounter vagaries with "step", for example one can wonder does "step" mean:
- One step = agent selection + environment step
- Two steps = agent selection + environment step

I went with timestep to exactly match `ldp.data_structures.Transition.timestep`

jamesbraza: Let me know if you think we should name it otherwise

paperqa/agents/main.py
jamesbraza: Sure, since it's outside the scope of this PR, I made https://github.com/Future-House/paper-qa/pull/516 to handle this

PR #518, opened 2024-10-02
Fixing `IndexSettings.use_absolute_paper_directory` leading to relative index file paths

jamesbraza:
We have a problem where, regardless of `IndexSettings.use_absolute_paper_directory`, index keys are full filesystem paths. This defeats us from sharing indexes across computers.

This PR fixes it by:
- Making it clear what paths are relative and what are absolute
- Careful book keeping revolving around relative paths
- Extending unit tests
    - Also, removes an extra `reset_log_levels` from `test_get_directory_index_w_manifest`

Sadly, indexes made before this PR are *not backwards compatible*, so this PR requires a minor version bump at least.

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -369,25 +369,40 @@ FAILED_DOCUMENT_ADD_ID = "ERROR"
 
 
 async def process_file(
-    file_path: anyio.Path,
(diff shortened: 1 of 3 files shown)


mskarlin (approved):
🔥 yes

whitead (approved):

PR #526, opened 2024-10-04
Made it possible to skip chunking

whitead:
Made a few changes to get closer to using Gemini, where we do not chunk and work with whole documents.

* a chunk_size of `0` will trigger skipping chunking
* deferred embedding documents until retrieval, to avoid embedding docs all the time in tests (or in cases like when you're not doing vector retrieval)
* added a small helper function to truncate before embedding, in case a text is too long

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -401,7 +401,7 @@ class Docs(BaseModel):
         self,
         texts: list[Text],
         doc: Doc,
-        settings: MaybeSettings = None,
+        settings: MaybeSettings = None,  # noqa: ARG002 future proofing
         embedding_model: EmbeddingModel | None = None,
     ) -> bool:
         """
@@ -412,11 +412,6 @@ class Docs(BaseModel):
         Returns:
             True if the doc was added, otherwise False if already in the collection.
         """
-        all_settings = get_settings(settings)
-
-        if embedding_model is None:
-            embedding_model = all_settings.get_embedding_model()
-
         if doc.dockey in self.docs:
             return False
         if not texts:
@@ -424,11 +419,13 @@ class Docs(BaseModel):
         # 1. Calculate text embeddings if not already present, but don't set them into
         # the texts until we've set up the Doc's embedding, so callers can retry upon
         # OpenAI rate limit errors
-        text_embeddings: list[list[float]] | None = (
-            await embedding_model.embed_documents(texts=[t.text for t in texts])
-            if texts[0].embedding is None
-            else None
-        )
+        text_embeddings: list[list[float]] | None = None
+        if embedding_model:
+            text_embeddings = (
+                await embedding_model.embed_documents(texts=[t.text for t in texts])
+                if texts[0].embedding is None
+                else None
+            )
         # 2. Now we can set the text embeddings
         if text_embeddings is not None:
             for t, t_embedding in zip(texts, text_embeddings, strict=True):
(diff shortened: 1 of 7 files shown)


paperqa/llms.py
```diff
@@ -65,9 +65,26 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
...
 
+    def _truncate_if_large(self, texts: list[str]) -> list[str]:
+        """Truncate texts if they are too large specifically for an openai model."""
+        if self.name not in {
+            "text-embedding-3-small",
+            "text-embedding-3-large",
+            "text-embedding-ada-002",
+        }:
+            return texts
+        max_tokens = 8192
```
mskarlin, line 76: we should probably move this and the model name list to module level variables

paperqa/docs.py
```diff
@@ -412,23 +412,20 @@ async def aadd_texts(
         Returns:
             True if the doc was added, otherwise False if already in the collection.
         """
-        all_settings = get_settings(settings)
-
-        if embedding_model is None:
```
mskarlin, line 417: Why cut the option to pull the model from settings here? I think we can still support it and skip if the chunksize = 0.

paperqa/llms.py
```diff
@@ -65,9 +65,26 @@ class LiteLLMEmbeddingModel(EmbeddingModel):
...
+        if any(len(t) > maybe_too_large for t in texts):
+            enct = tiktoken.get_encoding("cl100k_base")
+            enc_batch = enct.encode_ordinary_batch(texts)
+            return [enct.decode(t[:max_tokens]) for t in enc_batch]
+        return texts
+
     async def embed_documents(
         self, texts: list[str], batch_size: int = 16
     ) -> list[list[float]]:
+        texts = self._truncate_if_large(texts)
```
mskarlin, line 87: I doubt this will practically affect retrieval much (since 8k is already so large), but I wonder if we should support splitting + embedding then taking the averaging back into a single embedding for a large document.

paperqa/docs.py
whitead: I switched to embed later - if embedding_model isn't present, we will no longer fetch from settings. That means the settings isn't used anywhere in the code.

paperqa/llms.py
whitead: Fine - I wrote it correctly using litellm cost map

paperqa/llms.py
whitead: Oh I understand what you're saying - I'm not sure that works though. Would have to experiment

paperqa/docs.py
mskarlin: Right, I'm saying that instead of the flag being the presence of `embedding_model` in the entrypoint, what do you think of using chunksize=0 from settings? Then that way a user could still use this method just by inputting a settings object and they don't need to instantiate their embedding_model before passing.

whitead: Chunk_size / skipping chunking does come from settings.

The change I made here is unrelated to this. I just made it so that embedding is not done when a document is added, since it's kind of wasteful for tests and rate limits. Now we embed texts only when calling gather evidence. A side-effect of that change is the only time we embed is if an explicit embedding_model is passed

mskarlin: I see -- I was imagining that a user would only want to avoid the early embedding if they wanted to do full-doc embeddings.

One other edge case, when a user is building an index, we use `aadd`/`aadd_texts` for each doc as we index it. There we probably want to pre-embed because we save individual `Docs` with one paper each, and combine the candidate matching `Docs` after a paper search. If we only embedded at gather evidence time, then we'd need to either embed each time, or split back out into the constituent single-paper `Docs` objects and re-save them at that step. 

To your point that definitely can push the embedding rate limit, but I think it makes sense for the local index. This PR totally supports this use case if you just insert an `embedding_model`, but in practice we're only  passing a `Settings` object into our `aadd` method call in `search.process_file`. So it was clean to extract the model out of the settings there.

I think all we need to do instead is to build an `embedding_model` in `search.process_file` and pass that into `aadd` and then this will still work.

whitead: OK - added new setting (off by default) to defer_embeddings. It is default on for debug setting (to avoid so many embedding calls) and if we get to a point where we have extremely large indices, it may be more efficient to defer embedding.

mskarlin (approved):

PR #531, opened 2024-10-04
Disallowing confusing `None` from `IndexSettings.index_directory`, and `IndexSettings.get_named_index_directory`

jamesbraza:
The `None` from `IndexSettings.index_directory` was a failover to the default `IndexSettings.index_directory` inside of `SearchIndex`. This has two issues:
- That failover "desynchronizes" `IndexSettings` from `SearchIndex`, which tripped me up
- I think `None` for `index_directory` could imply "no index dir" as opposed to "default index dir"

Also, I implemented a convenience utility `IndexSettings.get_named_index_directory`, with a test


paperqa/settings.py
```diff
@@ -330,6 +330,20 @@ class IndexSettings(BaseModel):
...
+        Raises:
+            ValueError: If the index name was unset, because otherwise the name is
+                autogenerated.
+        """
+        if self.name is None:
+            raise ValueError(
+                "Getting a named index directory requires an index name to have been"
+                " specified, please specify a name."
+            )
+        return anyio.Path(self.index_directory) / self.name
```
whitead, line 345: Why is this a `anyio.Path` and `index_directory` is `os.PathLike`? Are those types compatible?

whitead (approved):

paperqa/settings.py
jamesbraza: Yeah good question, it comes down to "be vague in what you accept, and specific in what you return"

`index_directory`  is an input that allows general types of `str | os.PathLike`:
- `str` is for places like configs
- `os.PathLike` is basically an `ABC` for `pathlib.Path`-like objects

The return is a concretion of `os.PathLike`, which is an `anyio.Path`. We use `anyio.Path` so we can do `async` operations with the `Path`

PR #532, opened 2024-10-04
Add router_kwargs in separate control flow step

mskarlin:
We had coupled the default `router_kwargs` insertion with the presence of the `model_list` key -- but we needed to do a separate check for it. 

Adds a test that ensures it's present via the defaults.

### DIFF
--- a/paperqa/llms.py
+++ b/paperqa/llms.py
@@ -524,8 +524,15 @@ class LiteLLMModel(LLMModel):
                         ),
                     }
                 ],
-                "router_kwargs": {"num_retries": 3, "retry_after": 5, "timeout": 60},
             } | data.get("config", {})
+
+        if "router_kwargs" not in data.get("config", {}):
+            data["config"]["router_kwargs"] = {
+                "num_retries": 3,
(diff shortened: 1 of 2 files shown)


paperqa/llms.py
```diff
@@ -524,8 +524,15 @@ def maybe_set_config_attribute(cls, data: dict[str, Any]) -> dict[str, Any]:
...
                 ],
-                "router_kwargs": {"num_retries": 3, "retry_after": 5, "timeout": 60},
             } | data.get("config", {})
+
+        if "router_kwargs" not in data.get("config", {}):
+            data["config"]["router_kwargs"] = {
+                "num_retries": 3,
+                "retry_after": 5,
+                "timeout": 60,
+            }
```
jamesbraza, line 534: Can we somehow avoid redefining these numbers twice (other place being `settings.make_default_litellm_router_settings`)? Perhaps we can use a module level variable to share the parameters

jamesbraza: NEVERMIND! 😆

jamesbraza (approved):

PR #537, opened 2024-10-06
Add limits+coredis to mypy

mskarlin:
I forgot to add to mypy, not it's added. It revealed a couple of type signatures to clean up.

### DIFF
--- a/paperqa/rate_limiter.py
+++ b/paperqa/rate_limiter.py
@@ -25,6 +25,13 @@ logger = logging.getLogger(__name__)
 
 GLOBAL_RATE_LIMITER_TIMEOUT = float(os.environ.get("RATE_LIMITER_TIMOUT", "60"))
 
+MATCH_ALL = None
+MatchAllInputs = Literal[None]
+MATCH_MACHINE_ID = "<machine_id>"
+
+FALLBACK_RATE_LIMIT = RateLimitItemPerSecond(3, 1)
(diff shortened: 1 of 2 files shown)


paperqa/rate_limiter.py
```diff
@@ -301,7 +310,7 @@ async def try_acquire(
         """Returns when the limit is satisfied for the namespace_and_key.
 
         Args:
-            namespace_and_key (:obj:`tuple[str, str | MatchAllInputs]`): is
+            namespace_and_key (:obj:`tuple[str, str]`): is
```
jamesbraza, line 313: What is this `:obj:` syntax here? Normally it's good to not restate types if they're in the docstring already

jamesbraza (approved):
Thanks for doing this!

paperqa/rate_limiter.py
mskarlin: it's in the style guide for the google doc strings, I think for sphinx compatibility

jamesbraza: Fwiw, either Sphinx autodoc or napoleon (I can't remember which) will pull in type hints if they are there

https://google.github.io/styleguide/pyguide.html#383-functions-and-methods note that it does not restate type hints in `Args` or `Returns`

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html in this one, note there's not type hints at all

Anyways, moral of the story I have seen it where people later update or expand type hints, but then don't update the repeated version in the docstring `Args` or `Returns`

PR #538, opened 2024-10-07
Controlling for `pymupdf` version in `test_pdf_reader_match_doc_details` VCR

jamesbraza:
As we see in [this CI run](https://github.com/Future-House/paper-qa/actions/runs/11206115231/job/31146413938?pr=534), `test_pdf_reader_match_doc_details` has failed again. But it has a VCR, so why was it failing?

I traced it to going from `pymupdf==1.24.10` to `pymupdf==1.24.11` invalidating the VCR. So we can fix this now-failing test in one of two ways:
1. Pinning `pymupdf` in dev dependencies
2. Adding `pymupdf` to VCR key for this test

In this PR, I did the folowing:
- Adjusted the `test_pdf_reader_match_doc_details`'s VCR name to include `pymupdf.__version__` in the name
- Removed less specific `match_on` configuration from our `conftest`, and regenerated all VCR cassettes


tests/conftest.py
```diff
@@ -45,7 +47,6 @@ def fixture_vcr_config() -> dict[str, Any]:
             ANTHROPIC_API_KEY_HEADER,
             "cookie",
         ],
-        "match_on": ["method", "host", "path", "query"],
```
jamesbraza, line 48: @mskarlin do you know why we overrode the default `match_on` here? The default has 6 entries, why did we down-specify to 4?

tests/cassettes/TestTaskDataset.test_tool_failure.yaml
jamesbraza, line 1: Huge win here lol, a latent bug was captured in this

nadolskit (approved):

tests/cassettes/test_pdf_reader_match_doc_details_pymupdf_v1.24.11.yaml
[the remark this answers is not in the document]
jamesbraza: What happens is when a future dev/Renovate runs `uv.lock` (and assuming a new `pymupdf` version is out), CI will break until you regenerate this cassette

You're right that other tests don't have this naming special case, but it's because other tests were unaffected by the `pymupdf` version

PR #551, opened 2024-10-09
Making parsing resistant to failed inference of citations

### LINKED ISSUE #546: JSON decode error when adding docs
Hello,

I've got this error when I try to add document with 'add' method from Doc() object : 

```
Failed to parse all of title, DOI, and authors from the ParsingSettings.structured_citation_prompt's response  {
"title": "La politique culturelle de la Communauté française de Belgique : origines, évolutions et enjeux",
"authors": ["Auby, Bénédicte"],
"doi": null
}

Explication :

Le titre de l'ouvrage est "La politique culturelle de la Communauté française de Belgique : origines, évolutions et enjeux".

Les auteurs de l'ouvrage sont Bénédicte Auby. Comme il y a un seul auteur, je l'ai renvoyé sous la forme d'une liste à une entrée.

Le DOI (Digital Object Identifier) n'était pas spécifié dans la citation MLA, donc je l'ai renvoyé comme null.</s>, consider using a manifest file or specifying a different citation prompt.
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\docs.py", line 321, in aadd
    citation_json = json.loads(clean_text)
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\json\__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\json\decoder.py", line 340, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 7 column 1 (char 158)
```

Is the JSON error is a consequence of lack of DOI in text ?

whitead:
Fixes #546.

Tried to improve JSON extraction and reduced exception to warning (@mskarlin  - not sure why this was an exception?).

Example colab
https://colab.research.google.com/drive/1dXEuhvmnC-0Pzwedy0HIEt6TRli7k3kB#scrollTo=vbOB5XjbO2dd

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -314,13 +314,13 @@ class Docs(BaseModel):
                 data={"citation": citation},
                 skip_system=True,
             )
-            clean_text = result.text.strip("`")
-            if clean_text.startswith("json"):
-                clean_text = clean_text.replace("json", "", 1)
-            # there should be no nesting here, so we can just split on curlies
-            # Note - this is why we cannot extract it to a general method
-            # because JSON in general has nested curlies
-            clean_text = clean_text.split("{", 1)[-1].split("}", 1)[0]
+            # This code below tries to isolate the JSON
+            # based on observed messages from LLMs
+            # it does so by isolating the content between
+            # the first { and last } in the response.
+            # Since the anticipated structure should  not be nested,
+            # we don't have to worry about nested curlies.
+            clean_text = result.text.split("{", 1)[-1].split("}", 1)[0]
             clean_text = "{" + clean_text + "}"
             try:
                 citation_json = json.loads(clean_text)
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -318,10 +318,12 @@ class Docs(BaseModel):
             if clean_text.startswith("json"):
                 clean_text = clean_text.replace("json", "", 1)
             # there should be no nesting here, so we can just split on curlies
+            # Note - this is why we cannot extract it to a general method
+            # because JSON in general has nested curlies
             clean_text = clean_text.split("{", 1)[-1].split("}", 1)[0]
             clean_text = "{" + clean_text + "}"
             try:
-                citation_json = json.loads(clean_text.split("}", 1)[0] + "}")
+                citation_json = json.loads(clean_text)
                 if citation_title := citation_json.get("title"):
                     title = citation_title
                 if citation_doi := citation_json.get("doi"):
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -317,8 +317,11 @@ class Docs(BaseModel):
             clean_text = result.text.strip("`")
             if clean_text.startswith("json"):
                 clean_text = clean_text.replace("json", "", 1)
+            # there should be no nesting here, so we can just split on curlies
+            clean_text = clean_text.split("{", 1)[-1].split("}", 1)[0]
+            clean_text = "{" + clean_text + "}"
             try:
-                citation_json = json.loads(clean_text)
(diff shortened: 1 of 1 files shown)


paperqa/docs.py
```diff
@@ -317,6 +317,11 @@ async def aadd(  # noqa: PLR0912
             clean_text = result.text.strip("`")
             if clean_text.startswith("json"):
                 clean_text = clean_text.replace("json", "", 1)
+            # there should be no nesting here, so we can just split on curlies
+            # Note - this is why we cannot extract it to a general method
+            # because JSON in general has nested curlies
+            clean_text = clean_text.split("{", 1)[-1].split("}", 1)[0]
+            clean_text = "{" + clean_text + "}"
```
jamesbraza, line 324: It looks like in that linked issue, the model responded with JSON and also an explanation. The trailing explanation broke `json.loads`.

I would say your comment here does not seem to track with what is happening. The code is stripping the leading/trailing text around the JSON, but your comment talks about nesting of JSON

Can you update the comment to say that its just stripping leading/trailing text?

paperqa/docs.py
```diff
@@ -328,9 +333,7 @@ async def aadd(  # noqa: PLR0912
             except (json.JSONDecodeError, AttributeError):
                 # json.JSONDecodeError: clean_text was not actually JSON
                 # AttributeError: citation_json was not a dict (e.g. a list)
-                # NOTE: we only want to warning log when we fail on title and
-                # authors we may not care if we fail on DOI here
```
jamesbraza, line 332: Why delete this comment? It adds value imo, DOI extraction is the only critical one

whitead: It's a restatement of what's in the logger warning imo

jamesbraza: So the thinking here is extracting DOI is really valuable, extracting title and authors is a nice-to-have

Why do you think it's a restatement? The message below doesn't talk about DOI being more important than title and authors

paperqa/docs.py
whitead: I made a longer docstring explaining what clean_text is trying to do.

My comment was specifically about the fact that the code I wrote will not work on JSON in general, anticipating that the code looks incorrect. I wrote a longer comment about the rationale and goals of the code

jamesbraza: Yeah nice work on the comment, it makes sense 👍

jamesbraza (approved):
Approving, note the

paperqa/docs.py
jamesbraza: So I decided this comment can be removed, as it's suboptimal.

Basically:
- Succeeding in an extraction here saves us from trying to infer this information in downstream logic
- We get different gains from extracting title, author, and DOI

This comment was trying to make one aware of that, but clearly it's not doing its job as an informative comment.

It's unclear to me if a suboptimal comment making one aware of some behavior is better or worse than no comment

whitead: Oh - yea I guess you're right. Then I deleted the comment because it doesn't reflect the code which previously just threw an exception with any failure. Maybe @mskarlin -  can you comment on if the comment was correct or the code was correct?

jamesbraza: Fwiw it didn't throw an exception earlier, it just logged a stack trace and continued. The user didn't actually experience a crash, just the code took a different path and inferred the title/author/DOI later

whitead: Oh - I thought `logger.exception` actually created an exception. Thanks - didn't realize that.

PR #552, opened 2024-10-09
Exposed log verbosity configuration function

jamesbraza:
Callers may want to configure logs manually themselves, so this provides flexibility for that


nadolskit (approved):

mskarlin (approved):
Nice -- we shoulda done this initially

PR #553, opened 2024-10-09
Allowed `None` log verbosity

jamesbraza:
Allowing `None` log verbosity so we can avoid setting a log verbosity for PaperQA logs entirely


jamesbraza:
Decided this was a bad idea, and I have extracted some of the value-add changes here into https://github.com/Future-House/paper-qa/pull/554

PR #561, opened 2024-10-10
Refactoring `LiteLLMModel` before removing `Router`

jamesbraza:
I wanted to make sure I am not going to be breaking anything, so I:
1. Made it clear what comes from `litellm`
2. Expanded `LiteLLMModel` tests to check the `config` is used for `acompletion`

### DIFF
--- a/tests/test_llms.py
+++ b/tests/test_llms.py
@@ -4,6 +4,7 @@ import pickle
 import pytest
 
 from paperqa import LiteLLMModel
+from paperqa.llms import Chunk
 from tests.conftest import VCR_DEFAULT_MATCH_ON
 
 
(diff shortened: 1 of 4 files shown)


tests/test_llms.py
```diff
@@ -65,6 +67,26 @@ async def ac(x) -> None:
...
+            config={
+                "model_list": [
+                    {
+                        "model_name": "gpt-4o-mini",
+                        "litellm_params": {"model": "gpt-4o-mini", "max_tokens": 3},
+                    }
+                ]
+            },
+        )
+        chunk = await llm.acomplete("Please tell me a story")  # type: ignore[call-arg]
```
jamesbraza, line 84: @mskarlin it looks like the `rate_limited` decorator breaks client typing as well. We should fix this at some point

nadolskit (approved):

PR #520, opened 2024-10-03
Add rate limits for LLMs and Embedding Models

mskarlin:
LiteLLM's rate limits weren't suitable for PaperQA in that we wanted rate limits that could span models. This PR adds them in with both an in-memory based rate limiter, as well as a Redis based one for rate limiting across processes. 

The implementation adds a new decorator, `rate_limited` to the `LiteLLMModel` class across all 4 inference methods. This decorator checks for rate limits before (with prompt tokens) and after (with completion tokens) inference. If tokens aren't known (like when using the *_iter methods), then it uses an estimate with character count divided by the `CHARACTERS_PER_TOKEN` constant (4). It's technically possible, with low rate limits that don't correspond with a max_token cutoff, that the completions tokens could exceed your maximum allowable tokens in your window of time (say your limit is 20 tokens per second and 100 tokens come back). In this case the rate limiter will wait it out such that your amortized rate will fall back to 20 tokens per second. 

The implementation is similar for  the `LiteLLMModel` and the `LiteLLMEmbeddingModel`, where you give the `config` attribute a key for rate limits like this: 
```
llm = LiteLLMModel(name='gpt-4o-mini', config={
            "rate_limit": {"gpt-4o-mini": RateLimitItemPerSecond(20, 3)},
        })
```
or 
```
llm = LiteLLMModel(name='gpt-4o-mini', config={
            "model_list": [
                {
                    "model_name": "gpt-4o-mini",
                    "litellm_params": {
                        "model": "gpt-4o-mini",
                        "temperature": 0,
                    },
                }
            ],
            "rate_limit": {"gpt-4o-mini": RateLimitItemPerSecond(20, 1)},
        },
    })
```
and for the embedding model:
```
embedding = LiteLLMEmbeddingModel(name='text-embedding-3-small', config={"rate_limit": RateLimitItemPerSecond(20, 5)})
```

PR #563, opened 2024-10-10
Ability to bypass usage of `litellm.Router`

jamesbraza:
There are multiple outstanding issues with LiteLLM's `Router` as of `litellm==1.48.10`:

1. It doesn't support TPM: https://github.com/BerriAI/litellm/issues/5783
2. The `Router` has several race conditions that keep killing our training runs

https://github.com/Future-House/paper-qa/pull/520 in-housed rate limits management, which was most of the reason we adopted the `Router` in https://github.com/Future-House/paper-qa/pull/315. Now the `Router` just enables model grouping, which we don't really use internally at FutureHouse. Paired with the `Router`'s race conditions, we decided to make usage of the `Router` optional.

So, this PR:
- Creates a back door to bypass `Router` usage, instead using LiteLLM's free functions
- Removes "router" mentions from the code base where stuff is actually just "model list" specific


paperqa/llms.py
```diff
@@ -510,21 +518,37 @@ def get_litellm_retrying_config(timeout: float = 60.0) -> dict[str, Any]:
...
+    config: dict = Field(
+        default_factory=dict,
+        description=(
+            "Configuration of this model containing several important keys. The"
+            " optional `model_list` key stores a list of all model configurations"
+            " (SEE: https://docs.litellm.ai/docs/routing). The optional"
+            " `router_kwargs` key is keyword arguments to pass to the Router class."
+            " Inclusion of a key `pass_through_router` with a truthy value will lead"
+            " to using not using LiteLLM's Router, instead just LiteLLM's free"
+            f" functions (see {PassThroughRouter.__name__}). The optional `rate_limit`"
```
mskarlin, line 546: We might note that the rate limiter will only work if using the router based config.

paperqa/llms.py
```diff
@@ -622,19 +652,22 @@ async def acomplete_iter(  # type: ignore[override]
...
         self, messages: Iterable[dict[str, str]]
     ) -> Chunk:
-        response = await self.router.acompletion(self.name, messages)
+        response = await self.router.acompletion(self.name, list(messages))
         return Chunk(
-            text=response.choices[0].message.content,
-            prompt_tokens=response.usage.prompt_tokens,
-            completion_tokens=response.usage.completion_tokens,
+            text=cast(litellm.Choices, response.choices[0]).message.content,
+            prompt_tokens=response.usage.prompt_tokens,  # type: ignore[attr-defined]
```
mskarlin, line 658: these are defined even without a router, right?

paperqa/llms.py
```diff
@@ -622,19 +652,22 @@ async def acomplete_iter(  # type: ignore[override]
...
         )
 
     @rate_limited
     async def achat_iter(  # type: ignore[override]
         self, messages: Iterable[dict[str, str]]
     ) -> AsyncIterable[Chunk]:
         completion = await self.router.acompletion(
-            self.name, messages, stream=True, stream_options={"include_usage": True}
+            self.name,
+            list(messages),
```
mskarlin, line 668: I think if messages is a generator here, this will pull every value into the list which might not be desired here.

mskarlin (approved):
Couple minor comments, looks great!

paperqa/llms.py
jamesbraza: The issue is `litellm` types with `list[Message]`, so I was adhering to their type.

Earlier, we didn't type `self.router`, so this snuck under the radar

jamesbraza: What do you think we should do here?

paperqa/llms.py
jamesbraza: Nice suggestion, added a note here about this. We discussed in person, but rate limits apply to both cases

paperqa/llms.py
jamesbraza: We discussed this in-person

paperqa/llms.py
jamesbraza: Yes they are, and I have two different unit tests now proving it 😈

krrishdholakia:
> The Router has several race conditions that keep killing our training runs

Hi @jamesbraza what are these race conditions?

PR #587, opened 2024-10-14
Unsilencing manifest read failure when no CSV header is present

jamesbraza:
It turns out there were two problems with our manifest loading:

1. If no CSV header is present, our `maybe_get_manifest` silently fails
2. It used `logging.error`, which bypasses our logging handlers

This PR fixes both issues

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -11,7 +11,6 @@ import warnings
 import zlib
 from collections.abc import Awaitable, Callable, Collection, Sequence
 from enum import StrEnum, auto
-from io import StringIO
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 from uuid import UUID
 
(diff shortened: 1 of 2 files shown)


paperqa/agents/search.py
```diff
@@ -359,18 +358,27 @@ async def maybe_get_manifest(
...
+                )
             logger.debug(
-                f"Found manifest file at {filename} and read {len(records)} records"
-                " from it."
+                f"Found manifest file at {filename}, read {len(records)} records"
+                f" from it, which maps to {len(file_loc_to_records)} locations."
             )
-            return {str(r.file_location): r for r in records if r.file_location}
         except FileNotFoundError:
-            logging.warning(f"Manifest file at {filename} could not be found.")
```
jamesbraza, line 369: @mskarlin were you intentionally using `logging.warning` as opposed to `logger.warning` here?

paperqa/agents/search.py
```diff
@@ -359,18 +358,27 @@ async def maybe_get_manifest(
...
-            records = [DocDetails(**row) for row in csv.DictReader(StringIO(content))]
+            records = [DocDetails(**r) for r in csv.DictReader(content.splitlines())]
+            file_loc_to_records = {
+                str(r.file_location): r for r in records if r.file_location
+            }
+            if not file_loc_to_records:
+                raise ValueError(  # noqa: TRY301
+                    f"No mapping of file location to details extracted from manifest"
+                    f" file {filename}."
+                )
```
jamesbraza, line 369: Without this check, manifest CSV's without a header line would silently pass this

nadolskit (approved):

paperqa/agents/search.py
mskarlin: nah -- must have just been a mental lapse

jamesbraza: Okay thaanks

mskarlin (approved):

PR #590, opened 2024-10-15
Add callback support in settings and tools

nadolskit:
Adds callback support to Settings
Tools can invoke these callbacks

I wrote this in a way so that the callbacks are immutable, future flexibility, and direct.
There will likely be more tools using this infrastructure in the future.

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -1,5 +1,6 @@
 """Base classes for tools, implemented in a functional manner."""
 
+import asyncio
 import inspect
 import logging
 import re
@@ -191,7 +192,12 @@ class GatherEvidence(NamedTool):
             raise EmptyDocsError("Not gathering evidence due to having no papers.")
 
         if f"{self.TOOL_FN_NAME}_initialized" in self.settings.callbacks:
-            await asyncio.gather(*(c(state) for c in self.settings.callbacks[f"{self.TOOL_FN_NAME}_initialized"]))
+            await asyncio.gather(
+                *(
+                    c(state)
+                    for c in self.settings.callbacks[f"{self.TOOL_FN_NAME}_initialized"]
+                )
+            )
 
         logger.info(f"{self.TOOL_FN_NAME} starting for question {question!r}.")
         original_question = state.answer.question
@@ -227,9 +233,14 @@ class GatherEvidence(NamedTool):
         )
 
         if f"{self.TOOL_FN_NAME}_completed" in self.settings.callbacks:
-            callback_list = self.settings.callbacks[f"{self.TOOL_FN_NAME}_completed"]
-            for callback in callback_list:
-                await callback(state)
+            await asyncio.gather(
+                *(
+                    callback(state)
+                    for callback in self.settings.callbacks[
+                        f"{self.TOOL_FN_NAME}_completed"
+                    ]
+                )
+            )
 
         return f"Added {l1 - l0} pieces of evidence.{best_evidence}\n\n" + status
 
@@ -263,9 +274,14 @@ class GenerateAnswer(NamedTool):
         logger.info(f"Generating answer for '{question}'.")
 
         if f"{self.TOOL_FN_NAME}_initialized" in self.settings.callbacks:
-            callback_list = self.settings.callbacks[f"{self.TOOL_FN_NAME}_initialized"]
-            for callback in callback_list:
-                await callback(state)
+            await asyncio.gather(
+                *(
+                    callback(state)
+                    for callback in self.settings.callbacks[
+                        f"{self.TOOL_FN_NAME}_initialized"
+                    ]
+                )
+            )
 
         # TODO: Should we allow the agent to change the question?
         # self.answer.question = query
@@ -289,9 +305,14 @@ class GenerateAnswer(NamedTool):
         logger.info(status)
 
         if f"{self.TOOL_FN_NAME}_completed" in self.settings.callbacks:
-            callback_list = self.settings.callbacks[f"{self.TOOL_FN_NAME}_completed"]
-            for callback in callback_list:
-                await callback(state)
+            await asyncio.gather(
+                *(
+                    callback(state)
+                    for callback in self.settings.callbacks[
+                        f"{self.TOOL_FN_NAME}_completed"
+                    ]
+                )
+            )
 
         return f"{answer} | {status}"
 
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -222,11 +222,7 @@ class GatherEvidence(NamedTool):
         # named this way to support lifecycle callbacks in the future
         # gather_evidence_init, generate_evidence_progress, etc.
         if "gather_evidence_completed" in self.settings.callbacks:
-            callback = self.settings.callbacks["gather_evidence_completed"]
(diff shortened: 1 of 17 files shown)


paperqa/settings.py
```diff
@@ -572,6 +573,26 @@ class Settings(BaseSettings):
         frozen=True,
     )
 
+    # imported here to avoid circular ref
+    if TYPE_CHECKING:
+        from .agents.env import EnvironmentState
+    callbacks: Mapping[str, Callable[["EnvironmentState"], Any]] = Field(
```
jamesbraza, line 579: Pydantic `Field` descriptors only make sense within a `BaseModel`. Let's not use that here

nadolskit: This is part of `Settings` is a class of `BaseSettings` which is a class of `BaseModel`
It makes sense here, no?

nadolskit: This is within `Settings`, a class within `BaseSettings`, a class within `BaseModel`.
I think it makes sense here, no?

paperqa/settings.py
```diff
@@ -572,6 +573,26 @@ class Settings(BaseSettings):
         frozen=True,
     )
 
+    # imported here to avoid circular ref
+    if TYPE_CHECKING:
+        from .agents.env import EnvironmentState
```
jamesbraza, line 578: Can we move this `TYPE_CHECKING` to the module level, not within the class?

paperqa/settings.py
jamesbraza: I got thrown off by the `TYPE_CHECKING` not being at module level, resolving this one

paperqa/settings.py
nadolskit: yup, my bad

paperqa/agents/tools.py
```diff
@@ -218,6 +218,16 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
             if sorted_contexts
             else ""
         )
+
+        # named this way to support lifecycle callbacks in the future
+        # gather_evidence_init, generate_evidence_progress, etc.
+        if "gather_evidence_completed" in self.settings.callbacks:
```
mskarlin, line 224: we could probably use the `TOOL_FN_NAME` here to keep it a little dryer

paperqa/settings.py
```diff
@@ -580,6 +584,23 @@ class Settings(BaseSettings):
         frozen=True,
     )
 
+    callbacks: Mapping[str, Callable[["EnvironmentState"], Any]] = Field(
```
jamesbraza, line 587: Let's put `exclude=True` on this since it's not serializable

tests/test_agents.py
```diff
@@ -421,10 +421,11 @@ async def test_gather_evidence_rejects_empty_docs(
     ), "Agent should have hit its max timesteps"
 
 
+@pytest.mark.parametrize("callback_type", ["none", "async"])
```
jamesbraza, line 424: Let's use `None` over `"none"`, more clear imo

paperqa/agents/tools.py
```diff
@@ -190,19 +190,29 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
...
+            aget_evidence_callbacks = self.settings.callbacks.get(
+                f"{self.TOOL_FN_NAME}_aget_evidence", None
+            )
             # TODO: refactor answer out of this...
             state.answer = await state.docs.aget_evidence(
                 query=state.answer,
                 settings=self.settings,
                 embedding_model=self.embedding_model,
                 summary_llm_model=self.summary_llm_model,
+                callbacks=aget_evidence_callbacks,
```
mskarlin, line 215: style nitpick, but you can probably just insert the expression directly rather than making a variable via `aget_evidence_callbacks`. Also the get("my key", None) is the default, so we usually don't specify it.

nadolskit: ah, good points! will do!

paperqa/agents/tools.py
```diff
@@ -190,19 +190,28 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
         if not state.docs.docs:
             raise EmptyDocsError("Not gathering evidence due to having no papers.")
 
+        if f"{self.TOOL_FN_NAME}_initialized" in self.settings.callbacks:
+            callback_list = self.settings.callbacks[f"{self.TOOL_FN_NAME}_initialized"]
+            for callback in callback_list:
+                await callback(state)
```
jamesbraza, line 196:
```suggestion
            await asyncio.gather(*(c(state) for c in self.settings.callbacks[f"{self.TOOL_FN_NAME}_initialized"])
```

Let's parallelize the callback calling

nadolskit: oh neat! Similar to `Promise.all`
thx!!

paperqa/agents/tools.py
```diff
@@ -218,6 +225,12 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
             if sorted_contexts
             else ""
         )
+
+        if f"{self.TOOL_FN_NAME}_completed" in self.settings.callbacks:
+            callback_list = self.settings.callbacks[f"{self.TOOL_FN_NAME}_completed"]
+            for callback in callback_list:
+                await callback(state)
```
jamesbraza, line 232: Let parallelize all of the callbacks

paperqa/agents/tools.py
jamesbraza: Yeah it's a nice one. Cheers!

paperqa/settings.py
```diff
@@ -580,6 +584,35 @@ class Settings(BaseSettings):
         frozen=True,
     )
 
+    callbacks: Mapping[str, list[Callable[["EnvironmentState"], Any]]] = Field(
```
mskarlin, line 587: @nadolskit sorry I missed this before, but could we move this into "AgentSettings" since these only correspond to agent tool calls?

nadolskit: done!

tests/test_agents.py
```diff
@@ -421,10 +421,11 @@ async def test_gather_evidence_rejects_empty_docs(
     ), "Agent should have hit its max timesteps"
 
 
+@pytest.mark.parametrize("callback_type", [None, "async"])
 @pytest.mark.flaky(reruns=3, only_rerun=["AssertionError", "EmptyDocsError"])
 @pytest.mark.asyncio
 async def test_agent_sharing_state(
-    agent_test_settings: Settings, subtests: SubTests
+    agent_test_settings: Settings, subtests: SubTests, callback_type: str
```
jamesbraza, line 428: This should be `callback_type: str | None`

tests/test_agents.py
```diff
@@ -433,6 +434,22 @@ async def test_agent_sharing_state(
...
     embedding_model = agent_test_settings.get_embedding_model()
 
+    callbacks = {}
+    if callback_type == "async":
+        gen_answer_initialized_callback = AsyncMock()
+        gen_answer_completed_callback = AsyncMock()
+        gather_evidence_initialized_callback = AsyncMock()
+        gather_evidence_completed_callback = AsyncMock()
+
+        callbacks = {
```
jamesbraza, line 444: Not sure it really matters, but it's more future proof to use `callbacks.update` here instead of redefining

PR #592, opened 2024-10-15
Limiting parsed page size to 1.28 million chars

jamesbraza:
We encountered a PDF (for DOI `10.1038/s41593-019-0491-3`) with `PyMuPDF==1.24.11` where one of the pages was parsed to about 400 million characters, of junk (mainly whitespace). This didn't actually crash us, but takes hours to index.

As this was junk, this PR moves us to discard PDF that we fail to parse to be a reasonable amount of text (1.28 million chars), which is already 10X greater than a 128k token context window (and let's ignore chars != tokens in this 10X)

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -269,6 +269,7 @@ class Docs(BaseModel):
                 Doc(docname="", citation="", dockey=dockey),  # Fake doc
                 chunk_chars=parse_config.chunk_size,
                 overlap=parse_config.overlap,
+                page_limit=parse_config.page_limit,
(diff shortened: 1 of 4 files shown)


whitead:
Good idea! Why did you call it page limit but compare it with number of characters?

jamesbraza:
> Good idea! Why did you call it page limit but compare it with number of characters?

We talked in person, it was a bad name. I renamed to `page_size_limit`

mskarlin (approved):
lgtm!!!

PR #593, opened 2024-10-16
Stripping DOI URL prefix for `sources`

jamesbraza:
This enables us to directly compare on DOI downstream

### DIFF
--- a/paperqa/agents/task.py
+++ b/paperqa/agents/task.py
@@ -12,7 +12,7 @@ import re
 from abc import ABC
 from collections.abc import Awaitable, Callable, Sequence
 from enum import StrEnum
-from typing import TYPE_CHECKING, ClassVar, assert_never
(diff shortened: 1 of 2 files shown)


mskarlin (approved):

mskarlin:
fwiw I think DOIs always via the following URLS (from our `DocDetails` class): 

DOI_URL_FORMATS: ClassVar[Collection[str]] = {
        "https://doi.org/",
        "http://dx.doi.org/",
    }

jamesbraza:
> fwiw I think DOIs always via the following URLS (from our `DocDetails` class):
> 
> DOI_URL_FORMATS: ClassVar[Collection[str]] = { "[doi.org](https://doi.org/)", "[http://dx.doi.org](http://dx.doi.org/)", }

Didn't know about that, just moved to use it 👌

PR #594, opened 2024-10-16
Fix sequential client query clobbering

mskarlin:
DocDetails objects that were requested with sequential client queries (rather than simultaneous) were clobbering each other. Our only test worked to check that they were all being called, but not if they overwrote each other. 

Now this works as intended with a test.


jamesbraza (approved):
Nice catch

PR #597, opened 2024-10-16
Cleaned and clarified deferred embeddings code

jamesbraza:
https://github.com/Future-House/paper-qa/pull/289 changed the logic of `add_texts` such that the retrying logic of OpenAI rate limits is no longer relevant. Other comments were stale too, so this PR just cleans them up

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -472,11 +472,11 @@ class Docs(BaseModel):
     async def _build_texts_index(self, embedding_model: EmbeddingModel) -> None:
         texts = [t for t in self.texts if t not in self.texts_index]
         # For any embeddings we are supposed to lazily embed, embed them now
-        to_embed = [t.text for t in texts if t.embedding is None]
+        to_embed = [t for t in texts if t.embedding is None]
         if to_embed:
             for t, t_embedding in zip(
-                texts,
-                await embedding_model.embed_documents(texts=to_embed),
+                to_embed,
+                await embedding_model.embed_documents(texts=[t.text for t in to_embed]),
                 strict=True,
             ):
                 t.embedding = t_embedding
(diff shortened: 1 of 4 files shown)


paperqa/docs.py
```diff
@@ -475,11 +471,15 @@ def delete(
...
         texts = [t for t in self.texts if t not in self.texts_index]
+        # For any embeddings we are supposed to lazily embed, embed them now
         to_embed = [t.text for t in texts if t.embedding is None]
         if to_embed:
-            embeddings = await embedding_model.embed_documents(texts=to_embed)
-            for t, e in zip(texts, embeddings, strict=True):
-                t.embedding = e
+            for t, t_embedding in zip(
+                texts,
+                await embedding_model.embed_documents(texts=to_embed),
```
whitead, line 479: to_embed and texts will not be same shape. Think this is a bug.

jamesbraza: Oh nice catch

So you think it should have `zip(to_embed, ..., strict=True)` instead of `zip(texts, ..., strict=True)`?

I will also look into if this code is covered by our tests

whitead: Only want to embed the to_embed list here.

jamesbraza: Okay just pushed the fix, thanks for confirming

whitead (approved):
should probably add a better unit test to call query, then add a few more, then call query again to get coverage of mixed `to_embed != self.texts` But can wait until future for that

jamesbraza:
> should probably add a better unit test to call query, then add a few more, then call query again to get coverage of mixed `to_embed != self.texts` But can wait until future for that

@whitead this was actually somewhat hard to test (which is a good sign) because our API has so many checks in place for this. That being said, I have written a test `test_partly_embedded_texts` that fails on `main` and passes here. Cheers 🥳

PR #606, opened 2024-10-18
Increased testing of `NumpyVectorStore` and `defer_embeddings`

jamesbraza:
Adding more regression protection before I push some new features today


whitead (approved):
nice

Issue #484, opened 2024-09-26
Bug: supporting `~` in `--paper_directory` in CLI


jamesbraza:
We should allow for `pqa --paper_directory=~/foo`

PR #610, opened 2024-10-18
Testing we can use `~` as `paper_directory`

jamesbraza:
Testing we can:

1. Save `Settings` with `~` in the `paper_directory`
2. Run inferences with `Settings` containing `~` in the `paper_directory`

Closes https://github.com/Future-House/paper-qa/issues/484


whitead (approved):
cool

PR #612, opened 2024-10-18
Exposed `LDPRolloutCallback` and `on_agent_action_callback` for `fake` agent

jamesbraza:
- Exposed `LDPRolloutCallback` so clients can subclass and add state
- Exposed `on_agent_action_callback` for the `fake` agent, to have symmetry with `run_aviary_agent`


mskarlin (approved):
❤️

PR #613, opened 2024-10-19
Fixed `NumpyVectorStore.__eq__`'s `NotImplemented` case

jamesbraza:
I got burned by treating https://github.com/Future-House/paper-qa/pull/608 like a `git rebase` checkpoint instead of including tests of `NumpyVectorStore.__eq__`

### DIFF
--- a/paperqa/llms.py
+++ b/paperqa/llms.py
@@ -845,13 +845,21 @@ class NumpyVectorStore(VectorStore):
     _embeddings_matrix: np.ndarray | None = None
 
     def __eq__(self, other) -> bool:
-        if isinstance(other, type(self)):
-            raise NotImplementedError
+        if not isinstance(other, type(self)):
+            return NotImplemented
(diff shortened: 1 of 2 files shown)


paperqa/llms.py
```diff
@@ -845,8 +845,8 @@ class NumpyVectorStore(VectorStore):
     _embeddings_matrix: np.ndarray | None = None
 
     def __eq__(self, other) -> bool:
-        if isinstance(other, type(self)):
```
jamesbraza, line 848: I didn't pay close attention here, tsk tsk

mskarlin (approved):
Nice fix — fwiw the equality operation could be kind of expensive for a large matrix. If that becomes an issue you could keep a checksum after instantiation and use that for comparison.

jamesbraza:
Thanks yeah I agree, good to keep in mind

PR #614, opened 2024-10-19
Implemented `__deepcopy__` on all `Environment`s

jamesbraza:
- Implemented `__deepcopy__` on `PaperQAEnvironment` and `GradablePaperQAEnvironment`, with a test
    - This needed a `make_tools`, so I decomposed `PaperQAEnvironment.make_initial_state_and_tools` for it. Note this is a breaking change
- Documented `LitQAEvaluation.from_question`, as its inner workings were requiring too much thought
- Fixed another `model_config` not being at the top I missed in https://github.com/Future-House/paper-qa/pull/607

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -1,5 +1,6 @@
 import logging
-from typing import cast
+from copy import deepcopy
+from typing import Any, Self, cast
 
 from aviary.env import Environment, Frame
 from aviary.message import Message
@@ -172,3 +173,26 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
             ),
             False,
         )
+
+    def __deepcopy__(self, memo) -> Self:
+        copy_state = deepcopy(self.state, memo)
+        # We don't know the side effects of deep copying a litellm.Router,
+        # so we force a shallow copy of these LiteLLMModels
+        env_model_kwargs: dict[str, Any] = {
+            name: model if model is None else type(model)(**model.model_dump())
+            for name, model in (
+                ("llm_model", self._llm_model),
+                ("summary_llm_model", self._summary_llm_model),
+                ("embedding_model", self._embedding_model),
+            )
+        }
+        copy_self = type(self)(
+            query=deepcopy(self._query, memo),  # deepcopy for _docs_name
+            docs=copy_state.docs,
+            **env_model_kwargs,
+        )
+        copy_self.state = copy_state
+        # Because we shallow copied the LiteLLMModels, we need to re-make the
+        # tool functions within the tools
+        copy_self.tools = copy_self.make_tools()
+        return copy_self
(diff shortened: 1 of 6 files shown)


paperqa/agents/env.py
```diff
@@ -111,29 +112,30 @@ def __init__(
         self._summary_llm_model = summary_llm_model
         self._embedding_model = embedding_model
 
-    def make_initial_state_and_tools(self) -> tuple[EnvironmentState, list[Tool]]:
```
jamesbraza, line 114: Note: this does break the interface, but I don't think it will affect anyone, it's practically a private method

tests/test_agents.py
```diff
@@ -709,3 +711,47 @@ def test_agent_prompt_collection_validations(
     else:
         with pytest.raises(result):
             AgentSettings(**kwargs)
+
+
+@pytest.mark.asyncio
+async def test_deepcopy_env(agent_test_settings: Settings) -> None:
```
jamesbraza, line 717: This unit test should be considered really baller

tests/test_agents.py
```diff
@@ -709,3 +711,47 @@ def test_agent_prompt_collection_validations(
...
+
+    # 3. Generate an answer for both, and confirm they are identical
+    gen_answer_action = ToolRequestMessage(
+        tool_calls=[ToolCall.from_name("gen_answer", question=question)]
+    )
+    _, _, done, _ = await env.step(gen_answer_action)
+    assert done
+    assert not env.state.answer.could_not_answer
+    assert env.state.answer.used_contexts
+    _, _, done, _ = await env_copy.step(gen_answer_action)
```
mskarlin, line 751: Is there a chance this could be flaky given it’s inference based?

paperqa/agents/env.py
mskarlin: Oddly enough this was the method I had to overwrite in the server but it’s nbd to swap to this interface

paperqa/agents/env.py
```diff
@@ -171,3 +173,26 @@ async def step(
             ),
             False,
         )
+
+    def __deepcopy__(self, memo) -> Self:
+        copy_state = deepcopy(self.state, memo)
+        # We don't know the side effects of deep copying a litellm.Router,
```
mskarlin, line 179: Smart

mskarlin (approved):
Looks good to me!

tests/test_agents.py
jamesbraza: Yeah we definitely can be flaky, I added `pytest.mark.flaky` just now

paperqa/agents/env.py
jamesbraza: Well, let's hope no one else did 😆

PR #615, opened 2024-10-21
Made embedding default

whitead:
mypy seems to be not liking this missing kwarg. It is present on all other fields, so just making them consistent.


mskarlin (approved):
Nice catch!

jamesbraza (approved):

PR #618, opened 2024-10-21
Pulled in latest PyMuPDF for `set_messages`

jamesbraza:
Pulling in the resolution of https://github.com/pymupdf/PyMuPDF/issues/3914

Also, https://github.com/Future-House/paper-qa/pull/538 introduced a control for the PyMuPDF version for test cassettes. I found the cassette made with `pymupdf==1.24.11` still works on `pymupdf==1.24.12` (pulled in by https://github.com/Future-House/paper-qa/pull/617 ), so this PR also drops the cassette naming special casing.


whitead (approved):
nice

PR #620, opened 2024-10-21
Added configuration to adjust how contexts are displayed

whitead:
Added two new prompt configuration variables to better control the formatting of the context string.

Fixed non-executing prompt validation fields. Led to eliminating a broken test.

Made it so contexts are sorted by score (reverse) and alphabetical


paperqa/settings.py
```diff
@@ -20,6 +20,7 @@
     model_validator,
 )
 from pydantic_settings import BaseSettings, CliSettingsSource, SettingsConfigDict
+from typing_extensions import deprecated
```
jamesbraza, line 23: We don't have `typing_extensions` here, let's not use this

Just use the built-in `DeprecationWarning` as we do elsewhere in the code base. Imo it's not worth adding another dep for this

tests/test_paperqa.py
```diff
@@ -1083,3 +1083,49 @@ def test_external_doc_index(stub_data_dir: Path) -> None:
...
+    docs = Docs()
+    docs.add(stub_data_dir / "bates.txt", "WikiMedia Foundation, 2023, Accessed now")
+    response = docs.query("What country is Bates from?", settings=settings)
+    assert "@@@@@" in response.context
+    assert "WikiMedia Foundation, 2023" in response.context
+
+
+def test_evidence_detailed_citations_shim(stub_data_dir: Path) -> None:
+
+    # TODO: delete in v6
```
jamesbraza, line 1108: Delete what, the entire test, or just a specific part of this test? Can you just update the comment to state one or the other

tests/test_paperqa.py
```diff
@@ -1083,3 +1083,49 @@ def test_external_doc_index(stub_data_dir: Path) -> None:
...
+    settings = Settings.from_name("fast")
+    settings.answer.evidence_detailed_citations = False
+    docs = Docs()
+    docs.add(stub_data_dir / "bates.txt", "WikiMedia Foundation, 2023, Accessed now")
+    response = docs.query("What country is Bates from?", settings=settings)
+    assert "WikiMedia Foundation, 2023, Accessed now" not in response.context
+
+
+def test_context_outer_prompt(stub_data_dir: Path) -> None:
```
jamesbraza, line 1117: Is there some way we can combine the `inner_prompt` and `outer_prompt` tests? Mainly to avoid so extra `docs.add` in our test suite

And feel free to ignore this one, it's unimportant overall

paperqa/settings.py
whitead: Ok - followed examples from below

tests/test_paperqa.py
whitead: Added comment - delete whole test

tests/test_paperqa.py
whitead: Combined the two

jamesbraza (approved):
Nice work

PR #627, opened 2024-10-22
Caching opened `tantivy.Index`es in the package

jamesbraza:
### Motivation

https://github.com/quickwit-oss/tantivy-py/issues/359#issuecomment-2428087169 reveals that one of our servers can only handle three opened indexes at once. The reason why remains unclear, but this PR at least reacts to the issue by caching opened indexes in the `search.py` module's scope. Now we can invoke `await SearchIndex(index_name="normal-index").index` as many times as desired within one Python process.

### Implementation Details

Why a global scope? We want to accommodate caching the `Index` across:
- Across `deepcopy` of an `PaperQAEnvironment`, whose `tools` which contains a `PaperSearch` tool instance.
    - So we can't cache the `Index` in `PaperSearch`, or else we'd be making `Index` copies (and the side effects of this are unknown)
- Across a `Trajectory`, where we (1) build the index and (2) use the index in 0+ paper searches
- Across a `TaskDataset`, where we (1) build the index and (2) run 0+ envs for one trajectory each
- Across many `TaskDataset`, for larger experiments

This can only be accomplished using global scope, whose lifetime matches the entire Python process. This unfortunately requires callers to invoke the newly added `reap_opened_index_cache` at runtime if intermediary cleaning of the cache is desired.

The caching added here can be disabled by setting `1` or `true` (case insensitive) to the newly-added environment variable `PQA_INDEX_DONT_CACHE_INDEXES`.

### Risks

- Race conditions if invoking `reap_opened_index_cache` while also using `PaperSearch`
    - Our test suite intentionally doesn't use `reap_opened_index_cache` to avoid this, but the tradeoff is our testing slowly accrues indexes across cases
- Since caching is enabled by default, if the clients aren't aware of this, it can be considered unexpected statefulness

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -11,7 +11,7 @@ import warnings
 import zlib
 from collections.abc import Awaitable, Callable, Collection, Sequence
 from enum import StrEnum, auto
-from typing import TYPE_CHECKING, Any, ClassVar, cast
+from typing import TYPE_CHECKING, Any, ClassVar
 from uuid import UUID
 
 import anyio
@@ -127,9 +127,9 @@ class SearchIndex:
             )
         self.index_name = index_name
         self._index_directory = index_directory
-        self._schema = None
-        self._index = None
-        self._searcher = None
+        self._schema: Schema | None = None
+        self._index: Index | None = None
+        self._searcher: Searcher | None = None
         self._index_files: dict[str, str] = {}
         self.changed = False
         self.storage = storage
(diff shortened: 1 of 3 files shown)


whitead (approved):
Great battle - glad it's solved

paperqa/agents/search.py
```diff
@@ -165,27 +188,45 @@ def schema(self) -> Schema:
...
+            self._schema = schema_builder.build()
+        return self._schema
 
     @property
     async def index(self) -> Index:
         if not self._index:
-            index_path = await self.index_filename
-            if await (index_path / "meta.json").exists():
-                self._index = Index.open(path=str(index_path))  # type: ignore[assignment]
+            index_meta_directory = await self.index_filename
```
mskarlin, line 197: I think we can hit a race condition here (because the exists check is awaitable). We should prob. acquire a lock and then do the check or do the check synchronously. 

Otherwise I think a "gather" call could result in several `False` responses to this `exists` call.

mskarlin: Ok we talked about this, because we are only instantiating and Index and not opening one, this likely isn't an issue. @jamesbraza is going to add a comment for future reference.

jamesbraza: Yep thanks for the discussion! I added a comment to the code documenting some of our talking points

mskarlin (approved):

PR #629, opened 2024-10-23
Printing the `text` in a failed `llm_parse_json`

jamesbraza:
Making it possible to debug `llm_parse_json` failures


whitead (approved):
I actually do not like this function. Think it's too restrictive to blow up on parse error, but obv unrelated to this PR

PR #632, opened 2024-10-23
Change S2 client logic to use arxiv doi if it's defined

mskarlin:
We already had special logic handling this, we just didn't pass doi to the final DocDetails object being created. Added a test too.


tests/test_clients.py
```diff
@@ -632,3 +632,19 @@ async def test_crossref_retraction_status(stub_data_dir: Path) -> None:
 def test_reformat_name(name: str, expected: str) -> None:
     result = reformat_name(name)
     assert result == expected, f"Expected '{expected}', but got '{result}' for '{name}'"
+
+
+@pytest.mark.vcr
+@pytest.mark.asyncio
+async def test_arxiv_dois() -> None:
```
jamesbraza, line 639: Can you make this test name a bit more verbose/self-explanatory?

Something like `test_arxiv_doi_is_used_when_available`

tests/test_clients.py
```diff
@@ -632,3 +632,19 @@ async def test_crossref_retraction_status(stub_data_dir: Path) -> None:
 def test_reformat_name(name: str, expected: str) -> None:
     result = reformat_name(name)
     assert result == expected, f"Expected '{expected}', but got '{result}' for '{name}'"
+
+
+@pytest.mark.vcr
+@pytest.mark.asyncio
+async def test_arxiv_doi_is_used_when_available() -> None:
```
jamesbraza, line 639: Any chance we can merge this test with another test, with explanatory comment or docs? Mainly, we are getting a lot of tests that do mostly the same thing

Also, feel free to ignore this

mskarlin: nah -- I need it to be a different subset of scrapers (i.e. only Crossref then s2), and I don't want to reconfigure all of our other tests to take custom subsets of scrapers atm.

PR #640, opened 2024-10-25
Make string similarity case insensitive by default

mskarlin:
We were losing Crossref title-matched articles because they return titles with different casings than we expected, this makes us safe against that.


whitead (approved):
Should we just always be case insensitive? Can't think of why not.

PR #649, opened 2024-10-28
Add an after model validator ensuring temp=1 for o1 models

dakoner:
o1-preview and o1-mini in Beta only support temperature=1.

### DIFF
--- a/paperqa/settings.py
+++ b/paperqa/settings.py
@@ -678,6 +678,19 @@ class Settings(BaseSettings):
                 setattr(self.agent.index, new_name, value)  # Propagate to new location
         return self
 
+    @model_validator(mode="after")
+    def _validate_temperature_for_o1_preview(self) -> Self:
+        """Ensures temperature is 1 if the LLM is 'o1-preview' or 'o1-mini'.
+
+        o1 reasoning models only support temperature = 1.  See 
+        https://platform.openai.com/docs/guides/reasoning/quickstart
+        """
+        if self.llm in ("o1-preview", "o1-mini") and self.temperature != 1:
+                raise ValueError(
+                    "When using 'o1-preview' LLM, the temperature must be set to 1."
+                )
+        return self
+
     @computed_field  # type: ignore[prop-decorator]
     @property
     def md5(self) -> str:
--- a/paperqa/settings.py
+++ b/paperqa/settings.py
@@ -682,13 +682,13 @@ class Settings(BaseSettings):
     def _validate_temperature_for_o1_preview(self) -> Self:
         """Ensures temperature is 1 if the LLM is 'o1-preview' or 'o1-mini'.
 
-        o1 reasoning models only support temperature = 1.  See 
+        o1 reasoning models only support temperature = 1.  See
         https://platform.openai.com/docs/guides/reasoning/quickstart
         """
-        if self.llm in ("o1-preview", "o1-mini") and self.temperature != 1:
-                raise ValueError(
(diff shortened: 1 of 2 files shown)


paperqa/settings.py
```diff
@@ -678,6 +678,19 @@ def _deprecated_field(self) -> Self:
...
+    @model_validator(mode="after")
+    def _validate_temperature_for_o1_preview(self) -> Self:
+        """Ensures temperature is 1 if the LLM is 'o1-preview' or 'o1-mini'.
+
+        o1 reasoning models only support temperature = 1.  See 
+        https://platform.openai.com/docs/guides/reasoning/quickstart
+        """
+        if self.llm in ("o1-preview", "o1-mini") and self.temperature != 1:
+                raise ValueError(
+                    "When using 'o1-preview' LLM, the temperature must be set to 1."
```
jamesbraza, line 690: Can you put the specified temperature into the value error?  Something like `f"the temperature must be set to 1, {self.temperature} was specified."`

paperqa/settings.py
```diff
@@ -678,6 +678,19 @@ def _deprecated_field(self) -> Self:
...
         return self
 
+    @model_validator(mode="after")
+    def _validate_temperature_for_o1_preview(self) -> Self:
+        """Ensures temperature is 1 if the LLM is 'o1-preview' or 'o1-mini'.
+
+        o1 reasoning models only support temperature = 1.  See 
+        https://platform.openai.com/docs/guides/reasoning/quickstart
+        """
+        if self.llm in ("o1-preview", "o1-mini") and self.temperature != 1:
```
jamesbraza, line 688: Should we allow `None` temperature to be autocast to 1?

paperqa/settings.py
```diff
@@ -678,6 +678,19 @@ def _deprecated_field(self) -> Self:
                 setattr(self.agent.index, new_name, value)  # Propagate to new location
         return self
 
+    @model_validator(mode="after")
+    def _validate_temperature_for_o1_preview(self) -> Self:
```
jamesbraza, line 682: Can you modify one of our unit tests to use `o1` so this gets covered?

paperqa/settings.py
dakoner: Discussed in person- decided not to.

paperqa/settings.py
```diff
@@ -678,6 +678,19 @@ def _deprecated_field(self) -> Self:
...
+        """Ensures temperature is 1 if the LLM is 'o1-preview' or 'o1-mini'.
+
+        o1 reasoning models only support temperature = 1.  See
+        https://platform.openai.com/docs/guides/reasoning/quickstart
+        """
+        if self.llm.startswith("o1-") and self.temperature != 1:
+            raise ValueError(
+                f"the temperature must be set to 1, {self.temperature} was specified."
```
jamesbraza, line 690:
```suggestion
                "When dealing with OpenAI o1 models,  the temperature"
                f" must be set to 1, {self.temperature} was specified."
```

Sorry to double post here, but do you mind specifying "o1" somewhere in this message? It will make the failure message more useful

paperqa/settings.py
dakoner: Sure, should I put it in test_llms (already parameterized for gpt-4), test_configs (seems to be the central place for config testing), or some other file?

jamesbraza: Oh doesn't matter to me, I guess just pick one that seems to not rely on whatever model it was using previously. Defer to you here

jamesbraza (approved):

PR #651, opened 2024-10-29
Fixing flaky test `test_minimal_fields_filtering`

jamesbraza:
- Adds `record_mode` of once to VCR config (since it's a better default)
- Resolves type ignores that can be fixed with more assertions


dakoner (approved):
LGTM

PR #653, opened 2024-10-30
Renamed to PQASession type

whitead:
This renames the `Answer` object to be a `PQASession` object. This reflects the fact that the object often doesn't contain an answer and contains much more than that.

I've aliased it in the file, so that the change should not affect users.

### DIFF
--- a/paperqa/agents/models.py
+++ b/paperqa/agents/models.py
@@ -85,7 +85,7 @@ class QueryRequest(BaseModel):
 
 
 class AnswerResponse(BaseModel):
-    answer: PQASession
+    session: PQASession = Field(alias="answer")
     bibtex: dict[str, str] | None = None
     status: AgentStatus
     timing_info: dict[str, dict[str, float]] | None = None
@@ -94,7 +94,7 @@ class AnswerResponse(BaseModel):
     # about the answer, such as the number of sources used, etc.
     stats: dict[str, str] | None = None
 
-    @field_validator("answer")
+    @field_validator("session")
     def strip_answer(
         cls, v: PQASession, info: ValidationInfo  # noqa: ARG002, N805
     ) -> PQASession:
@@ -114,7 +114,7 @@ class AnswerResponse(BaseModel):
         )
         result = await model.run_prompt(
             prompt="{question}\n\n{answer}",
-            data={"question": self.answer.question, "answer": self.answer.answer},
+            data={"question": self.session.question, "answer": self.session.answer},
             system_prompt=sys_prompt,
         )
         return result.text.strip()
--- a/paperqa/agents/models.py
+++ b/paperqa/agents/models.py
@@ -85,6 +85,8 @@ class QueryRequest(BaseModel):
 
 
 class AnswerResponse(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
     session: PQASession = Field(alias="answer")
     bibtex: dict[str, str] | None = None
     status: AgentStatus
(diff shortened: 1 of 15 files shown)


paperqa/agents/models.py
```diff
@@ -85,7 +85,7 @@ def set_docs_name(self, docs_name: str) -> None:
 
 
 class AnswerResponse(BaseModel):
-    answer: Answer
+    answer: PQASession
```
jamesbraza, line 88: Should we rename this to `session` with an `alias` to `answer`?

paperqa/settings.py
```diff
@@ -302,9 +302,9 @@ def check_select(cls, v: str) -> str:
     def check_post(cls, v: str | None) -> str | None:
         if v is not None:
             # kind of a hack to get list of attributes in answer
-            from paperqa.types import Answer
+            from paperqa.types import PQASession
```
jamesbraza, line 305: Can we move this to the module level? Or will it cause a circular import

paperqa/types.py
```diff
@@ -248,6 +248,10 @@ def could_not_answer(self) -> bool:
         return "cannot answer" in self.answer.lower()
 
 
+# for backwards compatibility
+Answer = PQASession
```
jamesbraza, line 252: I am thinking you don't feel like doing this, but should we start a deprecation cycle for `Answer`?

tests/conftest.py
```diff
@@ -92,8 +92,8 @@ def agent_test_settings(agent_index_dir: Path, stub_data_dir: Path) -> Settings:
 
 
 @pytest.fixture
-def agent_stub_answer() -> Answer:
-    return Answer(question="What is is a self-explanatory model?")
+def agent_stub_answer() -> PQASession:
```
jamesbraza, line 95: Can you update this fixture name to `stub_session`?

paperqa/settings.py
whitead: yea - circular import is why this is here

paperqa/agents/models.py
whitead: Thanks - missed this one

mskarlin (approved):

paperqa/types.py
```diff
@@ -40,38 +41,39 @@
...
     """
 
     id: UUID = Field(default_factory=uuid4)
-    answer_id: UUID | None = Field(
-        default_factory=cvar_answer_id.get,
+    session_id: UUID | None = Field(
+        default_factory=cvar_session_id.get,
         description="A persistent ID to associate a group of LLMResults",
+        alias="answer_id",
     )
```
whitead, line 77: @mskarlin not sure how much this will change things on server side

jamesbraza: You'll need to opt into Pydantic `ModelConfig(populate_by_name=True)` for `alias` to work for assignment at construction time

mskarlin: I think since we overwrite this w. the query id we should be OK

whitead: @jamesbraza - do I need to do that? I think users can only view them, users should not be constructing them, and I changed the init code.

jamesbraza: Oh okay sounds like you're aware of this, then feel free to ignore it

whitead: @jamesbraza - on reflection, I do need to do this though, otherwise we cannot load objects from DB that use `answer` as the field name

paperqa/types.py
```diff
@@ -248,6 +252,17 @@ def could_not_answer(self) -> bool:
         return "cannot answer" in self.answer.lower()
 
 
+# for backwards compatibility
+class Answer(PQASession):
```
jamesbraza, line 256: Nice work!

jamesbraza (approved):

PR #661, opened 2024-11-03
[WIP] version 6 release

whitead:
- [x] Generalizes context to be multipart messages
- [x] Allows multimodal contexts
- [x] Enables caching control on messages
- [x] Adjusts flow to allow context to be either packed into system prompt or some separate part of query, so that it can be cached
- [ ] Update typing/update litellm types
- [ ] Create image context that encodes image to a URL
- [ ] Add a PDF reader that uses images instead of text 
- [x] Deprecate detailed evidence


whitead:
Moved to #704

PR #665, opened 2024-11-04
Add test for difficult-to-reconcile dois

mskarlin:
Tests that we're able to get back metadata for a tougher scenario.

### DIFF
--- a/tests/test_clients.py
+++ b/tests/test_clients.py
@@ -661,3 +661,36 @@ async def test_arxiv_doi_is_used_when_available() -> None:
     )
     assert result, "paper should be found"
     assert result.doi == "10.48550/arxiv.1706.03762"
+
+
+@pytest.mark.vcr
+@pytest.mark.asyncio
+async def test_hard_reconciles() -> None:
+    test_parameters: list[dict] = [
+        {
+            "title": (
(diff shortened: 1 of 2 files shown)


tests/test_clients.py
```diff
@@ -661,3 +661,36 @@ async def test_arxiv_doi_is_used_when_available() -> None:
...
+            ),
+            "doi": "10.1101/2024.04.01.587366",
+        },
+        {
+            "title": (
+                "High-throughput screening of human genetic variants by pooled"
+                " prime editing."
+            ),
+            "authors": ["garbage", "authors", "that"],
+            "doi": None,
```
jamesbraza, line 683: I might say we remove this and do `test.get("doi")` below to be more future looking to additional cases

tests/test_clients.py
```diff
@@ -661,3 +661,36 @@ async def test_arxiv_doi_is_used_when_available() -> None:
     )
     assert result, "paper should be found"
     assert result.doi == "10.48550/arxiv.1706.03762"
+
+
+@pytest.mark.vcr
+@pytest.mark.asyncio
+async def test_hard_reconciles() -> None:
```
jamesbraza, line 668: Can you add a docstring or code comment for each case saying what was "hard"?

PR #666, opened 2024-11-05
Convert citation to formatted_citation usage where necessary

mskarlin:
We were using `Doc.citation` in our queries, which skipped the extra info from our journal quality and citation count metadata.


whitead (approved):
Something is missing here - either DocDetails should not allow `None` citation or `Doc` should allow `None` citation. Otherwise we have some weird stuff going on I think

mskarlin:
good call -- defaulting to an empty string, which I think should be fine.

paperqa/types.py
```diff
@@ -467,6 +477,32 @@ def inject_clean_doi_url_into_data(data: dict[str, Any]) -> dict[str, Any]:
 
         return data
 
+    @staticmethod
+    def add_preprint_journal_from_doi_if_missing(
+        data: dict[str, Any],
+    ) -> dict[str, Any]:
+        if not data.get("journal"):
+            doi = data.get("doi", "") or ""
+            if "10.48550/" in doi or "ArXiv" in (data.get("other", {}) or {}).get(
```
jamesbraza, line 486: Can we move this to an external module-level lookup of journal to base DOI?

PR #670, opened 2024-11-07
Made it so you do not die on invalid tool

whitead:
^

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,7 +85,7 @@ repos:
         additional_dependencies:
           - aiohttp
           - coredis
-          - fhaviary[llm]>=0.8.2 # Match pyproject.toml
+          - fhaviary[llm]>=0.10.0 # Match pyproject.toml
           - ldp>=0.12 # Match pyproject.toml
           - html2text
           - httpx
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,17 +84,20 @@ repos:
         args: [--pretty, --ignore-missing-imports]
         additional_dependencies:
           - aiohttp
+          - PyMuPDF>=1.24.12
+          - anyio
           - coredis
           - fhaviary[llm]>=0.10.0 # Match pyproject.toml
           - ldp>=0.12 # Match pyproject.toml
           - html2text
+          - litellm>=1.44 # Match pyproject.toml
           - httpx
           - limits
+          - pybtex
           - numpy
-          - openai>=1 # Match pyproject.toml
           - pandas-stubs
           - pydantic~=2.0 # Match pyproject.toml
-          - pydantic-settings
+          - pydantic-settings<2.6 # Match pyproject.toml
           - rich
           - tantivy
           - tenacity
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -105,3 +105,4 @@ repos:
           - types-setuptools
           - types-PyYAML
           - sentence-transformers
+          - pyzotero
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -104,3 +104,4 @@ repos:
           - tiktoken>=0.4.0 # Match pyproject.toml
           - types-setuptools
           - types-PyYAML
+          - sentence-transformers
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -180,7 +180,6 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
                 action,
                 state=self.state,
                 handle_tool_exc=True,
-                ordered=True,  # prevent concurrent gather evidence steps or anything
             ),
         )
         return (
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -162,22 +162,14 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
     ) -> tuple[list[Message], float, bool, bool]:
         self.state.session.add_tokens(action)  # Add usage for action if present
 
-        valid_action, invalid_action = self.filter_invalid_tool_calls(action)
-        # NOTE: valid_action or invalid_action may have an empty list of tool calls
         response_messages = cast(
             list[Message],
-            # Ordered since things like search -> lookup need to be run in order.
-            # NOTE: Handling tool exceptions here keeps the trajectory going, but I don't
-            # think the returned message is useful to the agent/learning. Disabling for now.
             await self.exec_tool_calls(
-                valid_action, state=self.state, handle_tool_exc=True
-            )
-            + [
-                ToolResponseMessage.from_call(
-                    tool_call, content=f"Invalid tool call: {tool_call.function.name}"
-                )
-                for tool_call in invalid_action.tool_calls
-            ],
+                action,
+                state=self.state,
+                handle_tool_exc=True,
+                ordered=True,  # prevent concurrent gather evidence steps or anything
+            ),
         )
         return (
             response_messages,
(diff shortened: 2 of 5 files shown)


paperqa/agents/env.py
```diff
@@ -162,19 +162,31 @@ async def step(
...
-        # If the action has empty tool_calls, the agent can later take that into account
-        msgs = cast(
+        valid_action, invalid_action = self.filter_invalid_tool_calls(action)
+        # NOTE: valid_action or invalid_action may have an empty list of tool calls
+        response_messages = cast(
             list[Message],
-            await self.exec_tool_calls(action, state=self.state, handle_tool_exc=True),
+            # Ordered since things like search -> lookup need to be run in order.
+            # NOTE: Handling tool exceptions here keeps the trajectory going, but I don't
+            # think the returned message is useful to the agent/learning. Disabling for now.
```
jamesbraza, line 171:
```suggestion
```

This comment is for HotPotQA:
- Mentions `lookup` tool
- Doesn't have `handle_tool_exc`

Let's just remove it

paperqa/agents/env.py
```diff
@@ -162,19 +162,31 @@ async def step(
...
+            await self.exec_tool_calls(
+                valid_action, state=self.state, handle_tool_exc=True
+            )
+            + [
+                ToolResponseMessage.from_call(
+                    tool_call, content=f"Invalid tool call: {tool_call.function.name}"
```
jamesbraza, line 177:
```suggestion
                    tool_call, content=f"Tool call with invalid name {tool_call.function.name!r}."
```

Let's be more specific why it's invalid

paperqa/agents/env.py
```diff
@@ -162,19 +162,31 @@ async def step(
...
+                )
+                for tool_call in invalid_action.tool_calls
```
jamesbraza, line 179: One comment on this implementation is the order of `tool_calls` input vs output is not preserved here when there's invalid tool calls

paperqa/agents/env.py
```diff
@@ -162,19 +162,31 @@ async def step(
     ) -> tuple[list[Message], float, bool, bool]:
         self.state.session.add_tokens(action)  # Add usage for action if present
 
-        # If the action has empty tool_calls, the agent can later take that into account
-        msgs = cast(
+        valid_action, invalid_action = self.filter_invalid_tool_calls(action)
```
jamesbraza, line 165: Okay so, based on internal discussions today, I think we should check if `not action.tool_calls`, and return `done` if so

```python
if not action.tool_calls:
    return Message(content="Agent chose to finish."), 0, GenerateAnswer.FAILED_TO_ANSWER, False
```

jamesbraza: Let's save this for another PR happening after this one, I can take care of it

paperqa/agents/env.py
jamesbraza: We can save this for another PR too

paperqa/agents/env.py
jamesbraza: Handled this separately in https://github.com/Future-House/paper-qa/pull/671

pyproject.toml
```diff
@@ -42,13 +42,12 @@ dependencies = [
...
     "anyio",
     "coredis",
-    "fhaviary[llm]>=0.8.2",  # For namespace reorg
+    "fhaviary[llm]>=0.10.0",  # For invalid tool fixes
     "html2text",  # TODO: evaluate moving to an opt-in dependency
     "httpx",
     "limits",
     "litellm>=1.44",  # to prevent sys.stdout on router creation
     "numpy",
-    "openai<1.47",  # TODO: remove after https://github.com/BerriAI/litellm/issues/5854 resolution
```
jamesbraza, line 51: @whitead can you remove this from `pre-commit`'s `mypy` as well?

pyproject.toml
```diff
@@ -42,13 +42,12 @@ dependencies = [
     "aiohttp",  # TODO: remove in favor of httpx
     "anyio",
     "coredis",
-    "fhaviary[llm]>=0.8.2",  # For namespace reorg
```
jamesbraza, line 45: Can you match this in `pre-commit`'s `mypy`

paperqa/agents/env.py
```diff
@@ -174,19 +174,23 @@ async def step(
...
-        # If the action has empty tool_calls, the agent can later take that into account
-        msgs = cast(
+        response_messages = cast(
             list[Message],
-            await self.exec_tool_calls(action, state=self.state, handle_tool_exc=True),
+            await self.exec_tool_calls(
+                action,
+                state=self.state,
+                handle_tool_exc=True,
+                ordered=True,  # prevent concurrent gather evidence steps or anything
```
jamesbraza, line 183: Can you do `ordered=True` in another PR?

Alternately, can you put a `TODO` here to remove this?

pyproject.toml
whitead: Isn't in pre-commit

pyproject.toml
whitead: good catch

paperqa/agents/env.py
whitead: Do you think it's not a bug to allow concurrent gather evidence?

whitead: removed

.pre-commit-config.yaml
```diff
@@ -85,7 +85,7 @@ repos:
         additional_dependencies:
           - aiohttp
           - coredis
-          - fhaviary[llm]>=0.8.2 # Match pyproject.toml
+          - fhaviary[llm]>=0.10.0 # Match pyproject.toml
```
jamesbraza, line 88: A few line below, let's remove the:

```toml
- openai>=1 # Match pyproject.toml
```

pyproject.toml
jamesbraza: Are you sure

.pre-commit-config.yaml
whitead: Do you not want to have openai type checking?

jamesbraza: So `mypy` will still get `openai` through `litellm`, as `openai` is a `litellm` dependency. So we'll still have `openai` there.

The idea is we want the `mypy` list to match our first party deps in `dependencies`.

We made `openai` a first party dependency temporarily just to ensure its version matched. So we had to do the same in `mypy`. Now that this PR removes it from our first party deps, we want to remove it from our `mypy` list as well to keep them in sync

whitead: Got it - this section of mypy has many inconsistencies with the pyproject.toml file. Will fix it I guess

whitead: Like `litellm` isn't even in the `mypy` file

jamesbraza: Yes you are dead on 🥳 

One last dimension (we also just talked in person) is that some packages are untyped: https://github.com/Future-House/paper-qa/blob/v5.3.2/pyproject.toml#L144-L149, so we don't need to give them to `mypy`'s isolated list here. Respecifying them (even though they aren't typed):
- Pro: future looking for if they become typed
- Con: bloats `pre-commit`'s isolated venv a bit

That being said, it's an easier heuristic to just "exact match" dependency lists. So let's leave as is, thanks and resolving

jamesbraza (approved):
LGTM, thanks for doing this!

pyproject.toml
```diff
@@ -258,6 +257,8 @@ min-similarity-lines = 12
 # Add the specified OPTS to the set of command line arguments as if they had been
 # specified by the user.
 addopts = "--typeguard-packages=paperqa"
+# set how loops are scoped to avoid https://github.com/BerriAI/litellm/issues/5854
+asyncio_default_fixture_loop_scope = "session"
```
jamesbraza, line 261: Confirmed this worked in https://github.com/Future-House/paper-qa/pull/674

Nice work here @whitead, we are unblocked again

PR #671, opened 2024-11-08
Empty tool calls leading to `done` condition

jamesbraza:
This PR brings `PaperQAEnvironment` to support empty tool calls leading to the environment being done. This creates a back door for the agent to trigger a rollout to conclude.

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -1,6 +1,6 @@
 import logging
 from copy import deepcopy
-from typing import Any, Self, cast
+from typing import Any, ClassVar, Self, cast
 
 from aviary.core import (
     Environment,
@@ -169,10 +169,24 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
             > self._query.settings.answer.max_answer_attempts
         )
 
+    USE_POST_PROCESSED_REWARD: ClassVar[float] = 0.0
+
     async def step(
         self, action: ToolRequestMessage
     ) -> tuple[list[Message], float, bool, bool]:
         self.state.record_action(action)
+        if not action.tool_calls:
+            return (
+                # NOTE: don't put:
+                # - GenerateAnswer.FAILED_TO_ANSWER here because this wasn't a failure
+                # - 'cannot answer' because that information belongs in
+                #   PQASession.answer, not in the message history
+                # Let's just put a nice message about being done :)
+                [Message(content="Agent specified 0 tool calls, which means done.")],
+                self.USE_POST_PROCESSED_REWARD,
+                True,  # Matching LangChain: https://github.com/langchain-ai/langchain/blob/langchain%3D%3D0.2.17/libs/langchain/langchain/agents/output_parsers/openai_functions.py#L38-L77
+                False,  # Let caller determine truncations
+            )
 
         response_messages = cast(
             list[Message],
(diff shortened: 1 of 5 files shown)


paperqa/settings.py
```diff
@@ -448,8 +448,9 @@ class AgentSettings(BaseModel):
     agent_prompt: str = (
         "Use the tools to answer the question: {question}\n\nThe {gen_answer_tool_name}"
         " tool output is visible to the user, so you do not need to restate the answer"
-        " and can simply terminate if the answer looks sufficient. The current status"
-        " of evidence/papers/cost is {status}"
+        " and can simply terminate if the answer looks sufficient. If you do not"
```
mskarlin, line 451: I think this wording will confuse the agent (because it confuses me): "If you do not specify any tool calls, this means you failed to answer the question." -- I'm imagining our intended behavior is that the question seems impossible for the agent, right? Otherwise I think it should always keep trying. 

So maybe something like "If the answer does not look sufficient, and you have already tried to answer several times, you can terminate the question by specifying 0 tool calls."

mskarlin: we'll need to test after this change, because the agent might be tempted to bail out too often.

paperqa/agents/env.py
```diff
@@ -157,10 +157,26 @@ async def reset(self) -> tuple[list[Message], list[Tool]]:
...
         self, action: ToolRequestMessage
     ) -> tuple[list[Message], float, bool, bool]:
         self.state.session.add_tokens(action)  # Add usage for action if present
+        if not action.tool_calls:
+            return (
+                [
+                    Message(
+                        content=(
+                            f"{GenerateAnswer.FAILED_TO_ANSWER} Agent specified 0 tool"
+                            " calls, which means done."
```
mskarlin, line 172: Should we say "... which means unsure."?

jamesbraza: Yes this is a great point. I need to really think about why we have `GenerateAnswer.FAILED_TO_ANSWER`, it seems like it shouldn't exist.

I will circle back on this tomorrow after some sleep

paperqa/settings.py
jamesbraza: Okay, went with your suggestion here, thanks!

paperqa/agents/env.py
jamesbraza: Okay I updated this to be:

```python
return (
    # NOTE: don't put:
    # - GenerateAnswer.FAILED_TO_ANSWER here because this wasn't a failure
    # - 'cannot answer' because that information belongs in
    #   PQASession.answer, not in the message history
    # Let's just put a nice message about being done :)
    [Message(content="Agent specified 0 tool calls, which means done.")],
    ...
)
```

Hopefully that makes sense

paperqa/agents/env.py
```diff
@@ -169,30 +170,40 @@ def _has_excess_answer_failures(self) -> bool:
...
+            self.USE_POST_PROCESSED_REWARD,
             any(
                 isinstance(msg, ToolResponseMessage)
                 and msg.name == GenerateAnswer.gen_answer.__name__
                 and GenerateAnswer.did_not_fail_to_answer(msg.content)
                 for msg in response_messages
             )
             or self._has_excess_answer_failures(),
-            False,
+            False,  # Let caller determine truncations
```
mskarlin, line 206: nice

mskarlin (approved):

jamesbraza:
I ran some local testing and it seems if anything performance got better

PR #672, opened 2024-11-08
`fake` agent allowing timeouts or exceptions, 

jamesbraza:
Easier to read diff: https://github.com/Future-House/paper-qa/compare/standard-rollouts?expand=1&w=1

This PR:
- Creates a helper function `run_with_timeout_failure` that runs a rollout with `AgentStatus.TRUNCATED` (timeout) or `AgentStatus.FAIL` (`Exception`) support
- Moves all three runners (fake, aviary, LDP) to use it
    - Now the fake agent supports timeouts or failures 🥳 
- Adds `AgentStatus.TRUNCATED` support for the LDP agent runner

### DIFF
--- a/paperqa/agents/main.py
+++ b/paperqa/agents/main.py
@@ -142,6 +142,29 @@ async def run_agent(
     return AnswerResponse(session=session, status=agent_status)
 
 
+async def _run_with_timeout_failure(
+    rollout: Callable[[], Awaitable[AgentStatus]],
+    query: QueryRequest,
+    env: PaperQAEnvironment,
+) -> tuple[PQASession, AgentStatus]:
+    try:
+        async with asyncio.timeout(query.settings.agent.timeout):
+            status = await rollout()
+    except TimeoutError:
+        logger.warning(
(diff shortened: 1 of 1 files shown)


paperqa/agents/main.py
```diff
@@ -325,39 +339,31 @@ async def run_ldp_agent(
...
         )
-        await generate_answer_tool._tool_fn(question=query.query, state=env.state)
-    except Exception:
-        logger.exception(f"Agent {agent} failed.")
-        status = AgentStatus.FAIL
-    return env.state.session, status
+        traj = trajs[0]
+        if traj.steps[-1].truncated:
+            nonlocal status
+            status = AgentStatus.TRUNCATED
```
whitead, line 364: This is confusing - can you just return status in `rollout`? This modification is like three levels deep and I'm actually unclear on if it always works

jamesbraza: Believe it or not, locally I have this exact change in the works

jamesbraza: Okay, done. Good to see we are both still thinking

whitead (approved):
cool

PR #673, opened 2024-11-08
Optional `AnswerSetting.max_answer_attempts` to allow a new unsure branch

jamesbraza:
This PR adds in an optional `AnswerSetting.max_answer_attempts` so we can declare ourselves as unsure if we attempt `gen_answer` too many times, with test coverage.

To do so, it tracks a tool names used in environment state as `tool_history`

### DIFF
--- a/tests/test_task.py
+++ b/tests/test_task.py
@@ -159,18 +159,21 @@ class TestTaskDataset:
             base_query_request.settings.agent.tool_names = {
                 GenerateAnswer.gen_answer.__name__
             }
+            base_query_request.settings.answer.max_answer_attempts = 2
             base_query_request.settings.answer.get_evidence_if_no_contexts = False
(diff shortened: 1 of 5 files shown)


tests/test_task.py
```diff
@@ -159,18 +159,21 @@ async def test_evaluation(
...
-                config=EvaluatorConfig(batch_size=len(dataset), max_rollout_steps=2),
+                config=EvaluatorConfig(batch_size=len(dataset), max_rollout_steps=4),
                 agent=SimpleAgent(),
                 dataset=dataset,
                 callbacks=[storage_callback],
             )
             await evaluator.evaluate()
             for traj in storage_callback.eval_trajectories:
+                assert not traj.failed
+                assert traj.done
```
jamesbraza, line 176: This specific assertion will not pass before this PR without this new setting

mskarlin (approved):
ok nice -- looks good.

PR #680, opened 2024-11-13
Refactor to remove `skip_system` from `LLMModel.run_prompt`

jamesbraza:
When reading through our various control flows here, I found the presence of a system prompt coupled with a `skip_system` flag to add more thought and complexity. We don't need to cart around two separate pieces of information here, we can just make `system_prompt` be an optional variable and one source of truth.

This PR is a refactor that removes `skip_system` in favor of `bool` checks directly on `system_prompt`. Yes it's a breaking change, but I don't think our `LLMModel` implementation is really imported and used by our users.


mskarlin (approved):
👍 nice refactor

PR #684, opened 2024-11-13
Created `complete` tool to allow unsure answers

jamesbraza:
### Part 1

Based on our `Environment.step`'s `done` condition: https://github.com/Future-House/paper-qa/blob/v5.4.0/paperqa/agents/env.py#L199-L205

We currently (as of v5.4) incorrectly conclude a rollout is `done` on `gen_answer`'s `answer`s such as:

- "Based on the sources provided, it appears no one has done x."
- "The sources provide some evidence about gene X, but gene Y is not discussed."

### Part 2

We realized that https://github.com/Future-House/paper-qa/pull/671 is:

- Useful for agents backed by LLM `tool_choice="auto"` (OpenAI default). An example agent here is LangChain's `OpenAIFunctionsAgent`
- Not useful for agents backed by LLM `tool_choice="required"`, since they cannot specify empty tool calls. This includes our `ldp` 0.14 `Agent`s or `aviary` 0.10 `ToolSelector`

### Part 3

In general, empty tool calls signifying `done` is probably not a generalized assumption.

## Implementation

To be generally applicable, here we introduce another tool, the `complete` tool. When invoked, this tool signifies the rollout is `done`. This enables:

- The unsure answers (see motivation part 1 above) to be non-terminal/intermediary outputs
- Simplified `PaperQAEnvironment.step` such that it doesn't special case `done` logic for empty tool calls, which in turn enables simplification of the `PaperQAEnvironment.reset` observations
- Trajectories not ending in `complete` are now clear truncations or failures

Notably this change also simplifies `GradablePaperQAEnvironment.step`, now we can directly parse the `state.session.answer` when `done`, as opposed to parsing the `messages`.

The tradeoffs here are due to a fifth tool being added, which:

- Increases API prompt token usage
- Increases the number of moving parts in the system


paperqa/agents/env.py
```diff
@@ -39,17 +40,17 @@ def settings_to_tools(
...
 
-    NOTE: the last element of the return will always be GenerateAnswer.
+    NOTE: the last element of the return will always be Complete.
     """
     llm_model = llm_model or settings.get_llm()
     summary_llm_model = summary_llm_model or settings.get_summary_llm()
     embedding_model = embedding_model or settings.get_embedding_model()
     tools: list[Tool] = []
-    has_answer_tool = False
+    has_complete_tool = False
```
whitead, line 51: Is this check really necessary? If any of the tools are missing, it will fail. No PaperSearch means no papers, no GatherEvidence means no contexts, etc.

whitead: Ah I see - it's for the special zero-shot setting.

jamesbraza: Sure, I made https://github.com/Future-House/paper-qa/pull/685 to keep this PR minimal

jamesbraza: > Ah I see - it's for the special zero-shot setting.

The validation is really to protect:

- Clients who are doing edge case things (e.g. using a subset of the tools, and they should also override the `agent_prompt`)
- Unit tests from failing for unexpected reasons
    - However, this validation usually complicates life for unit test writers, not makes it better

whitead (approved):
Thanks for doing this - I think this is the best approach. Assuming perf looks same, this LGTM!

paperqa/agents/env.py
jamesbraza: Alright, `git rebase`d atop https://github.com/Future-House/paper-qa/pull/685, resolving this one

jamesbraza:
Thanks @whitead for the LGTM, appreciated.

@mskarlin and I were discussing today, and we figured out we can get rid of the `"cannot answer"` string literal checking in the code base by moving the `complete` tool to have a `bool` argument `is_sure` that basically plays the role of `AgentStatus.UNSURE`.

In other words, we're planning to move the "unsure" definition from the environment to the agent.

jamesbraza:
I decided to resolve the future comments in another PR. Going to merge this one as it's a somewhat atomic change

tests/test_agents.py
```diff
@@ -740,15 +778,21 @@ async def test_deepcopy_env(self, agent_test_settings: Settings) -> None:
...
+        await env.step(gen_answer_action)
+        _, _, done, _ = await env.step(
+            ToolRequestMessage(tool_calls=[ToolCall.from_name("complete")])
+        )
         assert done
         assert not env.state.session.could_not_answer
         assert env.state.session.used_contexts
-        _, _, done, _ = await env_copy.step(gen_answer_action)
+        await env_copy.step(gen_answer_action)
+        _, _, done, _ = await env.step(
```
jamesbraza, line 793: This should have `env_copy.step`, fixed in https://github.com/Future-House/paper-qa/pull/698

PR #685, opened 2024-11-14
Removed `tool_names` validation for `gen_answer` being present

jamesbraza:
If any of the tools are excluded, PaperQA will not function. They all play a distinct and required role (`paper_search` to get new papers, `gather_evidence` to get relevant evidence, and `gen_answer` for answering), and none are optional.

We can either:
- Validate for all required tools, leading to an explicit failure
    - Perhaps in the future when we add the citation traversal tool, this tool will be optional.
- Validate for none of them, leading to an implicit failure, as if any tools are unspecified the agent will either get truncated or fail

We can YAGNI on the first route, so this PR takes the second route as suggested in this PR comment: https://github.com/Future-House/paper-qa/pull/684#discussion_r1841332919

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -47,7 +47,6 @@ def settings_to_tools(
     summary_llm_model = summary_llm_model or settings.get_summary_llm()
     embedding_model = embedding_model or settings.get_embedding_model()
     tools: list[Tool] = []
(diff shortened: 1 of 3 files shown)


whitead (approved):

mskarlin (approved):
agreed w your logic here

PR #687, opened 2024-11-14
Implement support to BatchAPIs to gather evidence

maykcaldas:
This PR implements support to send requests to OpenAI and Anthropic batch APIs. Due to the parallel nature of gathering evidence and summarizing all candidate papers, we plan to use the batch API when possible.

The use of the batch API depends on `Settings.use_batch_in_summary`. Therefore, paperqa workflows would still be unchanged in case this setting is set to False (default). Currently, using a batch keeps the process busy while the batch isn't finished on the LLM provider side, which could take up to 24 hours. This scaling issue will be addressed in another PR.

Task list
- [x] Create a class to make batch calls to openai
- [x]  Create a class to make batch calls to anthropic
- [x]  Integrate the openai class to the get_evidence method
- [x]  Integrate the anthropic class to the get_evidence method
- [x]  Update `get_summary_llm` to decide which provider to use given the llm in the config
- [ ]  ❌ Use `pytest.mark.vcr` in the tests to avoid creating batches for every test 
- [x] Implement mock servers for testing purposes


paperqa/llms.py
```diff
@@ -761,6 +767,207 @@ def infer_llm_type(self) -> str:
...
+    async def acomplete_iter(self):
+        raise NotImplementedError("Async generator not supported for batch calls and nly chat models are supported by openAI batch API.")
+
+    async def _run_chat(
+        self, 
+        prompt: str,
+        data: list[dict[str,str]],
+        callbacks: list[Callable] | None = None,
+        name: str | None = None,
+        skip_system: bool = False,
```
jamesbraza, line 814: I refactored out `skip_system` in https://github.com/Future-House/paper-qa/pull/680, can you propagate that change to here?

paperqa/llms.py
```diff
@@ -761,6 +767,207 @@ def infer_llm_type(self) -> str:
...
+
+        return results
+
+    @rate_limited
+    async def achat(self,
+        messages: list[dict[str, str]]
+    ) -> list[Chunk]:
+        client = openai.OpenAI()
+
+        with tempfile.NamedTemporaryFile(suffix=".jsonl", delete=True) as tmp_file:
```
jamesbraza, line 867: I think `delete=True` is the default arg, no need to keep it as it's implied

paperqa/settings.py
```diff
@@ -609,6 +618,10 @@ class Settings(BaseSettings):
             " router_kwargs key with router kwargs as values."
         ),
     )
+    use_batch_in_summary: bool = Field(
+        default=False,
+        description="Whether to use batch API for LLMs in summarization",
```
jamesbraza, line 623: Can you add a few words on how the batches are actually formed

paperqa/docs.py
```diff
@@ -600,23 +603,66 @@ async def aget_evidence(
...
-                        prompt_runner=prompt_runner,
-                        extra_prompt_data={
-                            "summary_length": answer_config.evidence_summary_length,
-                            "citation": f"{m.name}: {m.doc.formatted_citation}",
-                        },
-                        parser=llm_parse_json if prompt_config.use_json else None,
-                        callbacks=callbacks,
-                    )
+            if evidence_settings.use_batch_in_summary:
+                # TODO: Should we implement a `gather_with_batch` function that receives `matches` and return results to keep this dry?
```
mskarlin, line 607: I like this idea

paperqa/llms.py
```diff
@@ -761,6 +767,207 @@ def infer_llm_type(self) -> str:
...
+        batch = client.batches.create(
+            input_file_id=file.id,
+            endpoint="/v1/chat/completions",
+            completion_window="24h",
+            metadata={
+                "description": ""
+            }
+        )
+
+        while batch.status != "completed":
```
mskarlin, line 884: we probably want "completed" and "failed" to be OpenAI enums here rather than free strings.

paperqa/llms.py
```diff
@@ -761,6 +767,207 @@ def infer_llm_type(self) -> str:
...
+            batch = client.batches.retrieve(batch.id)
+            if batch.status == "failed":
+                raise Exception("Batch failed. \n\nReason: \n" + "\n".join([k.message for k in batch.errors.data]))
+            await asyncio.sleep(5)
```
mskarlin, line 888: Let's parameterize this waiting, and maybe make the default longer? like 30 second or 1 min polling? 

We should probably add some debug/info logs here to track progress along with maybe a max-timeout which users can set.

paperqa/docs.py
maykcaldas: I refactored it

paperqa/llms.py
maykcaldas: Yeah, that's the default. I removed it

paperqa/settings.py
jamesbraza: Perhaps you can say something like:

> Whether to use batch API for LLMs in summarization, which means multiple messages are sent in one API request.

maykcaldas: It was updated to:
```
"Whether to use batch API for LLMs in summarization, "
"which means multiple messages are sent in one API request "
"to the LLM provider's batch API."
"This option is only available for Claude(https://docs.anthropic.com/en/api/creating-message-batches)"
"and OpenAI (https://platform.openai.com/docs/guides/batch) chat models."
```

paperqa/llms.py
maykcaldas: Added logging and two settings variables to control the timelimit and the pooling

paperqa/core.py
```diff
@@ -115,3 +115,50 @@ async def map_fxn_summary(
         ),
         llm_result,
     )
+
+async def gather_with_batch(
```
jamesbraza, line 119: Can we merge this with `gather_with_concurrency`? This is still concurrent gathering, just the inner workings are via batching.

tests/test_llms.py
```diff
@@ -159,6 +166,199 @@ def test_pickling(self, tmp_path: pathlib.Path) -> None:
...
+        ##############################
+        #                            #
+        # Create a mock batch client #
+        #                            #
+        ##############################
+
+        mock_client = AsyncMock()
+    
+        # Define mock methods for the client
+        mock_files_create = AsyncMock()
```
jamesbraza, line 198:
```python
        mock_file_id = 'file-123'
        mock_client.files.create = AsyncMock(return_value=mock_file_id)
```

And when you refer to `mock_files_content` below, you'd change to `mock_client.files.create`. Sometimes it's easier to read the actual path, instead of temporary names like `mock_files_create`

tests/test_llms.py
```diff
@@ -159,6 +166,199 @@ def test_pickling(self, tmp_path: pathlib.Path) -> None:
...
+        
+        mock_client.files.create = mock_files_create
+        mock_client.batches.create = mock_batches_create
+        mock_client.batches.retrieve = mock_batches_retrieve
+        mock_client.files.content = mock_files_content
+
+        mock_file_id = 'file-123'
+        mock_files_create.return_value = MagicMock(id=mock_file_id)
+        
+        mock_batch_id = 'batch_123'
```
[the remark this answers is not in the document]
maykcaldas: I'll add a comment. It is exactly to match the real API

tests/test_llms.py
```diff
@@ -159,6 +166,199 @@ def test_pickling(self, tmp_path: pathlib.Path) -> None:
...
+        }
+
+    @pytest.mark.vcr
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "config",[
+            pytest.param("claude-3-haiku-20240307", id="chat-model"),
+        ], indirect=True
+    )
+    async def test_run_prompt(self, config: dict[str, Any], request) -> None:
```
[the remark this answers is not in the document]
maykcaldas: Yeah, this test wasn't finished in this commit. I'll push the complete test with the anthropic mocked API soon

tests/test_llms.py
```diff
@@ -159,6 +166,199 @@ def test_pickling(self, tmp_path: pathlib.Path) -> None:
...
 
+class TestOpenAIBatchLLMModel:
+    @pytest.fixture(scope="class")
+    def config(self, request) -> dict[str, Any]:
+        model_name = request.param
+        return {
+            "model": model_name,
+            "temperature": 0.0,
+            "max_tokens": 64,
+            "batch_summary_timelimit": 24*60*60,
```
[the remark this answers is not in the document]
maykcaldas: Because the API itself has a timelimit of 24h. I just matched it. 
Also, if we reach the timelimit, paperqa raises an exception that is not caught by the test.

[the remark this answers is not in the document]
jamesbraza: GitHub Actions only run a step (in this case, our `pytest` step) for 6 or 12 hrs by default, so I think 24 hours is too long

tests/test_llms.py
```diff
@@ -159,6 +166,199 @@ def test_pickling(self, tmp_path: pathlib.Path) -> None:
...
+    )
+    @pytest.mark.asyncio
+    async def test_run_prompt(self, monkeypatch, config: dict[str, Any], request) -> None:
+        ##############################
+        #                            #
+        # Create a mock batch client #
+        #                            #
+        ##############################
+
+        mock_client = AsyncMock()
```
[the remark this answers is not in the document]
maykcaldas: It seems that spec_set receives an instance of the API package. So I added `spec_set=openai.AsyncOpenAI()`. Did the same for anthropic

tests/test_llms.py
maykcaldas: Thanks for the suggestion. It's indeed much cleaner

paperqa/docs.py
```diff
@@ -623,28 +624,40 @@ async def aget_evidence(
...
+                        )
+                        for m in matches
+                    ],
+                )
 
         for _, llm_result in results:
             session.add_tokens(llm_result)
 
-        session.contexts += [r for r, _ in results if r is not None]
+        session.contexts += [r for r, _ in results]
```
mskarlin, line 660: why did we cut the `r is not None` filter here? I would think that the results from `gather_with_concurrency` could still be `None` on failure, but maybe I'm wrong

paperqa/llms.py
```diff
@@ -70,6 +75,24 @@ class EmbeddingModes(StrEnum):
     QUERY = "query"
 
 
+class OpenAIBatchStatus(StrEnum):
```
mskarlin, line 78: Minor point, but we could make this a bit DRY-er by making a single BatchStatus(StrEnum), with methods that convert OpenAI and Anthropic statuses, so like: 
```python
class BatchStatus(StrEnum):
    COMPLETE = 'complete'
    PROGRESS = 'progress'
    ....
    @classmethod
    def from_anthropic(cls, status: str):
        return {'completed': cls.COMPLETE,
                    'in_progress': cls.PROGRESS, ...
                     }
```

.github/workflows/tests.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           enable-cache: true
       - run: uv python pin ${{ matrix.python-version }}
       - uses: hynek/build-and-inspect-python-package@v2
-      - run: uv sync --python-preference=only-managed
+      - run: uv sync --python-preference=only-managed --all-extras
```
jamesbraza, line 38: So this change is suboptimal because what will happen for local users is they will run the test but won't have the `[batch]` extra because local users just run `uv sync`.

What we should do here is in `tool.uv`:

```toml
[dependency-groups]
dev = [
    ...
    "paper-qa[datasets,batch,ldp,typing,zotero,local]",
    ...
]
```

And then all use cases will be covered. No need for the `--all-extras`

paperqa/core.py
maykcaldas: `gather_with_concurrency ` is a wrapper to actually use `map_fxn_summary`. I think we can unify `gather_with_batch` and  `map_fxn_summary` because they work very similarly. But because the inputs and returns are different, I kept them separated to make the the static type definition easier

paperqa/docs.py
maykcaldas: This gets the `Context`s  from `gather_with_concurrency ` or `gather_with_batch`. And both always return list of tuples with (`Context`, `LLMResult`). What can happen is to have an empty text in `Context.text`, but it seems to me that r is always an instance of `Context`.
Also, I didn't see any case of `map_fxn_summary` returning `None` while studying the code, and mypy also complains that `r is None` is always a True statement.  

Maybe that's an edge case that I didn't see?

.github/workflows/tests.yml
maykcaldas: We have a similar declaration in `project.toml`, but I found no `tool.uv`. Are they the same thing?

jamesbraza: Ah crap, yeah it's `[dependency-groups]` now, yeah that was changed recently

paperqa/llms.py
maykcaldas: I think we want to do the opposite, right? We want to have something like `BatchStatus.from_<provider>.COMPLETE` that returns "completed" to openai and "ended" for anthropic. Maybe:

```python
class BatchStatus(StrEnum):
    COMPLETE = ""
    ...
    def from_openai(self) -> str:
        mapping = {
            BatchStatus.COMPLETE: "completed"
            ...
        }
        return mapping[self]
        
    def from_anthropic(self) -> str:
        mapping = {
            BatchStatus.COMPLETE: "ended"
            ...
        }
        return mapping[self]
```

then we can use it as `BathStatus.COMPLETE.from_openai()`

paperqa/core.py
```diff
@@ -115,3 +116,75 @@ async def map_fxn_summary(
...
+    parser: Callable[[str], dict[str, Any]] | None = None,
+    callbacks: list[Callable[[str], None]] | None = None,
+) -> list[tuple[Context, LLMResult]]:
+    """
+    Gathers evidence considering a batch of texts. The completions are obtained using a batch API.
+
+    Args:
+        matches (list[Text]): A list of text matches to gather evidence from.
+        question (str): The question to be answered.
+        prompt_runner (PromptRunner | None): The prompt runner to use for obtaining completions.
```
jamesbraza, line 135: When doing docstrings, if you put type hints in the signature (which you did, and which is the right move), don't duplicate the types in the docstring.

What happens is later on, people will update the type hint in the signature, but miss they were needed to also update the docstring

So make this:

> prompt_runner: Optional prompt runner to use for obtaining completions.

paperqa/core.py
```diff
@@ -115,3 +116,75 @@ async def map_fxn_summary(
...
+        else:
+            results_data.append({})
+            scores.append(extract_score(r.text))
+
+    return [
+        (
+            Context(
+                context=strip_citations(llm_result.text),
+                text=m,
+                model_extra={},
```
jamesbraza, line 181: We don't need `model_extra`, unless I am missing something? What even is `model_extra`?

paperqa/settings.py
```diff
@@ -612,6 +627,24 @@ class Settings(BaseSettings):
...
+        description=(
+            "Whether to use batch API for LLMs in summarization, "
+            "which means multiple messages are sent in one API request "
+            "to the LLM provider's batch API."
+            "This option is only available for Claude(https://docs.anthropic.com/en/api/creating-message-batches)"
+            "and OpenAI (https://platform.openai.com/docs/guides/batch) chat models."
+        ),
+    )
+    batch_summary_timelimit: int = Field(
+        default=24 * 60 * 60,
```
jamesbraza, line 641: Can you add a comment saying this matches OpenAI's default?

paperqa/settings.py
```diff
@@ -612,6 +627,24 @@ class Settings(BaseSettings):
...
+        default=False,
+        description=(
+            "Whether to use batch API for LLMs in summarization, "
+            "which means multiple messages are sent in one API request "
+            "to the LLM provider's batch API."
+            "This option is only available for Claude(https://docs.anthropic.com/en/api/creating-message-batches)"
+            "and OpenAI (https://platform.openai.com/docs/guides/batch) chat models."
+        ),
+    )
+    batch_summary_timelimit: int = Field(
```
jamesbraza, line 640: Why not name it `batch_summary_time_limit`? It seems weird to put no `_` between time and limit

paperqa/settings.py
maykcaldas: Added it in the description

paperqa/core.py
maykcaldas: Makes sense. Updated it

paperqa/settings.py
maykcaldas: No reason 😅 I just typed it like that and didn't think deeply about it. 
I updated the name

paperqa/docs.py
```diff
@@ -629,28 +630,40 @@ async def aget_evidence(
...
-                ],
-            )
+            if evidence_settings.use_batch_in_summary:
+                results = await gather_with_batch(
+                    matches=matches,
+                    question=session.question,
+                    prompt_runner=prompt_runner,
+                    extra_prompt_data={
+                        "summary_length": answer_config.evidence_summary_length,
+                    },
```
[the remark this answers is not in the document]
maykcaldas: Because the else processes only one context. So it can have the citation for that text. The `gather_with_batch` have multiple contexts, possibly from different sources. So citation is computed inside the function

paperqa/core.py
```diff
@@ -68,12 +68,13 @@ async def map_fxn_summary(
...
 
     if prompt_runner:
-        llm_result = await prompt_runner(
+        result = await prompt_runner(
             {"question": question, "citation": citation, "text": text.text}
             | (extra_prompt_data or {}),
             callbacks,
             "evidence:" + text.name,
         )
+        llm_result = result if isinstance(result, LLMResult) else result[0]
```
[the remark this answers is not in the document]
jamesbraza: Yeah the question is, what happens if 2+ results are returned? Wouldn't we just be throwing away useful data?

Can you make it safe by:

```python
if isinstance(result, Sequence) and len(result) != 1:
    raise NotImplementedError(f"We didn't handle blah blah {result} blah blah.")
llm_result = result if isinstance(result, LLMResult) else result[0]
```

In general, the design philosophy here is:

> It works, because it must work

paperqa/core.py
```diff
@@ -115,3 +122,75 @@ async def map_fxn_summary(
...
+    return [
+        (
+            Context(
+                context=strip_citations(llm_result.text),
+                text=m,
+                score=score,
+                **r,
+            ),
+            llm_result,
+        )
```
jamesbraza, line 192: Okay I think there's an extra parenthesis here, can you remove the `(Context())`

paperqa/llms.py
```diff
@@ -764,6 +802,378 @@ async def select_tool(
         return await tool_selector(*selection_args, **selection_kwargs)
 
 
+class LLMBatchModel(ABC, BaseModel):
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+
+    llm_type: str | None = None
```
jamesbraza, line 808: Can you add a `Field(description="...")` what this should be when it's a `str`?

Imo "type" wording makes me think it should be a `class`, but it's actually really more of an `Enum`. Maybe we should have this be an `str | Enum | None`

paperqa/llms.py
```diff
@@ -764,6 +802,378 @@ async def select_tool(
...
+        }
+
+        for i, d in enumerate(data):
+            batch_template["custom_id"] = str(i)
+            batch_template["body"]["model"] = self.config.get("model")
+            batch_template["body"]["messages"] = d
+            batch_template["body"]["max_tokens"] = self.config.get("max_tokens")
+            serialized_data = json.dumps(batch_template) + "\n"
+            tmp_file.write(serialized_data.encode())
+            # tmp_file.write(json.dumps(batch_template) + "\n")
```
jamesbraza, line 908: Can you remove stale comments

paperqa/llms.py
```diff
@@ -764,6 +802,378 @@ async def select_tool(
...
+        try:
+            import openai
+        except ImportError as exc:
+            raise ImportError(
+                "Please install paper-qa[batch] to use OpenAIBatchLLMModel."
+            ) from exc
+
+        client = openai.AsyncOpenAI()
+
+        with io.BytesIO() as tmp_file:
```
jamesbraza, line 978: This isn't a `tmp_file` (it's not a file in the `/tmp` dir), let's call it `mem_buffer` or `mem_file`

paperqa/llms.py
```diff
@@ -764,6 +802,378 @@ async def select_tool(
...
+    async def achat_iter(self):
+        raise NotImplementedError(
+            "Async generator not supported for batch calls. Use achat instead."
+        )
+
+
+class AnthropicBatchLLMModel(LLMBatchModel):
+    """A wrapper around the anthropic library to use the batch API."""
+
+    name: str = "claude-3-5-sonnet-20241022"
```
jamesbraza, line 1048: Can you add a `description` why we are using an exact name here? Imo we should prolly just use `claude-3-5-sonnet-latest ` (and not specify a date)

paperqa/llms.py
maykcaldas: oh, nice catch

paperqa/core.py
maykcaldas: Is there? It's a tuple with (Context, LLMResult). The `llm_result,` on line 192 is outside the Context()

paperqa/llms.py
maykcaldas: Yeah, it should be `chat` or `completion.` So it's a `str` with the model's "type." 
It is checked on inference time. If `llm_type` is `None`, then a `infer_llm_type` is called. But I added a description explaining the attribute in case the user wants to define it. Also changed the declaration in `LiteLLMModel`

paperqa/core.py
jamesbraza: Oh crap I misread, you're right. Resolving this one

whitead:
Closing for now - probably no longer relevant.

PR #688, opened 2024-11-15
Allowing serialized `Settings` in `get_settings`

jamesbraza:
This allows me to use this function with config files

### DIFF
--- a/paperqa/settings.py
+++ b/paperqa/settings.py
@@ -884,12 +884,18 @@ class Settings(BaseSettings):
         raise NotImplementedError(f"Didn't yet handle agent type {agent_type}.")
 
 
-MaybeSettings = Settings | str | None
+# Settings: already Settings
+# dict[str, Any]: serialized Settings
+# str: named Settings
+# None: defaulted Settings
+MaybeSettings = Settings | dict[str, Any] | str | None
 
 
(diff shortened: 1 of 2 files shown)


nadolskit (approved):

paperqa/settings.py
```diff
@@ -884,12 +884,18 @@ async def make_ldp_agent(
...
+# str: named Settings
+# None: defaulted Settings
+MaybeSettings = Settings | dict[str, Any] | str | None
 
 
 def get_settings(config_or_name: MaybeSettings = None) -> Settings:
     if isinstance(config_or_name, Settings):
         return config_or_name
+    if isinstance(config_or_name, dict):
+        return Settings(**config_or_name)
```
mskarlin, line 898: nit but I'm more partial to `.model_validate`

mskarlin (approved):

paperqa/settings.py
jamesbraza: Alright moved to it 👌 

Out of curiosity, why do you prefer `model_validate` here? I was playing around, and couldn't actually find a place where `Settings(**config_or_name)` validation fails while `Settings.model_validate(config_or_name)` succeeds

PR #694, opened 2024-11-15
Making sure we copy distractors

sidnarayanan:
If `LitQAEvaluation.from_question(distractors=...)` received a list, it would be modified in-place as we constructed the option list. This PR fixes that and adds tests.

### DIFF
--- a/paperqa/litqa.py
+++ b/paperqa/litqa.py
@@ -212,8 +212,8 @@ def read_litqa_v2_from_hub(
         from datasets import load_dataset
     except ImportError as exc:
         raise ImportError(
-            "Reading in LitQA2 requires the 'datasets' extra for 'datasets'. Please: "
-            "`pip install paper-qa[datasets]`."
+            "Reading in LitQA2 requires the 'datasets' extra for 'datasets'. Please:"
+            " `pip install paper-qa[datasets]`."
         ) from exc
 
     litqa_v2 = load_dataset(labbench_dataset, "LitQA2")["train"].to_pandas()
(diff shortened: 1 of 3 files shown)


paperqa/litqa.py
```diff
@@ -43,13 +43,13 @@ def make_mc_options(
...
-        options = distractors
+        # We are going to modify options in-place, so copy it first
+        options = distractors.copy()
 
     if ideal == "null":
         if not unsure_option:
             raise ValueError(
-                'Dataset configured for "unsure" options via '
-                'ideal="null", please specify "unsure_option".'
+                'Dataset configured for "unsure" options via ideal="null", please specify "unsure_option".'
```
jamesbraza, line 52: Can you undo these string wrapping changes (and messages below)? It no longer fits in 88 chars with these

mskarlin (approved):
lgtm -- do we need the uv.lock in this one? jc

tests/test_litqa.py
```diff
@@ -1,36 +1,85 @@
...
+class TestLitQAEvaluation:
+    @staticmethod
+    def _assert_prompt_is_valid(
+        qa_prompt: str, question: str, ideal: str, distractors: list[str]
+    ):
+        for substr in (question, "Insufficient information", ideal, *distractors):
+            assert qa_prompt.count(substr) == 1
+
+    @pytest.mark.asyncio
+    async def test_from_question(self) -> None:
```
jamesbraza, line 17: Can you add `pytest.mark.vcr` here?

jamesbraza (approved):

PR #697, opened 2024-11-17
Default to ordered tool calls, w env variable control

mskarlin:
We've been seeing a production bug where parallel gather_evidence tool calls are clobbering our environment state. We can try to prompt engineer our way out of this, but deterministically forcing ordered tool calls feels safer. 

This change forces them into ordered calls along with a test. We can turn it off with an env var if we'd like in the future. I don't think this should go into settings since it's effectively a bug to turn on for now.

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from copy import deepcopy
 from typing import Any, ClassVar, Self, cast
 
@@ -95,11 +94,6 @@ def settings_to_tools(
     return tools
 
 
-ALLOW_PARALLEL_TOOL_CALLS: bool = os.environ.get(
-    "PQA_ALLOW_PARALLEL_TOOL_CALLS", ""
-).lower() in {"true", "1"}
-
-
 class PaperQAEnvironment(Environment[EnvironmentState]):
     """Environment connecting paper-qa's tools with state."""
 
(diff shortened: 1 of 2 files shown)


paperqa/agents/env.py
```diff
@@ -94,6 +95,11 @@ def settings_to_tools(
     return tools
 
 
+ALLOW_PARALLEL_TOOL_CALLS: bool = os.environ.get(
+    "PQA_ALLOW_PARALLEL_TOOL_CALLS", ""
+).lower() in {"true", "1"}
+
+
```
whitead, line 102: How about we just YAGNI this and remove it? If we want to explore later, we can make it a setting.

mskarlin: I don't feel strongly abt it -- can delete it

whitead (approved):

tests/test_agents.py
```diff
@@ -749,6 +750,46 @@ def test_answers_are_striped() -> None:
...
+                    ),
+                    ToolCall.from_name(
+                        "gather_evidence",
+                        question="XAI for chemical property prediction",
+                    ),
+                ]
+            )
+        )
+
+        assert time.time() - tic > 2 * SLEEP_TIME  # since they are sequential
```
jamesbraza, line 790: Nice PR and happy to see a test change. 

This test directly mirrors the updated implementation (`ordered=True`) not so much the actual issue (concurrent `gather_evidence` clobbering `state.session.question`). This pattern means this new test isn't useful if we want to fix the race condition(s) in our tools and remove `ordered=True`.

It might be better to move the `gather_evidence` stub here to manipulate `state.session.question` 2+ times to emulate the race condition being faced.

Also, would be good to use `GradablePaperQAEnvironment` and then move this to the below test class `TestGradablePaperQAEnvironment`

PR #698, opened 2024-11-18
Removed `"cannot answer"` literals and added `reset` tool

jamesbraza:
There many of places where we depended on the string  literal `"cannot answer"` in the `qa` prompt, mainly the environment being done (prior to https://github.com/Future-House/paper-qa/pull/684) or the `answer` being considered unsure.

This environment check of `"cannot answer"` also has some downsides:

1. A coupling of environment functionality to a caller-specified `qa` prompt
    - We don't validate that `"cannot answer"` is present in the `qa` prompt
2. Added statefulness to our environment (checks for a string literal in `answer`)

So, what is "unsure"? Really it should be:

1. `gen_answer` tool call updates the answer
2. Given the answer, the _agent_ (and not the environment's `gen_answer` tool) decides if it's unsure
3. If unsure, agent keeps trying to get better evidence, until it gives up

To resolve this, we moved the unsure call directly to the `complete` tool. Now:

- When unsure: agent just keeps running
- When finally sure: agent calls `complete(is_sure=True)`
- If giving up: agent calls `complete(is_sure=False)`

The `"cannot answer"` check was mostly easy to remove, other than the `AnswerSettings.wipe_context_on_answer_failure`, since we no longer have a way of checking unsure within `gen_answer`.

Since the agent controls unsureness now, we needed to make a new tool: `reset`, which basically performs the use case of `wipe_context_on_answer_failure`.

---

After this PR, we have:
- Removed dependence on `"cannot answer"` string literal
- Deprecates `AnswerSettings.wipe_context_on_answer_failure`
- Agent defines unsureness, not the output of the environment's `gen_answer` tool
- A "learnable" dimension, the agent controlling wiping contexts


paperqa/agents/tools.py
```diff
@@ -336,31 +330,53 @@ def extract_answer_from_message(cls, content: str) -> str:
...
+
+
+class Reset(NamedTool):
+    TOOL_FN_NAME = "reset"
+
+    async def reset(self, state: EnvironmentState) -> None:
+        """
+        Reset by clearing all current evidence from the system.
+
+        This tool is useful when repeatedly failing to answer because new evidence can give new perspective.
```
mskarlin, line 343: I might say: "This tool is useful when repeatedly failing to answer because the existing evidence may unsuitable for the question."

paperqa/prompts.py
```diff
@@ -13,13 +13,17 @@
...
+# 2. Can be used for unit testing
+CANNOT_ANSWER_PHRASE = "I cannot answer"
 qa_prompt = (
     "Answer the question below with the context.\n\n"
     "Context (with relevance scores):\n\n{context}\n\n----\n\n"
     "Question: {question}\n\n"
     "Write an answer based on the context. "
     "If the context provides insufficient information reply "
-    '"I cannot answer."'
+    f'"{CANNOT_ANSWER_PHRASE}." '
```
mskarlin, line 26: nice!

paperqa/agents/tools.py
mskarlin: Only because new evidence can always be added, even without using this tool.

paperqa/agents/env.py
```diff
@@ -49,7 +50,7 @@ def settings_to_tools(
     embedding_model = embedding_model or settings.get_embedding_model()
     tools: list[Tool] = []
     for tool_type in (
-        (PaperSearch, GatherEvidence, GenerateAnswer, Complete)
+        (PaperSearch, GatherEvidence, GenerateAnswer, Reset, Complete)
```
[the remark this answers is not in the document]
jamesbraza: Yep agreed

paperqa/agents/tools.py
jamesbraza: Nice I see what you're saying, implemented

mskarlin (approved):

PR #703, opened 2024-11-19
Fixing `PaperQAEnvironment.reset` respecting `mmr_lambda` and `text_hashes`

jamesbraza:
This PR started as a regression test of `PaperQAEnvironment.reset`, then we found out some bugs were present.

So it writes the unit test, and to make it pass:
- Adds `Docs.__eq__` so we can compare `Docs` in unit tests
- Consolidates `mmr_lambda` defaults to 1.0 (disabled)
- Fixes `Docs.clear_docs` calling `texts_index.clear`

### DIFF
--- a/paperqa/llms.py
+++ b/paperqa/llms.py
@@ -775,7 +775,11 @@ class VectorStore(BaseModel, ABC):
     model_config = ConfigDict(extra="forbid")
 
     # can be tuned for different tasks
-    mmr_lambda: float = Field(default=0.9)
+    mmr_lambda: float = Field(
+        default=1.0,
+        ge=0.0,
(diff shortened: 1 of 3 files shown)


paperqa/llms.py
```diff
@@ -775,7 +775,11 @@ class VectorStore(BaseModel, ABC):
     model_config = ConfigDict(extra="forbid")
 
     # can be tuned for different tasks
-    mmr_lambda: float = Field(default=0.9)
+    mmr_lambda: float = Field(
+        default=1.0,
+        ge=0.0,
```
jamesbraza, line 780: I added this validation of `mmr_lambda >= 0` to be comprehensive

whitead (approved):
cool

PR #707, opened 2024-11-19
Adding filters to paper-qa Docs

whitead:
This adds a new filter mechanism to exclude papers from the `Docs` object via settings. 

For example, to exclude a specific DOI

```py
settings = Settings()
settings.parsing.doc_filters = [{"!doi": "xxxx/xxxxxx"}]
```

Or to only consider years 2020 and 2018:
```py
settings.parsing.doc_filters = [
    {"year": "2020"},
    {"year": "2018"}]
```

Description:

Optional filters to only allow documents that match this filter. This is a dictionary where the keys are the fields from DocDetails or Docs to filter on, and the values are the values to filter for. to invert filter, prefix the key with a '!'. If the key is not found, by default the Doc is rejected. To change this behavior, prefix the key with a '?' to allow the Doc to pass if the key is not found. For example, {'!title': 'bad title', '?year': '2022'} would only allow Docs with a title that is not 'bad title' and a year of 2022 or no year at all.

### DIFF
--- a/paperqa/settings.py
+++ b/paperqa/settings.py
@@ -196,6 +196,17 @@ class ParsingSettings(BaseModel):
         ),
     )
     chunking_algorithm: ChunkingOptions = ChunkingOptions.SIMPLE_OVERLAP
+    doc_filters: list[dict] | None = Field(
+        default=None,
+        description=(
+            "Optional filters to only allow documents that match this filter. This is a dictionary where the keys"
+            " are the fields from DocDetails or Docs to filter on, and the values are the values to filter for."
+            " to invert filter, prefix the key with a '!'. If the key is not found, by default the Doc is rejected."
+            " To change this behavior, prefix the key with a '?' to allow the Doc to pass if the key is not found."
+            " For example, {'!title': 'bad title', '?year': '2022'} would only allow Docs with a title that is not"
+            " 'bad title' and a year of 2022 or no year at all."
+        ),
+    )
 
     def chunk_type(self, chunking_selection: ChunkingOptions | None = None) -> str:
         """Future chunking implementations (i.e. by section) will get an elif clause here."""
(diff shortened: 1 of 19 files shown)


paperqa/settings.py
```diff
@@ -196,6 +196,17 @@ class ParsingSettings(BaseModel):
         ),
     )
     chunking_algorithm: ChunkingOptions = ChunkingOptions.SIMPLE_OVERLAP
+    doc_filters: list[dict] | None = Field(
```
jamesbraza, line 199: More of an aside, but for stuff like this I have seen people use Django REST: https://github.com/encode/django-rest-framework/blob/3.15.1/rest_framework/filters.py#L68-L73

It basically lets you query the DB for any model using URL filters, as opposed to us DIY'ing it behind FastAPI endpoints

More just sharing a way one can do this without DIY'ing filtering logic

tests/test_configs.py
```diff
@@ -90,3 +91,95 @@ def test_o1_requires_temp_equals_1() -> None:
...
+                    "year": 2020,
+                },
+                "filter_criteria": {"?volume": "10", "!title": "Another Paper"},
+                "expected_result": True,
+            },
+            id="DocDetails-relaxed-missing-volume",
+        ),
+    ],
+)
+def test_matches_filter_criteria(test_case):
```
jamesbraza, line 179: Why not have `doc_class` and others be a separate field outside of the `dict`?

So the first argument to `pytest.mark.parameterize` would be `("test_class", "doc_data", ""filter_criteria", "expected_result")`, and then:

```python
def test_matches_filter_criteria(test_class, doc_data, filter_criteria, expected_result) -> None:
```

paperqa/settings.py
whitead: Agree - would be nice to use something built in, but those look like they're slightly different meanings. We have a lot of optional/possibly missing params. But thanks for sharing - interesting to see!

jamesbraza (approved):
Nice

Issue #693, opened 2024-11-15
Evaluation bug when answer is not in option list


sidnarayanan:
We have observed two incorrect edge cases when the evaluation language model is provided an answer not in the options list

Here is one example:
```
INPUT:
 Extract the single letter answer from the following question and answer

Q: What is the meaning of life?

Options:
A) 11
B) cheesecake
C) -84
D) Insufficient information to answer this question
E) 42

14

Single Letter Answer:
OUTPUT:
 The single letter answer, based on the popular cultural reference from "The Hitchhiker's Guide to the Galaxy" where the answer to the ultimate question of life, the universe, and everything is humorously given as 42, is:

E
```
The LM ignores `14` (the supplied answer) and tries to answer directly. The response then breaks our single-letter-extraction regex, which returns `T`. This ultimately leads to us marking the answer as `INCORRECT` (as expected), but for the wrong reason.

Here is another example:
```
Q: What is my office's zip code?

Options:
A) 94107
B) -8
C) 94106
D) cheesecake
E) Insufficient information to answer this question

the answer is 14004

Single Letter Answer:
OUTPUT:
 To find the single letter answer, we need to determine which option corresponds to the answer "14004." Since none of the options directly match "14004," we need to consider if there's another way to interpret the answer.

Looking at the options:
A) 94107
B) -8
C) 94106
D) cheesecake
E) Insufficient information to answer this question

Since "14004" is a zip code and none of the provided zip codes match it, and since the question is about a zip code but the answer provided ("14004") isn't listed among the options, the most logical choice is:
E) Insufficient information to answer this question

Therefore, the single letter answer is:
E
```

Again, the supplied answer is ignored and another one is chosen (E). The longform response is parsed as `T`, and so we mark this as `INCORRECT`, not `UNSURE`.

jamesbraza:
To reproduce these (though `options` ordering isn't exactly the same):

```pycon
In [1]: from paperqa.litqa import LitQAEvaluation

In [2]: qa_prompt, eval_fn = LitQAEvaluation.from_question(
   ...:     ideal="42", distractors=["-84", "11", "cheesecake"],
   ...:     question="What is the meaning of life?", seed=0
   ...: )

In [3]: await eval_fn("14")
Out[3]: <LitQAEvaluation.CORRECT: 0>

In [4]: qa_prompt2, eval_fn2 = LitQAEvaluation.from_question(
   ...:     ideal="94107", distractors=["-8", "94106", "cheesecake"],
[...]

PR #708, opened 2024-11-19
Fixing `LitQAEvaluation` bugs: incorrect reward indices, not using LLM's native knowledge

jamesbraza:
This PR:
- Fixes both issues identified in https://github.com/Future-House/paper-qa/issues/693
    1. Using LLM's innate knowledge, which this PR fixes by updating `EVAL_PROMPT_TEMPLATE`
    2. Incorrectly handling options beyond the specified options, which this PR fixes by declaring those as "incorrect"
- Fixes bad `discounted_returns` logic and incorrect reward indices
    - Previously, incorrect answers were given 0.1 reward, and unsure answers were given -1.0 reward
    - Adds test coverage of this, both upon `discounted_returns` and the `TaskDataset`
- Makes the input `distractors` be a `Sequence` to avoid in-place edits, further robust-ification after https://github.com/Future-House/paper-qa/pull/694

Closes https://github.com/Future-House/paper-qa/issues/693


paperqa/litqa.py
```diff
@@ -77,8 +98,8 @@ class LitQAEvaluation(IntEnum):
     """Possible evaluation results for a LitQA question."""
 
     CORRECT = 0
-    INCORRECT = 1
-    UNSURE = 2
+    INCORRECT = 2
+    UNSURE = 1
```
sidnarayanan, line 102: Why rely on indexing instead of e.g. using a mapping?

paperqa/prompts.py
```diff
@@ -99,8 +99,8 @@
 QA_PROMPT_TEMPLATE = "Q: {question}\n\nOptions:\n{options}"
 EVAL_PROMPT_TEMPLATE = (
     "Extract the single letter answer from the following question and answer"
-    "\n\n{qa_prompt}"
-    "\n\n{qa_answer}"
+    "\n\nQuestion: {qa_prompt}"
+    "\n\nAnswer: {qa_answer}"
     "\n\nSingle Letter Answer:"
```
sidnarayanan, line 104: I'd suggest something like:
```
"Given the following question and a proposed answer to the question, return the single-letter choice that matches the proposed answer."
"\n\nQuestion: ..."
"\n\nProposed Answer: ..."
```

If I didn't know the context of this method, as a human, I'd find the original prompt unclear.

paperqa/litqa.py
jamesbraza: Nice suggestion, done 👌

paperqa/litqa.py
```diff
@@ -112,6 +138,14 @@ def extract_answer(answer: str) -> str:
             return answer.split()[0][0].upper()
 
         result = extract_answer(text)
+        if (
+            total_options is not None
+            and ord(result[0]) - _CAPITAL_A_INDEX + 1 > total_options
+        ):
+            # The result extracted was not in the options
```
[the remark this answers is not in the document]
jamesbraza: Let's make a test case for this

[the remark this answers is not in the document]
jamesbraza: I added a test case for this. TL;DR, we correct state that answer as `LitQAEvaluation.INCORRECT`, but for the wrong reasons. This is a separate issue, so I made https://github.com/Future-House/paper-qa/issues/709 for this. It's beyond this PR

mskarlin (approved):

paperqa/prompts.py
jamesbraza: Yeah it's a nice suggestion, but making this change breaks two of our test cases. Let's save this for another PR

PR #714, opened 2024-11-20
Fixes `gather_with_concurrency` typing

maykcaldas:
Changes the type hint of `gather_with_concurrency` from

`gather_with_concurrency(n: int, coros: list[Coroutine]) -> list[Any]:`

to

```
T = TypeVar("T")
async def gather_with_concurrency(n: int, coros: Iterable[Awaitable[T]]) -> list[T]:
```

~Also pinned pydantic version to <2.10 until they release the v2.10 bugfix.~


paperqa/docs.py
```diff
@@ -650,7 +650,7 @@ async def aget_evidence(
         for _, llm_result in results:
             session.add_tokens(llm_result)
 
-        session.contexts += [r for r, _ in results if r is not None]
+        session.contexts += [r for r, _ in results]
```
maykcaldas, line 653: This is being discussed in #687. However, the typing change didn't fix the mypy error saying that `r is not None` is a always True statement

paperqa/agents/models.py
```diff
@@ -87,7 +87,7 @@ def set_docs_name(self, docs_name: str) -> None:
 class AnswerResponse(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
-    session: PQASession = Field(alias="answer")
+    session: PQASession = Field(alias="session")
```
maykcaldas, line 90: @jamesbraza , this attribute name was changed from `answer` to `session` in  #713 . 
Is it ok to adjust the alias accordingly? 
This solves an error saying that `PQASession` doesn't have a `session` attribute, which I understood was because of the alias.

jamesbraza: So we need the `alias="answer"` for backwards compatibility, thus we can't make this change

jamesbraza: If we had `session: ... = Field(alias="session")`, that would make the `alias` redundant to the actual name

maykcaldas: That's what I thought. But mypy kept giving me this error, saying that there was no `session` attribute. Even vsCode UI was showing the documentation (when you hover over a variable) for `PQSession` with `answer` instead of `session`.

Can it maybe be a cache problem? 
I changed it back to `alias="answer"`, cleaned the pre-commit and will run it again.

paperqa/docs.py
whitead: Cannot do this I think. You can get a `None` back I believe

paperqa/types.py
```diff
@@ -73,7 +77,7 @@ class LLMResult(BaseModel):
 
     id: UUID = Field(default_factory=uuid4)
     session_id: UUID | None = Field(
-        default_factory=cvar_session_id.get,
```
jamesbraza, line 76: I looked into this, it's a `pydantic` issue: https://github.com/pydantic/pydantic/issues/10950

Can you just do a `# type: ignore[arg-type]` here, and link that issue

jamesbraza (commented):
@maykcaldas looks like `pydantic==2.10.1` is out, can you rebase or merge main for the updated `uv.lock`

.pre-commit-config.yaml
```diff
@@ -96,7 +96,7 @@ repos:
           - pybtex
           - numpy
           - pandas-stubs
-          - pydantic~=2.0 # Match pyproject.toml
```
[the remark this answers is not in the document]
maykcaldas: Oh, removed it by mistake when I deleted the comment about removing the pin to <2.10. Added `#Match pyproject.toml` back

jamesbraza (approved):
LGTM from my side. Looks like Andrew has an outstanding comment

Issue #710, opened 2024-11-19
maybe_is_text() discards valid text due to spaces from titles and tables


loesinghaus:
I noticed that the maybe_is_text() check discards quite a few perfectly valid and well-parsed publications. The issue is that it checks the entropy of the first text chunk of a document. Document parsing by pymupdf can introduce a lot of spaces, especially if the first few pages contain a title page, tables, or something similar (which they very often do, especially for books). Might be better to average across text chunks in the middle of the document.

Alternatively, checking the entropy of the text without spaces fixed it for my pdfs:

def maybe_is_text(s: str, thresh: float = 2.5) -> bool:
    if not s:
        return False
    # Calculate the entropy of the string
    entropy = 0.0
    s_wo_spaces = s.replace(" ", "")
    for c in string.printable:
        p = s_wo_spaces.count(c) / len(s_wo_spaces)
        if p > 0:
            entropy += -p * math.log2(p)

    # Check if the entropy is within a reasonable range for text
    return entropy > thresh

jamesbraza:
I like what you're thinking, feel free to make a PR and expand the `test_maybe_is_text` in `tests/test_paperqa.py`

PR #717, opened 2024-11-24
Alternative maybe is text

loesinghaus:
Changed maybe_is_text() to calculate the entropy without spaces to avoid discarding of valid papers due to title pages or tables. Also use the first five text chunks instead of just the first text chunk in case of other potential issues peculiar to title pages.

Closes https://github.com/Future-House/paper-qa/issues/710


paperqa/utils.py
```diff
@@ -52,10 +52,12 @@ def name_in_text(name: str, text: str) -> bool:
 def maybe_is_text(s: str, thresh: float = 2.5) -> bool:
     if not s:
         return False
+
     # Calculate the entropy of the string
     entropy = 0.0
+    s_wo_spaces = s.replace(" ", "")
```
jamesbraza, line 58: Can you update the comment two lines up to mention the heuristic for calculating entropy?

paperqa/docs.py
```diff
@@ -402,7 +402,7 @@ async def aadd(  # noqa: PLR0912
             or len(texts[0].text) < 10  # noqa: PLR2004
             or (
                 not parse_config.disable_doc_valid_check
-                and not maybe_is_text(texts[0].text)
+                and not maybe_is_text("".join(text.text for text in texts[:5]))
```
jamesbraza, line 405: Can you add a comment here saying why `:5`?

jamesbraza (commented):
Looking good, thank you!

paperqa/docs.py
loesinghaus: Done.

paperqa/utils.py
loesinghaus: I added a docstring with the reasoning for entropy calculation and leaving out spaces.

jamesbraza (approved):
LGTM

loesinghaus:
Alright :) Running the tests on my setup was also sort of difficult because they tend to trip my rate limits (need to spend more money on OpenAI, I guess).

jamesbraza:
I am not sure why [the CI run](https://github.com/Future-House/paper-qa/actions/runs/12127798131/job/33813739247?pr=717) is seeing:

```none
httpx.LocalProtocolError: Illegal header value b'Bearer '
```

Ugh our contributor developer experience is trash.

Anyways, tests pass locally on my Mac, so I am going to merge

loesinghaus:
No trouble, thanks for merging this one.

Issue #709, opened 2024-11-19
Failures to correct evaluate non-single letter answers


jamesbraza:
For the `qa_prompt`:

> Q: What is my office's zip code?
> 
> Options:
> A) 94107
> B) 94106
> C) cheesecake
> D) Insufficient information to answer this question
> E) -8

And the `qa_answer` (where correct is A):

> the answer is 94106 or 94107

The LLM evaluation says:

> The single letter answer is B.

Which gets converted to `"T"`, and declares it as `LitQAEvaluation.INCORRECT`.

Even though `LitQAEvaluation.INCORRECT` is actually the right output, we got it right here for the wrong reasons.

Issue #723, opened 2024-11-26
Empty session.answer for exhaustive search without an answer 


mskarlin:
The new complete tool allows the agent to call it quits after trying many search iterations. This is great but if it finds no relevant evidence, the GenerateAnswer tool won't be called either, so the session has nothing populated for the answer. We should have some kind of sentinel auto-injected here that can be displayed as output rather than just an empty string for this case.

jamesbraza:
Yeah to add an example:

```python
import asyncio

from paperqa.litqa import LitQAEvaluation

qa_prompt2, eval_fn2 = LitQAEvaluation.from_question(
    ideal="circular dichroism",
    distractors=["cryo EM", "x-ray crystallography", "NMR"],
    question=(
        "What method was used to demonstrate that the enzyme PafA is stable after"
        " incubation with 4M urea for 14 days?"
    ),
)
print(asyncio.run(eval_fn2("")))  # correct
```

I am actually working on this right now

The issue is without an answer, the LLM begins to pull on its innate knowledge

PR #724, opened 2024-11-26
Fixed `EVAL_PROMPT_TEMPLATE` to handle empty string or multiple match answers

jamesbraza:
This PR moves the evaluation prompt template to handle
- Empty string, which relates to https://github.com/Future-House/paper-qa/issues/723
- Multiple matches, which fixes https://github.com/Future-House/paper-qa/issues/709

### DIFF
--- a/paperqa/prompts.py
+++ b/paperqa/prompts.py
@@ -98,9 +98,12 @@ env_reset_prompt = (
 # Prompt templates for use with LitQA
 QA_PROMPT_TEMPLATE = "Q: {question}\n\nOptions:\n{options}"
 EVAL_PROMPT_TEMPLATE = (
-    "Extract the single letter answer from the following question and answer"
+    "Given the following question and a proposed answer to the question, return the"
+    " single-letter choice in the question that matches the proposed answer."
+    " If the proposed answer is blank or an empty string,"
+    " or multiple options are matched, respond with '0'."
     "\n\nQuestion: {qa_prompt}"
-    "\n\nAnswer: {qa_answer}"
+    "\n\nProposed Answer: {qa_answer}"
     "\n\nSingle Letter Answer:"
 )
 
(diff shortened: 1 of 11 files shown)


sidnarayanan (commented):
How do the benchmarks move with this change?

paperqa/prompts.py
```diff
@@ -98,9 +98,12 @@
 # Prompt templates for use with LitQA
 QA_PROMPT_TEMPLATE = "Q: {question}\n\nOptions:\n{options}"
 EVAL_PROMPT_TEMPLATE = (
-    "Extract the single letter answer from the following question and answer"
```
mskarlin, line 101: I think this is a solid iteration of the eval prompt template to address #709.

But I don't think this addresses [Empty session.answer for exhaustive search without an answer #723](https://github.com/Future-House/paper-qa/issues/723)

We still need the ability to inform the user of what happened at this step -- they just get a blank answer object currently. If the "user" is the eval script (as here) then this takes care of it.

mskarlin: Yea, exactly -- after the runner finishes, we need some sentinel to be added. The blank isn't intuitive enough for users to understand (especially via command line or UI). Can def be another PR.

mskarlin: In cases like this eval though -- the sentinel will create a need for another prompt, so we could make the sentinel a module level variable and check for it in the eval method.

jamesbraza:
> How do the benchmarks move with this change?

I am going to be running this shortly. With the empty answers, previously sometimes they got marked as correct, sometimes incorrect, and sometimes unsure. I am not sure which is more frequent, but that will likely dictate the impact on benchmark performance.

Fwiw our mini test suite's benchmark has grown here.

jamesbraza:
@sidnarayanan I will circle back with the impact of evaluation, going to merge this for now

jamesbraza:
Okay, including all of v5.6, it seems:

- Reward: mean drops by 0.05 to 0.18, standard deviation increases by 0.06 to 0.11
- Accuracy: mean drops by about 0.2% to 5.2%, standard deviation increases 0.8% to 4%

So in general, performance drops a bit and variability increases. So the evaluation flaws in v5.0 to v5.5 were pumping up the numbers a bit

PR #725, opened 2024-11-26
Unpins pydantic >2.10.2 requirement, removes TYPE_CHECKING

nadolskit:


tests/cassettes/test_crossref_retraction_status.yaml
nadolskit, line 1: unrelated to the change, but re-generating this cassette resolved my UT issues

tests/test_agents.py
```diff
@@ -464,7 +464,7 @@ async def gen_answer(self, state) -> str:  # noqa: ARG001
 @pytest.mark.flaky(reruns=3, only_rerun=["AssertionError", "EmptyDocsError"])
 @pytest.mark.asyncio
 async def test_agent_sharing_state(
-    agent_test_settings: Settings, subtests: SubTests, callback_type: str
+    agent_test_settings: Settings, subtests: SubTests, callback_type: str | None
```
nadolskit, line 467: added as this was an unresolved comment from James on https://github.com/Future-House/paper-qa/pull/590

pyproject.toml
```diff
@@ -8,7 +8,7 @@ dev = [
     "mypy>=1.8",  # Pin for mutable-override
     "paper-qa[datasets,ldp,typing,zotero,local]",
     "pre-commit>=3.4",  # Pin to keep recent
-    "pydantic~=2.0,>=2.10.1,<2.10.2",  # Pin 2.10 for typing breaks, downpin 2.10.2 until we fix AgentSettings.callbacks
+    "pydantic~=2.0",
```
jamesbraza, line 11: We still want `>=2.10.1` for the typing breaks

Also, can you propagate to `.pre-commit-config.yaml`

tests/test_agents.py
jamesbraza: Thanks!

paperqa/settings.py
```diff
@@ -58,8 +58,7 @@
 from paperqa.utils import hexdigest, pqa_directory
 from paperqa.version import __version__
 
-if TYPE_CHECKING:
-    from .agents.env import EnvironmentState
+EnvironmentState = TypeVar("EnvironmentState")
```
jamesbraza, line 61: Why does this fix work? It seems like we'd need to make `AgentSettings` a `Generic[EnvironmentState]`

nadolskit: importing EnvironmentState introduces a circular dependency which breaks things.
Unless I copy that class into `settings.py`, I don't believe `Generic[EnvironmentState]` will work.

When I added the type_checking originally, that's what I was aiming to resolve.

jamesbraza: Yeah we need to bite the bullet and resolve the circular dependency. I don't think this current fix here involving a `TypeVar` is actually valid.

I think we may need to move `EnvironmentState` + `make_status` to be in `settings.py`, and import them into `tools.py`

nadolskit: gotcha. on it!

paperqa/settings.py
```diff
@@ -505,7 +502,7 @@ class AgentSettings(BaseModel):
     )
     index: IndexSettings = Field(default_factory=IndexSettings)
 
-    callbacks: Mapping[str, list[Callable[["EnvironmentState"], Any]]] = Field(
+    callbacks: Mapping[str, list[Callable[..., Any]]] = Field(
```
jamesbraza, line 505:
```suggestion
    callbacks: Mapping[str, list[Callable[[Any], Any]]] = Field(
```

Let's do `Any` over `...` as the callbacks still take a single argument

And can you add a TODO comment saying to move from `Any` to `EnvironmentState`

jamesbraza (approved):
LGTM, thanks Tyler

nadolskit:
looks like failures from <title>api.openai.com | 502: Bad gateway</title>
merging -- unrelated to this change

jamesbraza:
Tests pass locally on Python 3.12, going to admin merge this to bypass OpenAI 429's in CI right now.

Need to keep the ball rolling, thanks for this @nadolskit

PR #726, opened 2024-11-27
Address missing `GenerateAnswer` in trajectories, no answers after `Complete` tools, and better history

mskarlin:
- Moved the tool_history into session -- so it can be externally accessed for tests + debugging.
- Added a history check for gen_answer in runner, it uses the history to ensure it's there.
- Added a sentinel into complete tool, this will allow trajectories where this tool is used outside of the runner to have something populated in the answer.


paperqa/agents/tools.py
```diff
@@ -373,6 +370,10 @@ async def complete(
         # TODO: eliminate race condition here if agent calls 2+ times in parallel
         # with opposite has_successful_answer values
         state.session.has_successful_answer = has_successful_answer
+
+        if not state.session.answer:
```
mskarlin, line 374: This technically won't be needed in a normal rollout with the extra logic above -- BUT if you're calling steps from some other orchestrator, then it covers your bases.

paperqa/agents/tools.py
```diff
@@ -76,7 +69,11 @@ def status(self) -> str:
 
     def record_action(self, action: ToolRequestMessage) -> None:
         self.session.add_tokens(action)
-        self.tool_history.append([tc.function.name for tc in action.tool_calls])
+        self.session.tool_history.append([tc.function.name for tc in action.tool_calls])
+
+    def query_tool_history(self, tool_name: str) -> list[int]:
+        """Return the step indices of the tool calls that used the given tool."""
+        return [i for i, tc in enumerate(self.session.tool_history) if tool_name in tc]
```
jamesbraza, line 76:
```suggestion
        return [i for i, tnames in enumerate(self.session.tool_history) if tool_name in tnames]
```

Let's use a name besides `tc` because that implies `ToolCall`

paperqa/agents/main.py
```diff
@@ -163,12 +163,17 @@ async def _run_with_timeout_failure(
     except Exception:
         logger.exception("Trajectory failed.")
         status = AgentStatus.FAIL
-    if status == AgentStatus.TRUNCATED:
+    if status == AgentStatus.TRUNCATED or not env.state.query_tool_history(
```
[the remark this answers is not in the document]
jamesbraza: > Now ldp or aviary rollouts

Just to clarify here, if directly using LDP's `RolloutManager` or a DIY rollout with aviary, this logic will not be hit. It's only used if one is using `paper-qa`'s `run_agent` runner

paperqa/agents/tools.py
```diff
@@ -76,7 +69,11 @@ def status(self) -> str:
 
     def record_action(self, action: ToolRequestMessage) -> None:
         self.session.add_tokens(action)
-        self.tool_history.append([tc.function.name for tc in action.tool_calls])
+        self.session.tool_history.append([tc.function.name for tc in action.tool_calls])
+
+    def query_tool_history(self, tool_name: str) -> list[int]:
```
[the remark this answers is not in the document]
jamesbraza: Also fwiw, I can see what you coded making sense too, just it loses a bit of information (it could be "called x times at steps a, b, c", but as is it's "called at least once at steps a, b, c")

paperqa/agents/tools.py
```diff
@@ -373,6 +370,10 @@ async def complete(
         # TODO: eliminate race condition here if agent calls 2+ times in parallel
         # with opposite has_successful_answer values
         state.session.has_successful_answer = has_successful_answer
+
+        if not state.session.answer:
+            state.session.answer = CANNOT_ANSWER_PHRASE
```
jamesbraza, line 375: I don't know if "I cannot answer" makes sense here -- it's not that it could not, it's that it did not

Maybe make another sentinel `DID_NOT_ANSWER_PHRASE = "No answer given."`

paperqa/agents/tools.py
jamesbraza, line 74: Can you add a test of this in one of the `test_agents`

paperqa/agents/tools.py
jamesbraza, line 374: Can you make test assertions for this in `test_agent_sharing_state`? You can add a `complete` tool call right at the start of the test

paperqa/agents/tools.py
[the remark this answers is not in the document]
mskarlin: The way it's implemented here makes more sense to me, it's the step number that a tool was called-- we don't have another use for this, I thought it could be something useful down the line. 

Since that's probably YAGNI territory 😄  -- I can just turn it into a straight up boolean because we likely don't need anything else. I'm just checking if it's in the history anywhere.

paperqa/agents/tools.py
mskarlin: agreed -- I was trying to make this as simple as possible, but it's better to add some extra stuff.

paperqa/agents/tools.py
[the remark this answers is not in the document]
jamesbraza: Ha you went the opposite direction, I see and YAGNIIIII

paperqa/agents/tools.py
```diff
@@ -357,6 +354,8 @@ class Complete(NamedTool):
         r" \| " + EnvironmentState.STATUS_SEARCH_REGEX_PATTERN
     )
 
+    NO_ANSWER_PHRASE: ClassVar[str] = "No answer generated."
```
jamesbraza, line 357: Should we move this to `PQASession` since that's where it will be used most likely? Imo housing this in the `Complete` tool makes it less accessible (since `Complete` tool is sort of an agent implementation detail)

jamesbraza (approved):
LGTM, and a Thanksgiving approval

mskarlin:
Note I'm merging w.o CI passing because we seem to still be getting 502s from openAI

PR #730, opened 2024-11-28
feat: Qdrant support

Anush008:
## Description

This PR adds support for Qdrant - https://qdrant.tech to be used an external database for vector search.

Qdrant can be run with :

```bash
docker run -p 6333:6333 qdrant/qdrant
```

A dashboard will be accessible at http://localhost:6333/dashboard.

## Testing

I've Q&A tested `QdrantVectorStore` implementation externally.


paperqa/llms.py
```diff
@@ -909,6 +910,94 @@ async def similarity_search(
         )
 
 
+class QdrantVectorStore(VectorStore):
+    client: Any = Field(
+        default=None, description="Instance of qdrant_client.QdrantClient"
```
jamesbraza, line 915: Can you document what `None` value here means in the `description`

paperqa/llms.py
```diff
@@ -909,6 +910,94 @@ async def similarity_search(
...
+            )
+
+        if not self.client:
+            self.client = QdrantClient()
+
+        return self
+
+    def clear(self) -> None:
+        super().clear()
+        from qdrant_client import models
```
jamesbraza, line 936: Can we just do this import at the module-level, within a `try`-`except`:

```python
try:
    from qdrant_client import QdrantClient, models
except importError:
    QdrantClient = models = None
```

Then later in `QdrantVectorStore`'s `model_validator` we do something like:

```python
if QdrantClient is None:
    raise importError(...)  # With nice message mentioning the extra
```

pyproject.toml
```diff
@@ -80,6 +80,9 @@ ldp = [
 local = [
     "sentence-transformers",
 ]
+qdrant = [
+    "qdrant-client",
```
jamesbraza, line 84: Can you add this to the `mypy` section of `.pre-commit-config.yaml`

Anush008: Done with https://github.com/Future-House/paper-qa/pull/730/commits/d15f3f38160399bae3af07c3d28ce9743ab50e8c.

.github/workflows/tests.yml
```diff
@@ -49,7 +54,7 @@ jobs:
         with:
           enable-cache: true
       - run: uv python pin ${{ matrix.python-version }}
-      - run: uv sync --python-preference=only-managed
+      - run: uv sync --python-preference=only-managed --extra qdrant
```
jamesbraza, line 57: So people running just `uv sync` locally and running tests will get fails on the Qdrant tests.

We need to either:
- Add the `qdrant` extra to our `dev` dependencies in the repo root's `pyproject.toml`'s `dependency-groups`' `dev`
- `pytest.skip` the Qdrant test cases if `qdrant-client` isn't installed

Can you pick one? I vote the second bullet

tests/test_paperqa.py
```diff
@@ -621,14 +623,17 @@ def test_duplicate(stub_data_dir: Path) -> None:
     ), "Unique documents should be hashed as unique"
 
 
-def test_docs_with_custom_embedding(subtests: SubTests, stub_data_dir: Path) -> None:
+@pytest.mark.parametrize("vector_store", [NumpyVectorStore, QdrantVectorStore])
+def test_docs_with_custom_embedding(
```
jamesbraza, line 627: Running this test locally I get this:

```
TypeError: cannot pickle '_thread.RLock' object
```

In the `docs_deep_copy = deepcopy(docs)` line

Can you run tests locally and get them to pass?

.github/workflows/tests.yml
Anush008: I found 1 more appropriate. Since we've parameterised tests to use both Numpy and Qdrant, `pytest.skip` would skip both if `qdrant-client` isn't installed.

The tests now use an in-memory instance. There's no need for Docker to be running.

tests/test_paperqa.py
Anush008: `QdrantClient` cannot be deep copied. So I've removed the deep copy checks. Since they're not applicable to the how `QdrantVectorStore` works anyway.

.github/workflows/tests.yml
Anush008: The in-memory mode is a Python implementation that mimics all the APIs of the Rust based Qdrant server.
It's used for prototyping and CIs.

tests/test_paperqa.py
jamesbraza: We need those assertions as they provide coverage on `PaperQAEnvironment`/`GradablePaperQAEnvironment.__deepcopy__`

You'll need to comprehend how `deepcopy`ing of an environment containing a `QdrantVectorStore` would play out. Perhaps you need to implement `QdrantVectorStore.__deepcopy__`

pyproject.toml
```diff
@@ -19,6 +19,7 @@ dev = [
     "pytest-xdist",
     "pytest>=8",  # Pin to keep recent
     "python-dotenv",
+    "qdrant-client",
```
jamesbraza, line 22: No you need to do `paper-qa[datasets,ldp,qdrant,typing,zotero,local]` on line 9, don't duplicate the package

tests/test_paperqa.py
Anush008: > TypeError: cannot pickle '_thread.RLock' object

Should be resolved now.

Will need to figure how to make 

```py
assert docs.texts_index != docs_shallow_copy.texts_index
```
pass without a duplicate `_embeddings_matrix`.

pyproject.toml
Anush008: Great save. Thanks.

tests/test_paperqa.py
Anush008: https://github.com/Future-House/paper-qa/pull/730/commits/9836e09dec8c5ffa6c44c30e4d4754584503b961

Should do it.

Anush008:
Hey @jamesbraza. Could you please approve the CI?

Anush008:
Weird that the mailman pre-commit doesn't complain locally. I've tried to do a patch.

Anush008:
Alright. That's through.
I guess the OpenAI errors are unrelated.

paperqa/llms.py
```diff
@@ -903,12 +911,138 @@ async def similarity_search(
...
 
 
+class QdrantVectorStore(VectorStore):
+    client: Any = Field(
+        default=None,
+        description="Instance of `qdrant_client.QdrantClient`. Defaults to an in-memory instance.",
+    )
+    collection_name: str = Field(default_factory=lambda: f"paper-qa-{uuid.uuid4().hex}")
+    vector_name: str | None = Field(default=None)
+    _ids: list[str] = []
```
jamesbraza, line 928: This exposes to mutable default issues, can you use `Field(default_factory=list)`

Also, should this be a `set`/`Collection` as opposed to a `list`? Does order matter?

Why not use `UUID` directly, why make it `str`?

Lastly, can you make the name more verbose, saying what the IDs correspond with?

paperqa/llms.py
```diff
@@ -903,12 +911,138 @@ async def similarity_search(
         # we could use arg-partition here
         # but a lot of algorithms expect a sorted list
         sorted_indices = np.argsort(-similarity_scores)
+
```
jamesbraza, line 914:
```suggestion
```

Let's avoid random added newlines

jamesbraza (commented):
Few more comments, looking good so far

paperqa/llms.py
Anush008: See https://github.com/Future-House/paper-qa/pull/730/commits/8b59afdf8fba71a67d849b5fa21ee468dc652bdc.

> Also, should this be a set/Collection as opposed to a list? Does order matter?

It's a set now.

> This exposes to mutable default issues, can you use Field(default_factory=list)

Field names can't start with underscores as they're private attrs.

> Why not use UUID directly, why make it str?

The client only allows `Union[int, str]` for point Ids.

> Lastly, can you make the name more verbose, saying what the IDs correspond with?

Done

jamesbraza: Thanks for doing that 👍

Anush008:
I believe these are the same OpenAI failures.

jamesbraza (commented):
Hi @Anush008 it looks great. Can you merge or rebase atop `main` for https://github.com/Future-House/paper-qa/pull/752, and confirm if `QdrantVectorStore` needs any changes?

Anush008:
> confirm if QdrantVectorStore needs any changes?

Should be fine.

jamesbraza (approved):
Nice work @Anush008 , thanks for this

ThomasRochefortB:
@Anush008 I am trying to use the new QdrantVectorStore using the following:
```python
from paperqa import QdrantVectorStore, Docs
from qdrant_client import QdrantClient

client = QdrantClient(url="http://localhost:6333")

vectorstore = QdrantVectorStore(client=client)

docs = Docs(texts_index=vectorstore)
docs.add("testpaper.pdf")
```

The code runs successfully but I don't see any collections being created in the dashboard of my Qdrant local instance.
In the `docs.aadd_texts` operation I see the [following comment](https://github.com/Future-House/paper-qa/blob/0dc007b15c5e5ea8d9fda956ffdc1155cb87c505/paperqa/docs.py#L479C68-L480C1) 
Could it be why I don't see the Qdrant collection being added to the vectorstore?

ThomasRochefortB:
@Anush008 I managed to successfully create a docs object and push it to Qdrant using the following:
```python
from paperqa import QdrantVectorStore, Docs
from qdrant_client import QdrantClient
import nest_asyncio

nest_asyncio.apply()

client = QdrantClient(url="localhost", port=6333)
vectorstore = QdrantVectorStore(client=client,
                                collection_name="test-collection")

docs = Docs(texts_index=vectorstore)
docs.add("testpaper.pdf")
docs.texts_index.add_texts_and_embeddings(docs.texts)
```

My question now is: 
- Is there a clever way to rebuild the `Docs()` object from the QdrantVectorStore directly? There seems to be everything we need persisted in the Qdrant collection.

ThomasRochefortB:
@Anush008 I have created a `Docs.load_docs_from_qdrant` function to reconstruct the `Docs()` object from Qdrant in the following #776

PR #733, opened 2024-11-29
Created `LitQAEvaluation.calculate_accuracy_precision` utility

jamesbraza:
Upstreaming some benchmarking logic to `paper-qa`


paperqa/litqa.py
```diff
@@ -219,6 +219,28 @@ async def llm_from_answer(answer: PQASession | str) -> LitQAEvaluation:
 
         return qa_prompt, llm_from_answer
 
+    @classmethod
+    def calculate_accuracy_precision(
+        cls, evaluations: Sequence[Self | str]
+    ) -> tuple[float, float]:
+        """
+        Calculate accuracy and precision given the input evaluations.
+
```
whitead, line 228: These are not the common definitions of accuracy and precision, so can you add the defs here (i.e., accuracy = num correct / num questions, precision = num correct / (num questions - num cannot answer))

jamesbraza: Fwiw the `Returns` entry a few lines below documents this (not sure if you saw or not):

> Returns: Two-tuple of accuracy (num correct) / (total num) and precision (num correct) / ((num correct) + (num incorrect)).

Note for precision's denominator, I said `(num correct) + (num incorrect)`, and you said `(num questions) - (num cannot answer)`. I like your definition more because it seems more grounded, so I updated both the summary and `Returns` wording:

> Calculate LitQA-specific accuracy and precision metrics upon evaluations.
>
> Returns: Two-tuple of accuracy = (num correct) / (num questions) and precision (num correct) / ((num questions) - (num unsure)).

Thanks for the comment

whitead (approved):

Issue #667, opened 2024-11-05
Paper QA Settings for AWS Bedrock


istranic:
Hi, do you have any guidance or examples for the `Settings` parameter in order to use paper QA with entirely AWS Bedrock models. I'm able to access by Bedrock's LLM and embedding models using `boto3` and `litellm`, but I can't figure out how to make it work in Paper QA Python API, and my errors are very generic and non-specific.

jamesbraza:
Hi @istranic can you post your stack trace(s)?

Also, check these `litellm` docs on how to use Bedrock, we are just passing through to them: https://docs.litellm.ai/docs/providers/bedrock

istranic:
Hi @jamesbraza I made progress and solved some of the issues. I can now manually add and query docs with all models hosted locally.


```
litellm_embedding_model = LiteLLMEmbeddingModel(
    name="openai/sentence-transformers/all-mpnet-base-v2",
    config=embedding_config,
    
## Add Docs
  await docs.aadd(doc, 
                  settings=settings,
                  citation=doc, 
                  docname=doc,
                  embedding_model=litellm_embedding_model)


## Query
answer = await docs.aquery(
    "What is the Nonlinearity Error and why does it matter?",
    settings=settings,
    embedding_model=litellm_embedding_model
)
```


However, when I add docs using the method above, it's not doing any of the advanced parsing, correct? In order to benefit from advanced parsing, I need to use something like:

```
answer = await agent_query(
    QueryRequest(
        query="What manufacturing challenges are unique to bispecific antibodies?",
        settings=settings,
        embedding=litellm_embedding_model
    )
```
    
 However, this doesn't work because `embedding`, `embedding_model`, etc. are not recognized arguments. If I try to specify the embedding model in the `settings`, it doesn't work either, and I get errors related to missing Open AI keys. What's the correct approach for specifying the `litellm_embedding_model` in the `agent_query` API?

jamesbraza:
> it's not doing any of the advanced parsing

So I can understand, what do you mean by "advanced parsing"?

---

> However, this doesn't work because `embedding`, `embedding_model`, etc. are not recognized arguments

Yeah when using `agent_query` you need to specify a `QueryRequest` which contains `settings: Settings`, kind of like we do here: https://github.com/Future-House/paper-qa#ask-manually

And for setting the `embedding_model` in `Settings`, start by reading here: https://github.com/Future-House/paper-qa#changing-embedding-model

[...]

PR #740, opened 2024-12-03
Cleared up 'Adding Documents Manually' docs

jamesbraza:
Clearing up some confusion from https://github.com/Future-House/paper-qa/issues/667

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,13 @@ but you can also specify a `"fake"` agent to use a hard coded call path of searc
 
 ### Adding Documents Manually
 
-If you prefer fine grained control, and you wish to add objects to the docs object yourself (rather than using the search tool), then the previously existing `Docs` object interface can be used:
+Normally via agent execution, the agent invokes the search tool,
+which adds documents to the `Docs` object for you behind the scenes.
+However, if you prefer fine-grained control,
+you can directly interact with the `Docs` object.
+
+Note that manually adding and querying `Docs` does not impact performance.
+It just removes the automation associated with an agent picking the documents to add.
 
 ```python
 from paperqa import Docs, Settings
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -334,29 +334,37 @@ but you can also specify a `"fake"` agent to use a hard coded call path of searc
...
+Note that manually adding and querying `Docs` does not impact performance.
+It just removes the automation associated with an agent picking the documents to add.
 
 ```python
+from pathlib import Path
 from paperqa import Docs, Settings
 
 # valid extensions include .pdf, .txt, and .html
-doc_paths = ("myfile.pdf", "myotherfile.pdf")
+doc_paths = (Path("myfile.pdf"), Path("myotherfile.pdf"))
```
whitead, line 350: Can we just use plain strings? Like keeping README syntax as simple as possible

README.md
```diff
@@ -334,29 +334,37 @@ but you can also specify a `"fake"` agent to use a hard coded call path of searc
...
 
-answer = docs.query(
+# Query the Docs object to get an answer
+session = docs.query(
     "What manufacturing challenges are unique to bispecific antibodies?",
     settings=settings,
 )
-
-print(answer.formatted_answer)
+print(session.formatted_answer)
```
whitead, line 367: I think we can just do `print(session)`

whitead (approved):
some optional comments, thanks!

README.md
jamesbraza: The issue is `Docs.aadd` accepts a `Path` only. I was trying to avoid a `TypeError` in our sample code

The right long-term move is supporting `str | os.PathLike`

README.md
jamesbraza: Nice suggestion, done in https://github.com/Future-House/paper-qa/pull/741

PR #743, opened 2024-12-04
Support env states with custom status functions

mskarlin:
Users may need custom status messages upon extra tools -- this allows custom status functions to be build into the `EnvironmentState`.

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -5,8 +5,9 @@ import inspect
 import logging
 import re
 import sys
+from collections.abc import Callable
 from itertools import chain
-from typing import ClassVar, cast
+from typing import ClassVar, Self, cast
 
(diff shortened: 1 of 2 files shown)


paperqa/agents/tools.py
```diff
@@ -31,13 +32,38 @@ def make_status(
     )
 
 
+def default_status(state: "EnvironmentState") -> str:
```
jamesbraza, line 35: You'll need a `TYPE_CHECKING` block that imports `EnvironmentState` I think for this to work

mskarlin: `EnvironmentState` is in this file, so I think we're good here

jamesbraza: Yep i was wrong here, sorry for false positive comment

jamesbraza (approved):

Issue #747, opened 2024-12-04
Pydantic error when parsing context


Snikch63200:
Hello, sometimes i have this error while using paper-qa :

```
   File "C:\Users\XXXXXXXXXXXXX\Desktop\PaperQA\paper_qa_question_sync.py", line 129, in <module>

    my_query()
  File "C:\Users\XXXXXXXXXXXXX\Desktop\PaperQA\paper_qa_question_sync.py", line 107, in my_query
    answer = docs.query(
             ^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\docs.py", line 665, in query
    return get_loop().run_until_complete(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\docs.py", line 705, in aquery
    session = await self.aget_evidence(
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\docs.py", line 632, in aget_evidence
    results = await gather_with_concurrency(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\utils.py", line 114, in gather_with_concurrency
    return await asyncio.gather(*(sem_coro(c) for c in coros))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\utils.py", line 112, in sem_coro
    return await coro
           ^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\paperqa\core.py", line 106, in map_fxn_summary
    Context(
  File "C:\ProgramData\anaconda3\envs\PaperQA_env\Lib\site-packages\pydantic\main.py", line 212, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
[...]

PR #748, opened 2024-12-04
Moved JSON summary prompt to mention score is an integer

jamesbraza:
We keep seeing (both internally and in https://github.com/Future-House/paper-qa/issues/747) the JSON context answers with float scores (e.g. 7.5).

I realized the non-JSON summary prompt explicitly states "integer":

> provide an integer score from 1-10

But the JSON summary prompt does not:

> relevance of `summary` to answer question (out of 10)

This PR standardizes the two prompts to mention integer scores between 1-10.

It also cleans up the `extract_score` function a bit and adds docs to it

### DIFF
--- a/paperqa/prompts.py
+++ b/paperqa/prompts.py
@@ -78,7 +78,7 @@ Provide a summary of the relevant information that could help answer the questio
   "relevance_score": "..."
 }}
 
-where `summary` is relevant information from text - {summary_length} words and `relevance_score` is the relevance of `summary` to answer question (out of 10).
+where `summary` is relevant information from text - {summary_length} words and `relevance_score` is an integer score from 1-10 for the relevance of `summary` to answer question.
 """  # noqa: E501
 
 env_system_prompt = (
--- a/paperqa/prompts.py
+++ b/paperqa/prompts.py
@@ -78,7 +78,7 @@ Provide a summary of the relevant information that could help answer the questio
   "relevance_score": "..."
 }}
 
-where `summary` is relevant information from text - {summary_length} words and `relevance_score` is an integer score from 1-10 for the relevance of `summary` to answer question.
+where `summary` is relevant information from the text - {summary_length} words. `relevance_score` is an integer 1-10 for the relevance of `summary` to the question.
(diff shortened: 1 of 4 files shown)


paperqa/prompts.py
```diff
@@ -78,7 +78,7 @@
   "relevance_score": "..."
 }}
 
-where `summary` is relevant information from text - {summary_length} words and `relevance_score` is the relevance of `summary` to answer question (out of 10).
+where `summary` is relevant information from text - {summary_length} words and `relevance_score` is an integer score from 1-10 for the relevance of `summary` to answer question.
```
whitead, line 81:
```suggestion
where `summary` is relevant information from the text - {summary_length} words. `relevance_score` is an integer 1-10 for the relevance of `summary` to answer the question.
```

paperqa/utils.py
```diff
@@ -130,12 +130,19 @@ def strip_citations(text: str) -> str:
...
+    """
+    # check for N/A, NA, not applicable, not relevant
     last_line = text.split("\n")[-1]
-    if "N/A" in last_line or "n/a" in last_line or "NA" in last_line:
-        return 0
-    # check for not applicable, not relevant in summary
-    if "not applicable" in text.lower() or "not relevant" in text.lower():
+    if (
+        "n/a" in last_line.lower()
+        or "NA" in last_line
```
whitead, line 142: This looks like a bug. For example:

"The gene NAR is explicitly mentioned in the text, thus the relevance score is 10/10"

mskarlin (approved):

paperqa/utils.py
jamesbraza: Fwiw, the `"NA"` check was previously there before this PR, this change was just a refactor for readability.

I agree with your comment through, so I removed the `"NA"` check and added this comment:

```python
# Don't check for NA, as there can genes containing NA
```

paperqa/prompts.py
whitead: Just want to clean up grammar here, as long as we're editing

whitead (approved):

PR #752, opened 2024-12-05
Add partitioning func capabilities to allow doc-types-based embedding ranking

mskarlin:
When you have different flavors of docs (say texts of JSON/Code along with some PDF-parsed paper text), the embedding ranking won't be based on semantics alone. All of the texts that are JSON may cluster together, and while they could be highly relevant, they are sometimes embedded further away from a natural text query based on structure alone. Our summary step deals with this downstream of the embedding ranking, but some good texts will never have the opportunity to enter into the summary step in this case.

To deal with this I added an optional `partitioning_fn` input into the `Docs` object. This allows a user to make a function which will turn each `Text` object's `Doc` into an integer category. The similarity search in the retrieval step will then be split by each integer category, ranked independently, then interleaved into a final output. This way each "partition" gets a shot at being included in the summary step.


paperqa/docs.py
```diff
@@ -80,6 +80,11 @@ class Docs(BaseModel):
         default=PAPERQA_DIR, description="Path to save index", validate_default=True
     )
     deleted_dockeys: set[DocKey] = Field(default_factory=set)
+    partitioning_fn: Callable[[Doc], int] | None = Field(
+        default=None,
+        description="Function to partition documents during retrieval,"
```
jamesbraza, line 85: Can you start wording with "Optional function" to make it clear it's optional

paperqa/docs.py
```diff
@@ -80,6 +80,11 @@ class Docs(BaseModel):
         default=PAPERQA_DIR, description="Path to save index", validate_default=True
     )
     deleted_dockeys: set[DocKey] = Field(default_factory=set)
+    partitioning_fn: Callable[[Doc], int] | None = Field(
```
jamesbraza, line 83: Won't this break serialization, so we need `exclude=True`

Or am I wrong here somehow?

paperqa/llms.py
```diff
@@ -794,16 +798,36 @@ def add_texts_and_embeddings(self, texts: Iterable[Embeddable]) -> None:
 
     @abstractmethod
     async def similarity_search(
-        self, query: str, k: int, embedding_model: EmbeddingModel
+        self,
+        query: str,
+        k: int,
+        embedding_model: EmbeddingModel,
```
jamesbraza, line 804:
```suggestion
        embedding_model: EmbeddingModel
```

Just avoiding diff churn

paperqa/llms.py
```diff
@@ -875,12 +910,56 @@ def clear(self) -> None:
...
 
+    async def partitioned_similarity_search(
+        self,
+        query: str,
+        k: int,
+        embedding_model: EmbeddingModel,
+        partitioning_fn: Callable[[Doc], int],
+    ) -> tuple[Sequence[Embeddable], list[float]]:
+        scores = []
+        texts = []
```
jamesbraza, line 928: Can you type annotate these? It's nice to know if:
- `scores` is a `float`, `int`, or `str`
- `texts` are a `str` or `Text`

tests/test_paperqa.py
```diff
@@ -1199,3 +1199,68 @@ def test_answer_rename(recwarn) -> None:
...
+
+    docs = Docs(partitioning_fn=partitioning_fn)
+
+    assert isinstance(
+        docs.texts_index, NumpyVectorStore
+    ), "We want this test to cover NumpyVectorStore"
+
+    # add docs that we can use our partitioning function on
+    stub_doc = Doc(docname="stub", citation="stub", dockey="stub")
+    special_doc = Doc(docname="special", citation="special", dockey="special")
```
jamesbraza, line 1231: Can you name these better? Something like:

```python
positive_statements_doc = Doc(...)
negative_statements_doc = Doc(...)
```

I think this contrast makes the test easier to read than "stub" vs "special"

paperqa/docs.py
mskarlin: Yea it's prob safer to exclude, these should be pretty widely pickle-able, but if someone put a DB connection in their `partitioning_fn` then it wouldn't be. 

We could try to specify this with typing, but it's easier to just exclude.

jamesbraza: Down with whatever here

paperqa/llms.py
```diff
@@ -802,8 +806,25 @@ async def similarity_search(
...
+    async def partitioned_similarity_search(
+        self,
+        query: str,
+        k: int,
+        embedding_model: EmbeddingModel,
+        partitioning_fn: Callable[[Doc], int],
+    ) -> tuple[Sequence[Embeddable], list[float]]:
+        raise NotImplementedError(
+            "Need to use the partitioning_fn argument to "
+            "stratify the similarity search."
```
jamesbraza, line 818: I feel like it's weird to use this wording in the base class, it can be misleading. It seems like a combination of:
- An`Args:` description for `partitioning_fn`
- An `abstractmethod`'s `NotImplementedError`

Can we move to:
1. A docstring with `Args:` (especially since this is an `abstractmethod`)
2. A more intuitive message in the `NotImplementedError`

paperqa/llms.py
```diff
@@ -812,14 +833,24 @@ async def max_marginal_relevance_search(
             k: Number of results to return.
             fetch_k: Number of results to fetch from the vector store.
             embedding_model: model used to embed the query
+            partitioning_fn: optional function to partition the documents into
+                different groups, performing MMR within each group
```
jamesbraza, line 837:
```suggestion
                different groups, performing MMR within each group.
```

https://peps.python.org/pep-0257/#one-line-docstrings
> The docstring is a phrase ending in a period.

paperqa/llms.py
```diff
@@ -875,12 +907,56 @@ def clear(self) -> None:
...
+        scores: list[list[float]] = []
+        texts: list[Sequence[Embeddable]] = []
+
+        # duck typing for self.texts since they don't technically
+        # have to have .doc attributes
+        if self.texts and not hasattr(self.texts[0], "doc"):
+            logger.warning(
+                "Skipping partitioning: A partitioning_fn was provided "
+                "but the texts do not have a 'doc' attribute."
+            )
```
jamesbraza, line 933: Won't we hit this logged warning a ton?

To me this seems like a code smell for we should do something different

tests/test_paperqa.py
```diff
@@ -1199,3 +1199,71 @@ def test_answer_rename(recwarn) -> None:
...
+    # embedding rank by itself
+    def partition_by_citation(doc: Doc) -> int:
+        if "special" in doc.citation:
+            return 1
+        return 0
+
+    partitioning_fn = None
+
+    if use_partition:
+        partitioning_fn = partition_by_citation
```
jamesbraza, line 1221:
```suggestion
    partitioning_fn = partition_by_citation if use_partition else None
```

Let's do it in one line for our dear friend `mypy`

jamesbraza (approved):

paperqa/llms.py
mskarlin: I'm actually not sure how this would ever be hit - you'd need to build your own version of a `NumpyVectorStore`, and instantiate the objects in it outside of the PaperQA flow.

We could subclass the whole `NumpyVectorStore`, and change the texts type from `Embeddable` to `Texts`, but that would be the only change otherwise. So I relied on duck typing. 

To think about the scenario that would produce this warning -- you'd need to write a partitioning function which takes in `Doc` objects, and returns and `int`, then you build your own `NumpyVectorStore` and insert to it objects which don't have any `Doc` objects as attributes. That seems really fringe.

paperqa/docs.py
```diff
@@ -80,6 +80,12 @@ class Docs(BaseModel):
...
```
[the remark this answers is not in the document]
mskarlin: I thought that too-- but I think the serialization issue seems worse there since that's the most frequently serialized object we use. The `Settings` objects are often initialized from a validated JSON (which this attribute wouldn't work with) and the `Docs` are initialized via code (or a deserialization from a pickle), but `Docs` objects already aren't JSON compatible because of the VectorStore. 

What do you think?

Also could you elaborate on "better mapping between str and functions"? 

In practice, I'm modifying this attribute dynamically upon env initialization if a particular tool is selected, so we could tack it on to a settings object just while it's in the lifecycle of the environment, then remove it, or just exclude from serialization in the same way I'm doing here.

paperqa/llms.py
```diff
@@ -875,12 +907,56 @@ def clear(self) -> None:
...
         self.texts = []
         self._embeddings_matrix = None
+        self._texts_filter = None
 
     def add_texts_and_embeddings(self, texts: Iterable[Embeddable]) -> None:
         super().add_texts_and_embeddings(texts)
         self.texts.extend(texts)
         self._embeddings_matrix = np.array([t.embedding for t in self.texts])
 
+    async def partitioned_similarity_search(
```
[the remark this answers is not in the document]
mskarlin: yea you're right I've gotta bit the bullet and write some thorough tests here.

paperqa/llms.py
mskarlin: Actually thought more about this — I think I’ll just add as an arg to all the requisite functions— then we skip this issue entirely. I’ll need to add the funcs to the tool initializations, which I’ll think through more.

paperqa/docs.py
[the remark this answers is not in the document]
mskarlin: Actually thought more about this — I think I’ll just add as an arg to all the requisite functions— then we skip this issue entirely. I’ll need to add the funcs to the tool initializations, which I’ll think through more.

paperqa/llms.py
mskarlin: good call -- added a full docstring here.

paperqa/llms.py
mskarlin: Re-structured this function, so we don't need this duck typing anymore!

paperqa/docs.py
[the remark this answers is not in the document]
mskarlin: Ok all set now! No more stateful stuff.

paperqa/llms.py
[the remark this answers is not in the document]
mskarlin: Ok I added a test that looks for exact rankings - I think we're solid on this one.

PR #760, opened 2024-12-11
Made it possible to get answers from litqa evaluations

whitead:
I'm not super happy with the pattern, but I modified `LitQAEvaluation` so that it stuffs the extracted answer into the `StrEnum` value. This makes it so we can recover the actual answers at the end, to help with consensus sampling.


jamesbraza (approved):
I posed a better alternative in comments, approving because this does work

Issue #761, opened 2024-12-11
Reactor: Overly Complicated Question Eval


whitead:
The litqa evaluations have become overly complex. We should simplify the logic to use as much `aviary` code as possible and not overload the StrEnum so much.

Context:
1. https://github.com/Future-House/aviary/pull/148
2. https://github.com/Future-House/paper-qa/pull/760
3. https://github.com/Future-House/paper-qa/pull/760#discussion_r1880756716

PR #762, opened 2024-12-11
Added answer an ideal to gradeable environments

whitead:
Actually expose answer/ideal in Frame

### DIFF
--- a/paperqa/agents/task.py
+++ b/paperqa/agents/task.py
@@ -86,6 +86,8 @@ class GradablePaperQAEnvironment(PaperQAEnvironment):
         )
         self._evaluation_callback = evaluation_callback
         self._rewards = rewards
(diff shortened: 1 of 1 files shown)


sidnarayanan (approved):

paperqa/agents/task.py
```diff
@@ -86,6 +86,8 @@ def __init__(
         )
         self._evaluation_callback = evaluation_callback
         self._rewards = rewards
+        self.answer = ""
```
jamesbraza, line 89: Fwiw this is now in some way duplicated across both the `evaluation_from_answer: Callable` passed in, and these new instance attributes.

We should attempt to fix this duplication when fixing https://github.com/Future-House/paper-qa/issues/761

PR #770, opened 2024-12-18
Fixed type problems from llmclient

maykcaldas:


paperqa/clients/__init__.py
```diff
@@ -164,16 +164,14 @@ async def query(self, **kwargs) -> DocDetails | None:
...
-                            ),
+                            task.processor_queries(doc_details, session),
                         )
                     )
                     or None
                 )
 
             if doc_details:
-                doc_details = cast(DocDetails, doc_details)
+                # doc_details = cast(DocDetails, doc_details)
```
jamesbraza, line 174: Can we just delete this line?

tests/test_paperqa.py
```diff
@@ -683,7 +683,7 @@ def test_sparse_embedding(stub_data_dir: Path, vector_store: type[VectorStore])
         citation="WikiMedia Foundation, 2023, Accessed now",
         embedding_model=SparseEmbeddingModel(),
     )
-    assert any(docs.texts[0].embedding)
+    assert any(cast(list, docs.texts[0].embedding))
```
jamesbraza, line 686: Instead of a `cast`, in unit tests it's best to do `assert isinstance(x, list)`, can you do that here?

paperqa/types.py
```diff
@@ -54,9 +54,9 @@ def set_llm_session_ids(session_id: UUID):
 
 
 class Doc(Embeddable):
-    docname: str
+    docname: str | None = None
+    dockey: DocKey | None = None
```
jamesbraza, line 58: Why make these optional and defaulted to `None`?

I am not against it, but more just wondering why. Can you add a `description` stating what these should be and what `None` means

tests/test_paperqa.py
```diff
@@ -683,7 +683,7 @@ def test_sparse_embedding(stub_data_dir: Path, vector_store: type[VectorStore])
         citation="WikiMedia Foundation, 2023, Accessed now",
         embedding_model=SparseEmbeddingModel(),
     )
-    assert any(docs.texts[0].embedding)
```
jamesbraza, line 686: Can we keep the `any` assertions around? Why move to `list`?

If you think the assertion is the same behavior, can you add an assertion failure message, to help make what it's checking for clear?

```diff
--- assert any(docs.texts[0].embedding), "We require the embedding to be populated"
+++ assert isinstance(docs.texts[0].embedding, list), "We require the embedding to be populated"
```

paperqa/agents/helpers.py
```diff
@@ -92,11 +92,14 @@ def table_formatter(
         table.add_column("Title", style="cyan")
         table.add_column("File", style="magenta")
         for obj, filename in objects:
+            # NOTE: Text.doc is typed as a `Doc`. But it should be a `DocDetails`.
+            # TODO: Change the type of Text.doc to DocDetails | Doc.
```
jamesbraza, line 96: Can we resolve this TODO in this PR

paperqa/agents/helpers.py
```diff
@@ -92,11 +92,14 @@ def table_formatter(
...
+            doc = cast(DocDetails, cast(Docs, obj).texts[0].doc)
             try:
-                display_name = cast(DocDetails, cast(Docs, obj).texts[0].doc).title
+                display_name = cast(str, doc.title)
```
jamesbraza, line 99: I think this route is supposed to be for `Doc`, but `title` is only for `DocDetails`.

What do you think of:
1. Removing the above `cast`, so we allow for `DocDetails | Doc`
2. Having this be `doc.docname`, because it's all we can do for a `Doc`

Cc @mskarlin

paperqa/types.py
```diff
@@ -80,6 +80,10 @@ def matches_filter_criteria(self, filter_criteria: dict) -> bool:
 class Text(Embeddable):
     text: str
     name: str
+    # TODO: doc is often used as `DocDetails`.
```
jamesbraza, line 83: Let's resolve this TODO here, I think it should be `doc: DocDetails | Doc`

paperqa/types.py
jamesbraza: Can add to `DocDetails` (but not touch `Doc`)

```python
    AUTOPOPULATE_VALUE: ClassVar[str] = ""

    docname: str = AUTOPOPULATE_VALUE
    dockey: DocKey = AUTOPOPULATE_VALUE
```

paperqa/types.py
```diff
@@ -80,7 +80,34 @@ def matches_filter_criteria(self, filter_criteria: dict) -> bool:
...
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_doc(cls, values: Any) -> Any:
+        if not isinstance(values, dict):
+            # NOTE: If I don't know what is being passed, I don't change it
+            # This happens in some tests
+            return values
+
+        doc_data = values.get("doc")
+        if isinstance(doc_data, Doc | DocDetails):
```
jamesbraza, line 98:
```suggestion
        if isinstance(doc_data, Doc):
```

Since `DocDetails` is a `Doc`

paperqa/types.py
```diff
@@ -306,6 +333,11 @@ def reduce_content(self) -> str:
 class DocDetails(Doc):
     model_config = ConfigDict(validate_assignment=True)
 
+    AUTOPOPULATE_VALUE: ClassVar[str] = ""
+
+    docname: str = AUTOPOPULATE_VALUE
+    dockey: DocKey = AUTOPOPULATE_VALUE
+
```
jamesbraza, line 340:
```suggestion
```

tests/test_cli.py
```diff
@@ -80,5 +80,5 @@ def test_cli_can_build_and_search_index(
     result = search_query("XAI", index_name, settings)
     assert len(result) == 1
     assert isinstance(result[0][0], Docs)
-    assert result[0][0].docnames == {"Wellawatte"}
+    assert result[0][0].docnames == {"Wellawatte2024"}
```
jamesbraza, line 83:
```suggestion
    assert all(d.startswith("Wellawatte") for d in result[0][0].docnames}
```

More general way

paperqa/types.py
```diff
@@ -80,7 +80,34 @@ def matches_filter_criteria(self, filter_criteria: dict) -> bool:
...
+            return values
+
+        if doc_data:
+            maybe_is_docdetails = all(
+                k in doc_data for k in list(DocDetails.model_fields.keys())
+            )
+            maybe_is_doc = all(k in doc_data for k in list(Doc.model_fields.keys()))
+            if maybe_is_doc and not maybe_is_docdetails:
+                doc = Doc(**doc_data)
+                values["doc"] = doc
```
jamesbraza, line 109: Few comments here:

So a shorter and more intuitive way for these check is: `maybe_is_docdetails = set(doc_data).issubset(DocDetails.model_fields)`. However, regardless the check is flawed because it doesn't take into account typing, just key alignment.

What we should do is lean more into Pydantic
1. Add `ConfigDict(extra="forbid")` to `Doc`
2. Add `ConfigDict(extra="ignore")` to `DocDetails`
3. Then we can just attempt instantiation for these two, and use the `pydantic.ValidationError` to determine compatibility

paperqa/agents/helpers.py
maykcaldas: Removed these notes once we're solving this issue in this PR

paperqa/agents/helpers.py
maykcaldas: Looking at it again, I think this route should be for `DocDetails`. Then if `doc` has no title, it falls back to `formatted_citation`. 
I still agree that we can just remove the `cast` and follow a fuck typing approach trying to use `doc.title` and then using `doc.formatted_citation` if it fails.

paperqa/types.py
maykcaldas: Oh, yeah. Makes sense.

I added the `ConfigDict` to `Doc`. `DocDetails` already had it (`extra="ignore"` is default). This revealed some cases where a `DocDetails` was being passed to `Text`. So this is a good change to keep the correct typing for `Text.doc`.

Now, the validation tries to create a `Doc` with whatever is passed to `Text.doc`. If it fails, it tries to instantiate a `DocDetails`. Because `DocDetails` accepts extras, this instantiation shouldn't fail. But we're catching this error anyway.

paperqa/agents/helpers.py
jamesbraza: Then can you add a comment explaining this, and do it? Also, since `title` will be present, it will not have an `AttributeError` chance, so can you update

paperqa/types.py
```diff
@@ -53,6 +57,15 @@ class Doc(Embeddable):
     def __hash__(self) -> int:
         return hash((self.docname, self.dockey))
 
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_no_extra(cls, data: Any) -> Any:
+        if isinstance(data, dict):
+            # formatted_citation is serialized by model_dump
+            # but it's not an attribute
```
jamesbraza, line 65:
```suggestion
            # Discard computed_field
```

paperqa/types.py
```diff
@@ -80,7 +93,43 @@ def matches_filter_criteria(self, filter_criteria: dict) -> bool:
...
     name: str
-    doc: Doc
+    doc: DocDetails | Doc
+
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_doc(cls, values: Any) -> Any:
+        if not isinstance(values, dict):
+            # NOTE: If I don't know what is being passed, I don't change it
+            # This happens in some tests
```
jamesbraza, line 103: Can you update this comment? Let's not leave mysterious comments in the code.

Try debugging here while running the unit tests and figuring out exactly what values it is

paperqa/agents/helpers.py
```diff
@@ -92,11 +92,15 @@ def table_formatter(
         table.add_column("Title", style="cyan")
         table.add_column("File", style="magenta")
         for obj, filename in objects:
+            doc = cast(Docs, obj).texts[0].doc
```
jamesbraza, line 95: Is this right? Wouldn't it be `Doc`, not `Docs`?

paperqa/agents/helpers.py
jamesbraza: So there is still the `except AttributeError`, but we won't be able to hit this because `DocDetails` has a `title` attribute. Can you change to a boolean check on `title`?

paperqa/types.py
```diff
@@ -53,7 +55,7 @@ class Doc(Embeddable):
     def __hash__(self) -> int:
         return hash((self.docname, self.dockey))
 
-    @computed_field  # type: ignore[prop-decorator]
+    # @computed_field  # type: ignore[prop-decorator]
```
jamesbraza, line 58: Can you bring this back?

tests/test_agents.py
```diff
@@ -340,7 +340,7 @@ async def llm_model_call(*args, **kwargs):
         # NOTE: "required" will not lead to thoughts being emitted, it has to be "auto"
         # https://docs.anthropic.com/en/docs/build-with-claude/tool-use#chain-of-thought
         kwargs.pop("tool_choice", MultipleCompletionLLMModel.TOOL_CHOICE_REQUIRED)
-        return await orig_llm_model_call(*args, tool_choice="auto", **kwargs)  # type: ignore[misc]
+        return await orig_llm_model_call(*args, tool_choice="auto", **kwargs)  # type: ignore [misc]
```
jamesbraza, line 343: Extra space here

paperqa/types.py
```diff
@@ -594,6 +600,8 @@ def populate_bibtex_key_citation(  # noqa: PLR0912
     @model_validator(mode="before")
     @classmethod
     def validate_all_fields(cls, data: dict[str, Any]) -> dict[str, Any]:
+        if isinstance(data, Doc):  # type: ignore[unreachable]
```
jamesbraza, line 603: I think we should type hint `data: dict[str, Any] | Doc`, then it will resolve this `unreachable`

Also, can you add a unit test proving we don't manipulate an input `dict`?

We shouldn't be manipulating the input `dict` during validations

uv.lock
```diff
@@ -240,7 +240,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "sys_platform == 'win32'" },
+    { name = "colorama", marker = "platform_system == 'Windows'" },
```
jamesbraza, line 243: Can you revert the `uv.lock` changes in this PR?

maykcaldas: the pre-commit automation are adding them back 😅

paperqa/types.py
```diff
@@ -593,7 +606,9 @@ def populate_bibtex_key_citation(  # noqa: PLR0912
 
     @model_validator(mode="before")
     @classmethod
-    def validate_all_fields(cls, data: dict[str, Any]) -> dict[str, Any]:
+    def validate_all_fields(cls, data: dict[str, Any] | Doc) -> dict[str, Any]:
+        if isinstance(data, Doc):
+            data = data.model_dump()
```
jamesbraza, line 611: Do you mind reverting this change? I just debugged the test suite, this `Doc` entrance doesn't actually happen in practice

paperqa/agents/helpers.py
jamesbraza: This comment was incorrect, resolving it

paperqa/agents/helpers.py
jamesbraza: Resolved in https://github.com/Future-House/paper-qa/pull/813

paperqa/types.py
```diff
@@ -313,16 +312,18 @@ def reduce_content(self) -> str:
 
 
 class DocDetails(Doc):
-    model_config = ConfigDict(validate_assignment=True, extra="ignore")
+    model_config = ConfigDict(validate_assignment=True)
```
jamesbraza, line 315:
```suggestion
    model_config = ConfigDict(validate_assignment=True, extra="ignore")
```

Actually, let's just revert this `DocDetails.model_config` change

jamesbraza (approved):

jamesbraza (approved):
Also, huge battle here and some real deep bugs found, excellent work on this one!! 🚀

paperqa/types.py
maykcaldas: Isn't `extra="ignore"`default?

PR #777, opened 2024-12-23
Add clinical trials search tool

mskarlin:
This upstreams FutureHouse's clinical trials search tool to make it open source. The tool is not turned on by default, but I'm going to make more docs on how to use the tool as well. 

I also found many typing regressions -- I've added those back in where necessary.

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -151,6 +151,13 @@ def make_clinical_trial_status(
     )
 
 
+# SEE: https://regex101.com/r/L0L5MH/1
+CLINICAL_STATUS_SEARCH_REGEX_PATTERN: str = (
+    r"Status: Paper Count=(\d+) \| Relevant Papers=(\d+)(?:\s\|\sClinical Trial Count=(\d+)\s"
+    r"\|\sRelevant Clinical Trials=(\d+))?\s\|\sCurrent Evidence=(\d+)"
+)
+
+
 def clinical_trial_status(state: "EnvironmentState") -> str:
     return make_clinical_trial_status(
         total_paper_count=len(
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from copy import deepcopy
 from typing import Any, ClassVar, Self, cast
 
@@ -26,6 +25,7 @@ from paperqa.utils import get_year
 from .models import QueryRequest
 from .tools import (
     AVAILABLE_TOOL_NAME_TO_CLASS,
+    DEFAULT_TOOL_NAMES,
     ClinicalTrialsSearch,
     Complete,
     EnvironmentState,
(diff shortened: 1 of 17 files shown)


paperqa/agents/env.py
```diff
@@ -33,8 +39,20 @@
 
 POPULATE_FROM_SETTINGS = None
 
+DEFAULT_TOOL_NAMES: list[str] = [
```
jamesbraza, line 42: I like this, nice.

What do you think of moving this to be co-located with `AVAILABLE_TOOL_NAME_TO_CLASS` in `tools`?

paperqa/agents/tools.py
```diff
@@ -399,6 +400,253 @@ async def complete(
         return f"{'Success' if has_successful_answer else 'Unsure'} | {state.status}"
 
 
+class ClinicalTrialsSearch(NamedTool):
+    TOOL_FN_NAME = "clinical_trials_search"
+
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
```
jamesbraza, line 406: Why `arbitrary_types_allowed=True`? I don't think we need it (may be wrong here tho)

paperqa/agents/tools.py
```diff
@@ -399,6 +400,253 @@ async def complete(
...
+        Args:
+            question: Specific question to gather evidence for.
+            state: Current state.
+
+        Returns:
+            String describing gathered evidence and the current status.
+        """
+    )
+
+    async def clinical_trials_search(self, query: str, state: EnvironmentState):
```
jamesbraza, line 431:
```suggestion
    async def clinical_trials_search(self, query: str, state: EnvironmentState) -> str:
```

Even though we don't use return type now, it's still nice to have for IDE usage

paperqa/sources/clinical_trials.py
```diff
@@ -0,0 +1,321 @@
...
+        params={
+            "query.term": query,
+            "fields": SEARCH_API_FIELDS,
+            "pageSize": SEARCH_PAGE_SIZE,
+            "countTotal": "true",
+            "sort": "@relevance",
+        },
+    ) as response:
+        if response.status == MALFORMATTED_QUERY_STATUS:
+            # the 400s from clinicaltrials.gov are not JSON
```
jamesbraza, line 53: Do you mind moving this comment to live with `MALFORMATTED_QUERY_STATUS`, or renaming `MALFORMATTED_QUERY_STATUS` to be `CLINICAL_TRIALS_DIDNT_GIVE_JSON = 400`

paperqa/agents/tools.py
mskarlin: ahh nice catch -- we removed some attributes in upstreaming

paperqa/agents/tools.py
```diff
@@ -399,6 +400,253 @@ async def complete(
...
+    TOOL_FN_NAME = "clinical_trials_search"
+
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
+
+    search_count: int = 8
+    previous_searches: dict[str, int] = Field(default_factory=dict)
+    settings: Settings = Field(default_factory=Settings)
+
+    # Gather evidence tool must be modified to understand the new evidence
+    GATHER_EVIDENCE_TOOL_PROMPT_OVERRIDE: ClassVar[str] = (
```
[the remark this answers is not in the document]
mskarlin: nix'd the space!

paperqa/sources/clinical_trials.py
mskarlin: But this comment is about the decoding, not the status (i.e. the code being 400 means it's malformed but the clinicaltrials.gov response isn't JSON in that case), so it really applies to running .json() vs. .text() on the response object.

jamesbraza (approved):
LGTM, great work here!!

PR #780, opened 2024-12-30
Docs around embeddings and agentic usage

jamesbraza:
I was explaining PaperQA and realized we could improve our docs around embeddings and when to use/not use `Docs.aadd` directly.

I also moved a `urllib.request` import to the module level.


mskarlin (approved):
thanks!

PR #785, opened 2025-01-03
Replace `raise_for_status` return `None` with Mock() in clinical trials test

mskarlin:
Two of our tests were trying to overwrite the return in response.raise_for_status, which pytest expected to be async, this fixes that.


jamesbraza (approved):
Thank you!!

PR #791, opened 2025-01-06
Fixed `Record` import from Qdrant not being in `TYPE_CHECKING` block

jamesbraza:
When not using Qdrant, installers as of v5.9.0 will get this error:

```none
.venv/lib/python3.12/site-packages/paperqa/llms.py:35: in <module>
    from qdrant_client.http.models import Record
E   ModuleNotFoundError: No module named 'qdrant_client'
```

This PR:
- Fixes the non-lazily imported type hint
- Removed an unnecessary `from __future__ import annotations`

### DIFF
--- a/paperqa/llms.py
+++ b/paperqa/llms.py
@@ -447,7 +447,7 @@ class QdrantVectorStore(VectorStore):
     @classmethod
     async def load_docs(
         cls,
-        client: AsyncQdrantClient,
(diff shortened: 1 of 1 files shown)


mskarlin (approved):

ThomasRochefortB (approved):
LGTM, it passes the `QdrantVectorStore` unit tests on my local

PR #757, opened 2024-12-10
Added llmclient dependency

maykcaldas:
This PR moves the LLMs and Embeddings used in `pqa` to [`llmclient`](https://github.com/Future-House/llm-client).

Because `pqa` also depends on `ldp`, ldp needs to be updated first. Check this PR after https://github.com/Future-House/ldp/pull/175 is merged.

TODO: Check downstream repos (`paperqa-server`, `paperqa-api`, `paperqa-app`, ...) to ensure this PR does not break any other repo.

PR #792, opened 2025-01-06
Fixing not using `set_llm_session_ids` from `fh-llm-client`

jamesbraza:
https://github.com/Future-House/paper-qa/pull/757 forgot to use `set_llm_session_ids` from `fh-llm-client`, which leads to the `session_id` being unpopulated. This PR fixes that mistake, with added test coverage

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -20,6 +20,7 @@ from llmclient import (
     LLMModel,
     LLMResult,
 )
+from llmclient.types import set_llm_session_ids
 from pydantic import (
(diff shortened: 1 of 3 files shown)


paperqa/docs.py
```diff
@@ -20,6 +20,7 @@
     LLMModel,
     LLMResult,
 )
+from llmclient.types import set_llm_session_ids
```
mskarlin, line 23: nice catch!

mskarlin (approved):

PR #793, opened 2025-01-07
Expose more contexts as output from gather_evidence tool

mskarlin:
When we approach more complex trajectories, the agent models need more info for decision making. We only return by default 1 top-ranked context, this allows us to expose more to the agent model.

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -253,11 +253,17 @@ class GatherEvidence(NamedTool):
         sorted_contexts = sorted(
             state.session.contexts, key=lambda x: x.score, reverse=True
         )
-        best_evidence = (
-            f" Best evidence:\n\n{sorted_contexts[0].context}"
-            if sorted_contexts
-            else ""
+
+        top_contexts = "\n".join(
+            [
+                f"{n + 1}. {sc.context}\n"
+                for n, sc in enumerate(
+                    sorted_contexts[: self.settings.agent.agent_evidence_n]
+                )
+            ]
         )
+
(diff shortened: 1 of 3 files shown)


paperqa/agents/tools.py
```diff
@@ -253,12 +253,18 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
...
+        top_contexts = "\n".join(
+            [
+                f"{n + 1}. {sc.context}\n"
+                for n, sc in enumerate(
+                    sorted_contexts[: self.settings.agent.agent_evidence_n]
+                )
+            ]
         )
 
+        best_evidence = f" Best evidence(s):\n\n{top_contexts}" if top_contexts else ""
```
jamesbraza, line 266: Two comments here:
- Is there a leading space by design?
- Can we do `"{self.settings.agent.agent_evidence_n} best evidence:"` to be more precise

paperqa/settings.py
```diff
@@ -467,6 +467,11 @@ class AgentSettings(BaseModel):
     )
     search_count: int = 8
     wipe_context_on_answer_failure: bool = True
+    agent_evidence_n: int = Field(
+        default=1,
```
jamesbraza, line 471:
```suggestion
        default=1,
        ge=1,
```

paperqa/agents/tools.py
mskarlin: yea, there's a leading space by design, it's been there. On your second bullet -- that seems more confusing to me? i.e. "3 best evidence:"

jamesbraza (approved):
Nice!

PR #799, opened 2025-01-09
Refactor to remove `QueryRequest` entity

jamesbraza:
The `QueryRequest` entity here served a few purposes oriented around a PaperQA web service infrastructure:

- Group query, settings, and session ID
- Provides a way to link `Docs` with a given run

`QueryRequest` also comes with a few tradeoffs:

- One more entity in the stack --> one more level to recurse down
- Complicates `deepcopy` since `QueryRequest` is stateful
- Most of the time we just need `(query, settings)`, a session ID is irrelevant

Since this repo is really more about CLI usage and the core algorithm, we don't need this entity here, so this PR removes it.

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -1,6 +1,7 @@
 import logging
 from copy import deepcopy
 from typing import Any, ClassVar, Self, cast
+from uuid import UUID
 
 from aviary.core import (
     Environment,
@@ -23,7 +24,6 @@ from paperqa.sources.clinical_trials import (
 from paperqa.types import PQASession
 from paperqa.utils import get_year
 
-from .models import QueryRequest
 from .tools import (
     AVAILABLE_TOOL_NAME_TO_CLASS,
     DEFAULT_TOOL_NAMES,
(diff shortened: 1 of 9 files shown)


paperqa/agents/env.py
```diff
@@ -207,25 +207,27 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
...
         super().__init__(**env_kwargs)
-        # Hold onto QueryRequest to create fresh tools and answer during each reset
         self._query = query
-        # Hold onto Docs to clear and reuse in state during each reset
+        self._settings = settings
         self._docs = docs
         self._llm_model = llm_model
         self._summary_llm_model = summary_llm_model
         self._embedding_model = embedding_model
+        self._session_id = session_id
```
mskarlin, line 226: One minor feature we lose here is that he query_request had the session id, which you could then match to a given response. So if you built out a bunch of queries you could match to answers after. 

Obviously the AnswerResponse object stores both, so I don't think this would cause an issue, but I'm trying to think through any edge cases.

jamesbraza: I think it's good to have less duplication of the ID within `paper-qa` itself. Ideally we have one source of truth

Let me know if you'd like any unit tests added

mskarlin:
Also we need to make a couple minor changes to Readme I think

jamesbraza:
> Also we need to make a couple minor changes to Readme I think

Done 👌 nice catch

mskarlin (approved):
I'm irrationally worried about this change because of the potential size of the affected interfaces, but I can't think of a rational reason not to approve 😄. Thanks @jamesbraza.

PR #800, opened 2025-01-09
Added `aviary-paper-data` as LitQA v2 test split

jamesbraza:
See PR title

### DIFF
--- a/paperqa/litqa.py
+++ b/paperqa/litqa.py
@@ -37,26 +37,30 @@ def make_discounted_returns(
 
 
 DEFAULT_LABBENCH_HF_HUB_NAME = "futurehouse/lab-bench"
+# Test split from Aviary paper's section 4.3: https://doi.org/10.48550/arXiv.2412.21154
+DEFAULT_AVIARY_PAPER_HF_HUB_NAME = "futurehouse/aviary-paper-data"
 
 
 def read_litqa_v2_from_hub(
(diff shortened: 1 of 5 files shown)


paperqa/litqa.py
```diff
@@ -67,9 +71,13 @@ def read_litqa_v2_from_hub(
             " `pip install paper-qa[datasets]`."
         ) from exc
 
-    litqa_v2 = load_dataset(labbench_dataset, "LitQA2")["train"].to_pandas()
-    litqa_v2["distractors"] = litqa_v2["distractors"].apply(list)
+    train_eval = load_dataset(train_eval_dataset, "LitQA2")["train"].to_pandas()
+    test = load_dataset(test_dataset, "LitQA2")["test"].to_pandas()
```
mskarlin, line 75: more of a question for me -- do we validate columns here? (i.e. is distractors checked for explicitly?)

mskarlin (approved):

paperqa/litqa.py
jamesbraza: Thanks to https://github.com/Future-House/paper-qa/blob/v5.9.2/paperqa/agents/task.py#L450-L455, we get:

- Confirmation that column attributes are present
- `typeguard` can check the types of the values during unit testing, for our defaults

To be clear, I added this comment:

> Let downstream usage in the TaskDataset's environment factories check for the
> presence of other DataFrame columns

PR #804, opened 2025-01-10
Latest tooling

jamesbraza:
Pulls in latest Ruff with `preview` rule updates

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -287,12 +287,12 @@ enable_all = true
 ignore = [
     "FURB101",  # Rely on ruff FURB101 for this
     "FURB103",  # Rely on ruff FURB103 for this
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -379,32 +383,51 @@ extend-select = [
...
+    "RUF037",
     "RUF038",
     "RUF039",
+    "RUF040",
+    "RUF041",
+    "RUF043",
     "RUF048",
+    "RUF049",
+    "RUF051",
+    "RUF052",
```
mskarlin, line 424: let's cut just this rule b.c. i think syntactically this is strange given our usage

mskarlin (approved):

pyproject.toml
jamesbraza: Done, added it to the `ignore` list

PR #805, opened 2025-01-10
Clarifying not to call `gather_evidence`, `gen_answer`, or `reset` first

jamesbraza:
@sidnarayanan showed me trajectories where the agent was calling the incorrect tool first. I just tweaked the prompts a bit here to be succinct and direct.

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -201,9 +201,9 @@ class GatherEvidence(NamedTool):
         """
         Gather evidence from previous papers given a specific question to increase evidence and relevant paper counts.
 
+        This tool is only useful when the paper count is above zero.
         A valuable time to invoke this tool is right after another tool increases paper count.
         Feel free to invoke this tool in parallel with other tools, but do not call this tool in parallel with itself.
-        Only invoke this tool when the paper count is above zero, or this tool will be useless.
 
         Args:
             question: Specific question to gather evidence for.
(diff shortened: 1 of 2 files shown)


paperqa/agents/tools.py
```diff
@@ -360,9 +361,8 @@ async def reset(self, state: EnvironmentState) -> None:
         """
         Reset by clearing all current evidence from the system.
 
-        This tool is useful when repeatedly failing to answer because the existing evidence may unsuitable for the question.
-        It does not make sense to call this tool in parallel with other tools, as its resetting all state.
-        Only invoke this tool when the current evidence is above zero, or this tool will be useless.
+        This tool is only useful when the current evidence is above zero and repeatedly failing to answer because the existing evidence may unsuitable for the question.
+        It does not make sense to call this tool in parallel with other tools, as it's resetting all state.
```
mskarlin, line 365:
```suggestion
        It does not make sense to call this tool in parallel with other tools, as it's resetting the state.
```

mskarlin (approved):
did we test if this works?

sidnarayanan:
I vote we don't proceed with this now:
1. We haven't tested if it works or changes performance
2. The particular trajectories I shared that kicked this off were (a) from one of my agents [nothing we have in the platform] and (b) sampled with a diversity bias. 

TLDR I'm wary of putting in a change based on some edge case I observed

jamesbraza:
Sure can hold off on this, I am ambivalent. We can keep in mind that we should add monitoring someday for undesirable trajectories

PR #808, opened 2025-01-10
Fixing tool failures crashing runners in failover answer

jamesbraza:
During normal `PaperQAEnvironment.step`, we call tools using `exec_tool_calls(..., handle_tool_exc=True)`. This means we don't crash on tool failures.

During the rollout's truncation's failover `gen_answer` invocation, we didn't have the same `handle_tool_exc=True` protection in place, so we could crash. This PR just moves the failover `gen_answer` invocation to use `exec_tool_calls` for the same safety.


mskarlin (approved):
nice catch!

PR #812, opened 2025-01-14
Fixing `gather_evidence` and `complete` response messages

jamesbraza:
This PR fixes two misconceptions:
- https://github.com/Future-House/paper-qa/pull/809 incorrectly based the relevant docs added count to `l0`, but it should have only been relevant ones
- If `complete` tool gets passed `has_successful_answer` its response starts with "Success". However, if it turns out the answer is wrong, this response may not make sense. Now the response starts with ~"Sure"~ "Certain" to be clear

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -17,7 +17,7 @@ from pydantic import BaseModel, ConfigDict, Field, computed_field
 from paperqa.docs import Docs
 from paperqa.settings import Settings
 from paperqa.sources.clinical_trials import add_clinical_trials_to_docs
-from paperqa.types import DocDetails, PQASession
+from paperqa.types import Context, DocDetails, PQASession
 
 from .search import get_directory_index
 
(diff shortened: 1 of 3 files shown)


paperqa/agents/tools.py
```diff
@@ -412,10 +405,10 @@ async def complete(
 
         logger.info(
             f"Completing '{state.session.question}' as"
-            f" '{'a success' if has_successful_answer else 'unsure'}'."
+            f" '{'sure' if has_successful_answer else 'unsure'}'."
         )
         # Return answer and status to simplify postprocessing of tool response
-        return f"{'Success' if has_successful_answer else 'Unsure'} | {state.status}"
+        return f"{'Sure' if has_successful_answer else 'Unsure'} | {state.status}"
```
whitead, line 411: Maybe use `Certain` instead? `Sure` feels weird here (but I get that it is symmetric with Unsure)

jamesbraza: Nice suggestion, implemented

whitead:
can you deploy on dev? Would love to just double-check this

jamesbraza:
> can you deploy on dev? Would love to just double-check this

It works, and I added several expansions to the `test_agent_sharing_state` unit test to handle this

whitead (approved):

PR #817, opened 2025-01-17
Documenting citation style

jamesbraza:
This PR aims to help with https://github.com/Future-House/paper-qa/discussions/814:

- Documents what citation style we use (MLA parenthetical in-text citation)
~- Adds this information into the `qa_prompt`~

### DIFF
--- a/paperqa/prompts.py
+++ b/paperqa/prompts.py
@@ -26,10 +26,9 @@ qa_prompt = (
     f'"{CANNOT_ANSWER_PHRASE}." '
     "For each part of your answer, indicate which sources most support "
     "it via citation keys at the end of sentences, "
-    "like {example_citation}, which uses the {citation_style} citation style. "
-    "Only cite from the context below and only use the valid keys. "
-    "Write in the style of a Wikipedia article, "
-    "with concise sentences and coherent paragraphs. "
+    "like {example_citation}. Only cite from the context "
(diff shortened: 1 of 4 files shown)


paperqa/prompts.py
```diff
@@ -26,9 +26,10 @@
     f'"{CANNOT_ANSWER_PHRASE}." '
     "For each part of your answer, indicate which sources most support "
     "it via citation keys at the end of sentences, "
-    "like {example_citation}. Only cite from the context "
-    "below and only use the valid keys. Write in the style of a "
-    "Wikipedia article, with concise sentences and coherent paragraphs. "
+    "like {example_citation}, which uses the {citation_style} citation style. "
+    "Only cite from the context below and only use the valid keys. "
```
mskarlin, line 30: I'm all for documenting this but why does the LLM need to know? I think this might just be distracting to it.

jamesbraza: I guess I was thinking perhaps information on the citation style could help the LLM, but since we make the references ourselves in https://github.com/Future-House/paper-qa/blob/v5.10.2/paperqa/docs.py#L296-L326, I guess this information mostly isn't useful to the LLM. I will revert this change, nice suggestion

jamesbraza: Alright, pushed

mskarlin (approved):

mskarlin (approved):

PR #818, opened 2025-01-17
Allowing more general types in `Settings`

jamesbraza:
- Allowing `Sequence` for `callbacks`
- Allowing `Mapping` for `doc_filter`
- Allowing `set[str] | Sequence[str]` for `tool_names`
    - https://github.com/Future-House/paper-qa/pull/289 had moved it from `Collection[str]` to `set[str]`

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -6,7 +6,7 @@ import os
 import re
 import tempfile
 import urllib.request
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from datetime import datetime
 from functools import partial
 from io import BytesIO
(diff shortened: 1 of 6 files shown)


paperqa/docs.py
```diff
@@ -690,12 +690,16 @@ async def aquery(  # noqa: PLR0912
...
         query: PQASession | str,
         settings: MaybeSettings = None,
-        callbacks: list[Callable] | None = None,
+        callbacks: Sequence[Callable] | None = None,
         llm_model: LLMModel | None = None,
         summary_llm_model: LLMModel | None = None,
         embedding_model: EmbeddingModel | None = None,
         partitioning_fn: Callable[[Embeddable], int] | None = None,
     ) -> PQASession:
+        # TODO: remove list cast after release of https://github.com/Future-House/llm-client/pull/35
```
mskarlin, line 699: why do we care about the flexibility of this being a sequence if we just cast to a list here?

jamesbraza: If that PR in `llm-client` was merged, we wouldn't need this `list` cast at all, we can just directly pass through

mskarlin (approved):

PR #819, opened 2025-01-17
Clinical trials docs and bugfixes

mskarlin:
- Adds a tutorial for using clinical trials, planning on pointing our gitbook to this subdirectory to build out a blog
- Fixes an issue where a index was being built even if it wasn't needed, we now require a `paper_search` tool to be included for it to trigger
- Silenced some warning logs for malformed cookies coming back from clincaltrials.gov, added a utility to do this in general
- Manually closed sessions that are built on the fly if using the clinical trials search without an external session 
- lowered the cap on clinical trial size truncation a tiny bit given some edge case responses I was seeing in testing


paperqa/sources/clinical_trials.py
```diff
@@ -29,45 +29,58 @@
 SEARCH_PAGE_SIZE = 1000
 TRIAL_API_FIELDS = "protocolSection,derivedSection"
 DOWNLOAD_CONCURRENCY = 20
-TRIAL_CHAR_TRUNCATION_SIZE = 30_000  # larger will prevent embeddings from working
+TRIAL_CHAR_TRUNCATION_SIZE = 28_000  # larger will prevent embeddings from working
 MALFORMATTED_QUERY_STATUS: int = 400
 
 
+class CookieWarningFilter(logging.Filter):
+    """Filters out bad cookie warning due to clincialtrials.gov response."""
```
jamesbraza, line 37: Can you update to state:

- Why the warning is "bad"
- What is "bad" exactly

Maybe reword to say """Filter out log spam from to clincialtrials.gov responses."""

paperqa/sources/clinical_trials.py
```diff
@@ -29,45 +29,58 @@
 SEARCH_PAGE_SIZE = 1000
 TRIAL_API_FIELDS = "protocolSection,derivedSection"
 DOWNLOAD_CONCURRENCY = 20
-TRIAL_CHAR_TRUNCATION_SIZE = 30_000  # larger will prevent embeddings from working
+TRIAL_CHAR_TRUNCATION_SIZE = 28_000  # larger will prevent embeddings from working
```
jamesbraza, line 32: 28k is pretty specific, can you expand the comment saying why 30k was too high

mskarlin: the comment is right next it no?

jamesbraza: Yeah but we changed from 30k to 28k and the comment didn't change. So there's more to the story, I was hoping we can document that

mskarlin: it's an estimate to try to hit 8k and we were off, added

paperqa/utils.py
```diff
@@ -519,3 +520,25 @@ def extract_thought(content: str | None) -> str:
     "peer-review": "misc",  # No direct equivalent, so 'misc' is used
     "other": "article",  # Assume an article if we don't know the type
 }
+
+
+@contextlib.contextmanager
+def logging_filters(loggers: set[str], filters: set[type[logging.Filter]]):
```
jamesbraza, line 526:
```suggestion
def logging_filters(loggers: Collection[str], filters: Collection[type[logging.Filter]]):
```

Can we do `Collection` or `Sequence` here

jamesbraza (approved):
Nice work

paperqa/utils.py
mskarlin: sure

PR #828, opened 2025-01-23
Automate fetching the current year

harryvu-futurehouse:
## Automate fetching the current year for `citation_prompt`

### DIFF
--- a/paperqa/prompts.py
+++ b/paperqa/prompts.py
@@ -1,3 +1,4 @@
+from .agents.helpers import get_year
 summary_prompt = (
     "Summarize the excerpt below to help answer a question.\n\nExcerpt from"
     " {citation}\n\n----\n\n{text}\n\n----\n\nQuestion: {question}\n\nDo not directly"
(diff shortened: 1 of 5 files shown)


paperqa/prompts.py
```diff
@@ -47,10 +47,11 @@
     "Papers: {papers}\n\n"
     "Selected keys:"
 )
+
 citation_prompt = (
     "Provide the citation for the following text in MLA Format. "
     "Do not write an introductory sentence. "
-    "If reporting date accessed, the current year is 2024\n\n"
+    "If reporting date accessed, the current year is 2025\n\n"
```
jamesbraza, line 54:
```suggestion
    f"If reporting date accessed, the current year is {get_year()}\n\n"
```

And you'll need to import this

tests/test_configs.py
```diff
@@ -167,3 +169,11 @@ def test_o1_requires_temp_equals_1() -> None:
 def test_matches_filter_criteria(doc_class, doc_data, filter_criteria, expected_result):
     doc = doc_class(**doc_data)
     assert doc.matches_filter_criteria(filter_criteria) == expected_result
+
+
+def test_citation_prompt_current_year():
+    expected_year_text = f"the current year is {datetime.now().year}"
```
jamesbraza, line 175: Let's use `utils.get_year` here to avoid directly aliasing the SUT

jamesbraza (approved):

PR #839, opened 2025-01-29
Propagates llmclient changes to pqa

maykcaldas:
- [x] Merge https://github.com/Future-House/llm-client/pull/49
- [x] Release llmclient v0.1.0
- [x] Merge https://github.com/Future-House/ldp/pull/226
- [x]  Update fh-llm-client and ldp dependencies in pyproject.toml
- [x]  Go through cassettes and check if all changes are needed
- [x]  Re-test this PR
- [ ]  Merge


paperqa/agents/models.py
```diff
@@ -79,12 +80,16 @@ async def get_summary(self, llm_model: LLMModel | str = "gpt-4o") -> str:
...
+        prompt = "{question}\n\n{answer}"
+        messages = [
+            Message(role="system", content=sys_prompt),
+            Message(role="user", content=prompt.format(**data)),
+        ]
+        result = await model.call_single(  # run_prompt is deprecated
+            messages=messages,
         )
-        return result.text.strip()
+        return cast(str, result.text).strip()
```
maykcaldas, line 92: `LLMResult.text` is typed as `str | None` to allow default to `None`, but it's populated in `call`. Is it better to `cast` of check `if isinstance(result.text, str)` here?

paperqa/agents/helpers.py
```diff
@@ -60,12 +61,14 @@ async def litellm_get_search_query(
...
-        system_prompt=None,
+    data = {"question": question, "count": count}
+    messages = [
+        Message(role="user", content=search_prompt.format(**data)),
+    ]
+    result = await model.call_single(  # run_prompt is deprecated
+        messages=messages,
     )
-    search_query = result.text
+    search_query = cast(str, result.text)
```
maykcaldas, line 71: `LLMResult.text` is typed as `str | None` to allow default to `None`, but it's populated in `call`. Is it better to `cast` of check `if isinstance(result.text, str)` here?

paperqa/agents/helpers.py
```diff
@@ -60,12 +61,14 @@ async def litellm_get_search_query(
...
     else:
         model = llm
-    result = await model.run_prompt(
-        prompt=search_prompt,
-        data={"question": question, "count": count},
-        system_prompt=None,
+    data = {"question": question, "count": count}
+    messages = [
+        Message(content=search_prompt.format(**data)),
+    ]
```
jamesbraza, line 67:
```suggestion
    messages = [
        Message(content=search_prompt.format(question=question, count=count)),
    ]
```

No need for kwargs

paperqa/agents/helpers.py
jamesbraza: Use `cast` when you're sure. Sounds like you're sure, so I think `cast` is fine

paperqa/agents/helpers.py
maykcaldas: I cleaned kwargs when up to two variables were being passed

paperqa/docs.py
```diff
@@ -712,12 +718,17 @@ async def aquery(  # noqa: PLR0912
         pre_str = None
         if prompt_config.pre is not None:
             with set_llm_session_ids(session.id):
-                pre = await llm_model.run_prompt(
-                    prompt=prompt_config.pre,
-                    data={"question": session.question},
+                messages = [
+                    Message(role="system", content=prompt_config.system),
+                    Message(
+                        role="user",
```
jamesbraza, line 724: We don't need to restate default user. Also, feel free to ignore

tests/test_agents.py
```diff
@@ -352,12 +352,12 @@ async def on_agent_action(  # noqa: RUF029
     async def llm_model_call(*args, **kwargs):
         # NOTE: "required" will not lead to thoughts being emitted, it has to be "auto"
         # https://docs.anthropic.com/en/docs/build-with-claude/tool-use#chain-of-thought
-        kwargs.pop("tool_choice", MultipleCompletionLLMModel.TOOL_CHOICE_REQUIRED)
+        # kwargs.pop("tool_choice", LiteLLMModel.TOOL_CHOICE_REQUIRED)
+        # tool_choice is now a arg, not a kwarg
+        args = args[:-1]  # removing last element from args
```
jamesbraza, line 357:
```suggestion
        args = args[:-1]  # removing last element (tool_choice) from args
```

No need for legacy comments

tests/test_paperqa.py
```diff
@@ -463,10 +467,12 @@ def accum(x) -> None:
...
 
     llm = anthropic_settings.get_llm()
-    completion = await llm.run_prompt(
-        prompt="The {animal} says",
-        data={"animal": "duck"},
-        system_prompt=None,
+    prompt = "The {animal} says"
+    messages = [
+        Message(role="user", content=prompt.format(animal="duck")),
+    ]
```
jamesbraza, line 473:
```suggestion
    messages = [
        Message(role="user", content="The duck says"),
    ]
```

Can we not just do this?

tests/test_paperqa.py
```diff
@@ -757,18 +763,50 @@ def test_hybrid_embedding(stub_data_dir: Path, vector_store: type[VectorStore])
 
 
 def test_custom_llm(stub_data_dir: Path) -> None:
-    from llmclient import Chunk
+    # ASK: Because acompletion_iter is @rate_limited in LiteLLMModel, we need to mock it here.
```
jamesbraza, line 766: Can we not upstream this entire thing to `llm-client`? And if not, what makes this `paper-qa` specific?

maykcaldas: This is not `pqa` specific. But `rate_limited`, changes the output from `acompletion_iter`. Instead of mocking `rate_limited`, I imported it from `llmclient` and implemented a dummy `check_rate_limit` here.

jamesbraza: What I am not following is, didn't we entirely remove `acompletion_iter` in the LLM client refactor?

maykcaldas: I think I understand the question now. We have `LLMModel`, which creates the user interface (`call` is here), manages the tools and callbacks, and should be independent of the LLM provider. 

And we have the child class `LiteLLMModel`, which implements `acompletion` and `acompletion_iter` using `litellm`. These two methods are called by `call` to get the completions. Because in this test we are mocking a custom_llm, it needs to implement `acompletion` and `acompletion_iter`. The `call` method is used when using the `StubLLMModel`.

jamesbraza: Oh wait a minute, why is `acompletion_iter` still in `LLMModel`? I thought we were removing everything except `call` and `call_single` in the refactor?

paperqa/agents/models.py
```diff
@@ -79,12 +80,20 @@ async def get_summary(self, llm_model: LLMModel | str = "gpt-4o") -> str:
...
-        result = await model.run_prompt(
-            prompt="{question}\n\n{answer}",
-            data={"question": self.session.question, "answer": self.session.answer},
-            system_prompt=sys_prompt,
+        prompt = "{question}\n\n{answer}"
+        messages = [
+            Message(role="system", content=sys_prompt),
+            Message(
+                role="user",
+                content=prompt.format(
```
jamesbraza, line 88:
```suggestion
        prompt_template = "{question}\n\n{answer}"
        messages = [
            Message(role="system", content=sys_prompt),
            Message(
                role="user",
                content=prompt_template.format(
```

[naming nit] let's name it a template if it's not a full prompt

paperqa/docs.py
```diff
@@ -294,12 +293,15 @@ async def aadd(  # noqa: PLR0912
...
-                prompt=parse_config.citation_prompt,
-                data={"text": texts[0].text},
-                system_prompt=None,  # skip system because it's too hesitant to answer
+            messages = [
+                Message(
+                    content=parse_config.citation_prompt.format(text=texts[0].text)
+                ),
+            ]
+            result = await llm_model.call_single(
+                messages=messages,
```
jamesbraza, line 302:
```suggestion
            result = await llm_model.call_single(
                messages=[
                    Message(
                        content=parse_config.citation_prompt.format(text=texts[0].text)
                    ),
                ],
```

If you're going to use keyword argument in `call_single`, imo no need to have the `messages` local too

tests/test_paperqa.py
```diff
@@ -757,18 +762,43 @@ def test_hybrid_embedding(stub_data_dir: Path, vector_store: type[VectorStore])
...
 
-        async def acomplete(self, prompt: str) -> Chunk:  # noqa: ARG002
-            return Chunk(text="Echo", prompt_tokens=1, completion_tokens=1)
+        async def acompletion(
+            self, messages: list[Message], **kwargs
+        ) -> list[LLMResult]:
+            # NOTE: Here we add kwargs to the function signature to make it compatible with the superclass
+            # and avoid #type: ignore[override]
+            # and we delete kwargs to avoid ARG002 (unused argument)
+            del kwargs
```
jamesbraza, line 774: Let's just use `noqa` over the `del` + explanation

tests/test_paperqa.py
jamesbraza: We discussed in person, thanks @maykcaldas . Resolving this one

paperqa/core.py
```diff
@@ -51,12 +54,13 @@ async def map_fxn_summary(
...
         text: The text to parse.
-        question: The question to use for the chain.
-        prompt_runner: The prompt runner to call - should have question, citation,
-            summary_length, and text fields.
-        extra_prompt_data: Optional extra kwargs to pass to the prompt runner's data.
-        parser: The parser to use for parsing - return empty dict on Failure to fallback to text parsing.
-        callbacks: LLM callbacks to execute in the prompt runner.
+        question: The question to use for summarization.
+        summary_llm_model: The LLM model to use for generating summaries.
+        prompt_details: Tuple containing the prompt template and system prompt.
```
jamesbraza, line 59: > the prompt template and system prompt

- Isn't the system prompt a prompt template
- It should say user prompt

paperqa/core.py
```diff
@@ -38,7 +40,8 @@ def escape_newlines(match: re.Match) -> str:
 async def map_fxn_summary(
     text: Text,
     question: str,
-    prompt_runner: PromptRunner | None,
+    summary_llm_model: LLMModel | None,
+    prompt_details: tuple[str, str] | None,
```
jamesbraza, line 44: I don't like the name "prompt details, really this is two prompt templates.

Can you make the name `prompt_templates` or something more explicit

paperqa/core.py
maykcaldas: Renamed it to `prompt_templates` and updated the docstring
```
prompt_templates: Tuple containing templates for the user and system prompts.
```

paperqa/core.py
```diff
@@ -71,14 +75,25 @@ async def map_fxn_summary(
...
-            | (extra_prompt_data or {}),
-            callbacks,
-            "evidence:" + text.name,
+    if summary_llm_model and prompt_templates:
+        data = {
+            "question": question,
+            "citation": citation,
+            "text": text.text,
+            **(extra_prompt_data or {}),
+        }
```
jamesbraza, line 84:
```suggestion
        data = {
            "question": question,
            "citation": citation,
            "text": text.text
        } | (extra_prompt_data or {}),
```

Previously we could pass `text` in `extra_prompt_data` and override, but now we can't.

Let's avoid doing niche breaking changes like this within a refactor PR

tests/test_paperqa.py
```diff
@@ -766,18 +771,39 @@ def test_hybrid_embedding(stub_data_dir: Path, vector_store: type[VectorStore])
...
 
-        async def acomplete_iter(
-            self, prompt: str  # noqa: ARG002
-        ) -> AsyncIterable[Chunk]:
-            yield Chunk(text="Echo", prompt_tokens=1, completion_tokens=1)
+        async def check_rate_limit(
+            self, token_count: float, **kwargs  # noqa: ARG002
+        ) -> None:
+            # Implements a dummy rate limit checker.
+            return
```
jamesbraza, line 806:
```suggestion
        ) -> None:
            """This is a dummy check."""
```

No need for `return`

paperqa/core.py
```diff
@@ -71,14 +75,22 @@ async def map_fxn_summary(
...
-            {"question": question, "citation": citation, "text": text.text}
-            | (extra_prompt_data or {}),
-            callbacks,
-            "evidence:" + text.name,
+    if summary_llm_model and prompt_templates:
+        data = {"question": question, "citation": citation, "text": text.text} | (
+            extra_prompt_data or {}
+        )
+        message_prompt = prompt_templates[0]
+        system_prompt = prompt_templates[1]
```
jamesbraza, line 83:
```suggestion
        message_prompt, system_prompt = prompt_templates
```

Two benefits here:
- We confirm the correct tuple length, otherwise it throws a `ValueError`
- Less lines

paperqa/docs.py
```diff
@@ -806,14 +819,19 @@ async def aquery(  # noqa: PLR0912
...
-                    prompt=prompt_config.post,
-                    data=session.model_dump(),
+                messages = [
+                    Message(role="system", content=prompt_config.system),
+                    Message(
+                        role="user",
+                        content=prompt_config.post.format(question=session.question),
+                    ),
+                ]
+                post = await llm_model.call_single(  # is deprecated
```
jamesbraza, line 829: Can you expand on this "is deprecated" -- what is deprecated here?

paperqa/core.py
maykcaldas: Should we add any more information about the specific order in the doctrine? Currently it says
`prompt_templates: Tuple containing templates for the user and system prompts.`

Maybe edit it to:
```
prompt_templates: Tuple containing templates for the user and system prompts.
    prompt_templates = (user_prompt_templat, system_prompt_template)
```
?

paperqa/docs.py
maykcaldas: This is a leftover. I removed the comment

.pre-commit-config.yaml
```diff
@@ -84,7 +84,7 @@ repos:
           - aiohttp>=3.10.6 # Match pyproject.toml
           - PyMuPDF>=1.24.12
           - anyio
-          - fh-llm-client[deepseek]>=0.0.11 # Match pyproject.toml
+          - fh-llm-client @ git+https://github.com/Future-House/llm-client.git@main
```
jamesbraza, line 87:
```suggestion
          - fh-llm-client[deepseek] @ git+https://github.com/Future-House/llm-client.git@main
```

paperqa/core.py
jamesbraza: Yeah that sounds good. Also maybe start with "Optional two-tuple" to make it clear it's `| None` and the tuple has two entries

paperqa/docs.py
jamesbraza: It's still here, did you forget to push?

.pre-commit-config.yaml
maykcaldas: Oh, cool. Didn't know I could do that

paperqa/agents/task.py
```diff
@@ -549,13 +550,14 @@ async def _evaluate_answer(self) -> dict:
...
         }
 
-        result = await pairwise_eval_llm.run_prompt(
-            prompt=lfrqa_prompt_template,
-            data=data,
-            system_prompt=lfrqa_system_prompt,
+        result = await pairwise_eval_llm.call_single(
+            messages=[
+                Message(role="system", content=lfrqa_prompt_template.format(**data)),
+                Message(role="user", content=lfrqa_system_prompt.format(**data)),
```
jamesbraza, line 556: I don't think the system prompt is a prompt template

maykcaldas: I applied a solution similar to what I did here:
https://github.com/Future-House/paper-qa/blob/update-llmclient/paperqa/core.py#L85.

There is this system prompt which has a variable: https://github.com/Future-House/paper-qa/blob/update-llmclient/paperqa/prompts.py#L76

Is it bad to be general? In case there is no `{var}` in the prompt template, `format` is not changing anything.

jamesbraza: I know what you're saying that `.format` is a no-op if there's no templates variables. However, it's not risk-free because a normal prompt could have template variables, without actually intending them to be a prompt template.

I think if it's a prompt template we use `.format`, otherwise we don't use `.format`. So in this case, I prefer we drop the `.format` on the system prompt (since it's not a template)

paperqa/agents/task.py
```diff
@@ -549,13 +550,14 @@ async def _evaluate_answer(self) -> dict:
...
         }
 
-        result = await pairwise_eval_llm.run_prompt(
-            prompt=lfrqa_prompt_template,
-            data=data,
-            system_prompt=lfrqa_system_prompt,
+        result = await pairwise_eval_llm.call_single(
+            messages=[
+                Message(role="system", content=lfrqa_prompt_template),
+                Message(role="user", content=lfrqa_system_prompt.format(**data)),
```
jamesbraza, line 556:
```suggestion
                Message(role="system", content=lfrqa_system_prompt),
                Message(role="user", content=lfrqa_prompt_template.format(**data)),
```

@maykcaldas you got it backwards in two ways

- You have "system prompt" being given `"user"` role
- You have `.format` on the system prompt, but it's the user prompt that is the template

paperqa/docs.py
```diff
@@ -766,19 +776,22 @@ async def aquery(  # noqa: PLR0912
...
-                    },
+                data = {
+                    "context": context_str,
+                    "answer_length": answer_config.answer_length,
+                    "question": session.question,
+                    "example_citation": prompt_config.EXAMPLE_CITATION,
+                }
+                messages = [
+                    Message(role="system", content=prompt_config.system),
+                    Message(role="user", content=prompt_config.qa.format(**data)),
```
jamesbraza, line 787: If we only use `data` once, it's extra indirection to store a `dict` then use `**` expansion in the next line. Can we use normal kwargs

```python
content=prompt_config.qa.format(
    context=context_str,
    answer_length=answer_config.answer_length,
))
```

jamesbraza (approved):
Nice work, a long PR but got it over the finish line.

Adios to the confusing `run_prompt`, hello directly made `Message`s

.mailmap
```diff
@@ -1,12 +1,13 @@
...
-Odhran O'Donoghue <odhran.r.odonoghue@gmail.com> <39832722+odhran-o-d@users.noreply.github.com>
-Samantha Cox <samc@futurehouse.org> <swrig30@ur.rochester.edu>
-Anush008 <anushshetty90@gmail.com> Anush <anushshetty90@gmail.com>
-Mayk Caldas <mayk@futurehouse.com> maykcaldas <maykcaldas@gmail.com>
 Harry Vu <harry@futurehouse.org> <harryvu-futurehouse@users.noreply.github.com>
 Harry Vu <harry@futurehouse.org> harryvu-futurehouse
+James Braza <james@futurehouse.org> <jamesbraza@gmail.com>
+Mayk Caldas <mayk@futurehouse.org> maykcaldas
+Mayk Caldas <mayk@futurehouse.org> <maykcaldas@gmail.com>
+ Michael Skarlinski <mskarlinski@futurehouse.org> mskarlin <12701035+mskarlin@users.noreply.github.com>
```
jamesbraza, line 10: Can you remove the leading spaces

paperqa/docs.py
jamesbraza: Can you still do this one? I don't think we need the `data` local

jamesbraza (approved):

Issue #749, opened 2024-12-04
Retrying relevance score prompt if score was a float or response isn't valid JSON


jamesbraza:
Our LLM response --> `Context` logic is a bit fragile:
- https://github.com/Future-House/paper-qa/issues/747 shows the LLM responding with a float relevance score.
- https://github.com/Future-House/paper-qa/issues/736 shows the LLM failing to respond with valid JSON (though actually the issue seems to have been a token limitation).

What currently happens is this blows up `Docs.aquery`.

However, this should not be a critical failure because:

1. It's not in the caller's (e.g. an agent) control, it's a failure internal to `paper-qa`.
2. We can retry the summary prompt once or twice, saying something like:
    - `"We failed to parse valid JSON according to exception {exc!r}. Please respond with JSON."`
    - `"We failed to extract a valid score according to exception {exc!r}. Please provide an integer score in 1-10."`

PR #843, opened 2025-01-31
Changes to llm_parse_json() to handle issues when LLM does not respond as per instructions

foreverallama:
Hey!
I was trying this project out using the new DeepSeek R1 (distilled) model with ollama, when I came across a few issues in the json parsing 

- The LLM sometimes would not respond with JSON syntax, throwing an error. This happened quite a lot with DeepSeek R1 and sometimes with Llama3.2 as well (running locally with ollama)
- When the LLM does not respond with JSON, the <think> tags were being retained in the response (specific to R1 and maybe whatever other reasoning models are out there)
- This happened more with DeepSeek, where the context score was being returned as a float (also mentioned in issue #749 ). Pydanctic does not like this apparently.
 
I added a few lines in `core.py` to address these issues I was facing, without really modifying any part of the original code. They go at the end of the `llm_parse_json()` method, essentially reformatting the string if it could not perform JSON decoding and trying again. Additions as follows:

1. Tries to decode JSON, if it fails it tries some more formatting
2. First is to remove <think> reasoning texts using a simple regex
3. Wrapping the remaining response in a dictionary with key as "summary" and trying to decode JSON again
4. Finally, to handle the LLM returning float `relevance_scores`, I just replaced the score as `round(float(score))` when returning the context within `map_fxn_summary()`. Not sure if this is a preferred way to handle it however.

I did feel it's a bit specific to the models I tested on _(4o, DeepSeek R1, Llama3.2)_. Also, it's still not perfect as there are still many more variations of the model responses that need to be handled. I'd be glad to add some tests if this is a direction to proceed in or maybe help with a rework as well!

I did try querying without JSON parsing, but this was taking at least **~3-4x** the amount of time to return a response when running the model locally. I didn't go through the codebase for this case, but my guess is with DeepSeek R1, the reasoning part of the response was being added, increasing context lengths and hence, processing time. If this is the case, maybe a simple stripping of anything between <think> and </think> might work well.


paperqa/core.py
```diff
@@ -110,7 +137,7 @@ async def map_fxn_summary(
                 name=text.name,
                 doc=text.doc.model_dump(exclude={"embedding"}),
             ),
-            score=score,  # pylint: disable=possibly-used-before-assignment
+            score=round(float(score)),  # pylint: disable=possibly-used-before-assignment
```
jamesbraza, line 140: We were rejecting scores like `5.6`, just to confirm does this change now accept it as score `6`?

paperqa/core.py
```diff
@@ -27,6 +27,33 @@ def escape_newlines(match: re.Match) -> str:
...
+        print(f"[WARNING] Failed to parse JSON. Attempting to wrap response in dictionary")
+    except Exception as e:
+        raise ValueError(
+            f"Failed to parse JSON from text {text!r}. Your model may not be capable of"
+            " supporting JSON output or our parsing technique could use some work. Try"
+            " a different model or specify `Settings(prompts={'use_json': False})`"
+        ) from e
+    
+    # Removing <think> tags for reasoning models (happens if the LLM does not produce a JSON response)
+    if "<think>" in ptext:
```
jamesbraza, line 40: Can you add a unit test for this?

paperqa/core.py
foreverallama: Yes! I was getting scores like 0.8 as well, so I decided to round instead of just typecasting to int so that it doesn't get filtered out later

paperqa/core.py
foreverallama: Sure! Will do in the next week.

foreverallama:
So I collected a bunch of LLM responses to observe response variability and grouped them based on the type of differences observed. Here's a list below. Few of them already have tests, I'll compile these cases together, update the code to handle these cases if needed, and add tests for each case (if not already present)

LLM Specs: 
- DeepSeek R1 (Distill Qwen 7B)
- Llama3.2:3B
- mxbai-embed-large (embed model)
- Ollama v0.5.7

### 1. Standard JSON in Markdown Code Blocks
The response is formatted correctly as JSON and enclosed within a code block.

### 2. Standard JSON but without Markdown Code Block
The JSON structure is valid but is not enclosed in a code block.

### 3. Standard JSON with Extra Remarks
JSON is present but preceded or followed by introductory or closing remarks. The JSON may or may not be enclosed within a code block.

### 4. Issues with `relevance_score` Key
Variations include:
  - Float: `"relevance_score": 8.5`
  - String: `"relevance_score": "8"`
  - This thing: `"relevance_score": "6/10"`
  - Key name variations like `"relevance"`, `"score"`, or "relevance score", etc.
  
### 5. Extra Keys with Relevant Information (happened rarely with Llama3.2)
- Some responses include additional keys (e.g., `"additional_notes"`, `"description"`, `"details"`) that contain text relevant to the summary which could be merged

### 6. <think> tags for Reasoning models (DeepSeek R1)
These models first print internal reasoning between <think> and </think> followed by the actual response.

### 6. Broken JSON Formatting (extremely rare)
Missing or misaligned:
  - Curly braces `{}`.
  - Quotes `""`.
  - Commas `,`.

### 7. Non-JSON Responses
- Expected response but in markdown formatting instead of a JSON code block. The relevance score is sometimes included in spoken language.
- Complete LLM response including any remarks, etc. Again may or may not contain a relevance score.
- Irrelevant response
Side note: got a response in Chinese with DeepSeek lol

foreverallama:
Updated `llm_parse_json()` to handle the following cases:

1. Removing <think> tags
2. JSON formatted response, with or without MD code block
3. Relevance score returned as a string, float, or fraction (5/10). (Unmade changes to handle this within `map_fxn_summary()`)
4. `relevance_score` key named incorrectly
5. JSON formatting issues due to misplaced `,`
6. Handling an unexpected non-JSON response 


Added multiple tests for each case in `tests/test_llm_json_parser.py`. Also incorporated couple of tests from `test_paperqa.py` to this one.

paperqa/core.py
```diff
@@ -12,9 +12,21 @@
...
+
+    # Removing <think> tags for reasoning models
+    ptext = re.sub(r"<think>.*?</think>", "", ptext, flags=re.DOTALL).strip()
+
+    # fetches from markdown ```json if present
+    ptext = ptext.split("```json")[-1].split("```")[0]
+
+    # Wrap non-JSON text in a dictionary
+    if "{" not in ptext and "}" not in ptext:
+        print("WRAPPING IN DICT")
```
jamesbraza, line 26: Can you get rid of `print`

paperqa/core.py
```diff
@@ -25,15 +37,63 @@ def escape_newlines(match: re.Match) -> str:
     # https://regex101.com/r/VFcDmB/1
     pattern = r'"(?:[^"\\]|\\.)*"'
     ptext = re.sub(pattern, escape_newlines, ptext)
+
+    def fraction_replacer(match):
```
jamesbraza, line 41:
```suggestion
    def fraction_replacer(match: re.Match) -> str:
```

tests/test_llm_json_parser.py
```diff
@@ -0,0 +1,142 @@
+import textwrap
+
+import pytest
+
+from paperqa.core import llm_parse_json
+
+
+# test cases obtained from test_paperqa.py
```
jamesbraza, line 8: No need for this comment

tests/test_llm_json_parser.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+
+{
+    "example": "json"
+}
+
+""",
+    ],
+)
+def test_llm_parse_json(example: str) -> None:
```
jamesbraza, line 37: Can you combine with the previous `test_llm_parse_json`? You might need to rebase atop main as we recently expanded it in https://github.com/Future-House/paper-qa/pull/858

tests/test_llm_json_parser.py
```diff
@@ -0,0 +1,142 @@
...
+        }"""
+    )
+    assert llm_parse_json(example) == {
+        "summary": "A line\n\nAnother line",
+        "relevance_score": 7,
+    }
+
+
+# Additional cases
+# Case 1: Removing Think Tags
```
jamesbraza, line 59: Can you move this case descriptor to `pytest.param(..., id="removing-think-cases")`

jamesbraza (commented):
Getting closer

tests/test_llm_json_parser.py
foreverallama: I think I just copied the same the tests which was in `test_paperqa.py`. Should I keep this file and remove the tests from `test_paperqa.py` or move them all back to the original file?

jamesbraza: Let's move to `test_paperqa.py` and if possible group them into a test class with a decent name.

Also, if you can use more parameterize cases + consolidate a few of these test functions, it would be appreciated

foreverallama: Did these changes and added them to the end of `test_paperqa.py`. Let me know if its okay

jamesbraza: Yeah looks like `pre-commit` CI is failing. Can you get it passing? A rough overview of commands we need to pass

```
# Install dev set up (from repo root)
uv sync
uv pip install ruff
source .venv/bin/activate
pre-commit install
pre-commit run -a
```

It looks like you have some invalid syntax, and some lines that are too long. Feel free to `# noqa` the long lines

foreverallama: Should be fixed now, I forgot to remove the old test file after integrating into `test_paperqa`

paperqa/core.py
```diff
@@ -29,15 +40,63 @@ def escape_double_backslashes(match: re.Match) -> str:
...
+        data = json.loads(ptext)
     except json.JSONDecodeError as e:
         raise ValueError(
             f"Failed to parse JSON from text {text!r}. Your model may not be capable of"
             " supporting JSON output or our parsing technique could use some work. Try"
             " a different model or specify `Settings(prompts={'use_json': False})`"
         ) from e
 
+    for key in list(data.keys()):
+        if re.search(r"re.?levance|score", key, re.IGNORECASE):
```
jamesbraza, line 85: Why `re.?levance`?

tests/test_paperqa.py
```diff
@@ -375,64 +374,6 @@ def test_extract_score() -> None:
...
-        (
-            """Sure here is the json you asked for!
-
-    {
-    "example": "json"
-    }
-
-    Did you like it?""",
-            {"example": "json"},
-        ),
```
jamesbraza, line 390: Just looking through these cases, did you incorporate them into your new ones? I don't see the original cases still being present

jamesbraza (commented):
Getting close, just a few more comments

tests/test_paperqa.py
```diff
@@ -1471,3 +1412,198 @@ def partition_by_citation(t: Embeddable) -> int:
...
+
+
+class TestLLMParseJson:
+    """Tests for extracting JSON strings from LLM Response and ensuring proper formatting."""
+
+    @pytest.mark.parametrize(
+        "input_text",
+        [
+            pytest.param(
+                "<think> Thinking </think>"
```
foreverallama, line 1424: > Just looking through these cases, did you incorporate them into your new ones? I don't see the original cases still being present

Added those cases here (up to Line 1466). Made a few formatting changes for readability without changing the string content:
- Used \n for newlines to represent line breaks
- Changed words only to match the newly added cases. For e.g., `Sure here is the json you asked for!` changed to `I am here to help` and `"example": "json"` changed to `"summary": "Lorem Ipsum"`

paperqa/core.py
foreverallama: I had an edge case where the key was `re_levance-score`. On second thought the `score` part of the regex should work here. From the cases I collected, the keys contained at least one word out of `relevance` or `score`. 

Will update

tests/test_paperqa.py
foreverallama: [Addressed here](https://github.com/Future-House/paper-qa/pull/843#discussion_r1957820382)

jamesbraza (approved):
Nice work @foreverallama , appreciate the attention to detail and keep the contributions coming 🦾

PR #844, opened 2025-02-01
Benchmark for rag-qa-arena

JoaquinPolonuer:


paperqa/agents/search.py
```diff
@@ -325,6 +331,14 @@ async def _add_document() -> None:
...
             )
             raise
+        
+    async def commit(self) -> None:
+        """Commit all pending changes to the index."""
+        if self._writer:
+            self._writer.commit()
+            self._writer.wait_merging_threads()
+            self._searcher = None
+            self._writer = None
```
JoaquinPolonuer, line 341: I'm not sure if I should introduce this method or instead run this code inside the .save_index(). 
Probably the second one is a better idea.

paperqa/agents/search.py
```diff
@@ -461,6 +518,8 @@ async def process_file(
     settings: Settings,
     progress_bar_update: Callable[[], Any] | None = None,
 ) -> None:
+    global processed
```
JoaquinPolonuer, line 521: Probably not the best way to do this. Any suggestions?

paperqa/agents/search.py
```diff
@@ -295,10 +303,8 @@ async def add_document(
         async def _add_document() -> None:
             if not await self.filecheck(index_doc["file_location"], index_doc["body"]):
                 try:
-                    writer: IndexWriter = (await self.index).writer()
+                    writer = await self.writer
```
JoaquinPolonuer, line 306: This is for concurrency.
It seems that .writer() blocks other threads and so concurrency doesn't work.
At least, I've done some experiments with different concurrency numbers and all of them took the same time to index.
This seems to be a fix

paperqa/agents/search.py
```diff
@@ -399,10 +428,12 @@ async def query(
...
             s[1]
             for s in searcher.search(
-                index.parse_query(self.clean_query(query), query_fields), top_n
+                index.parse_query(self.clean_query(query), query_fields),
+                top_n,
+                offset=offset,
             ).hits
             if s[0] > min_score
-        ][offset : offset + top_n]
+        ]
```
JoaquinPolonuer, line 436: This is a bugfix. We were getting top_n and then starting at offset.

JoaquinPolonuer, line 436: This is a bugfix. We were getting top_n and then starting at offset.

jamesbraza: I think the change is good, but it's orthogonal to RAG QA arena.

Can you (1) make a unit test for this and (2) do it in a separate PR?

paperqa/settings.py
```diff
@@ -631,6 +631,20 @@ class Settings(BaseSettings):
...
+    pair_eval_llm_config: dict | None = Field(
+        default=None,
+        description=(
+            "Optional configuration for the pair_eval_llm model. More specifically, it's"
+            " a LiteLLM Router configuration to pass to LiteLLMModel, must have"
+            " `model_list` key (corresponding to model_list inputs here:"
+            " https://docs.litellm.ai/docs/routing), and can optionally include a"
+            " router_kwargs key with router kwargs as values."
+        ),
+    )
```
JoaquinPolonuer, line 647: Is it okay adding this to settings?

jamesbraza: The `Settings` should be general purpose. This `pair_eval_llm`/`LFRQAPairwiseEvalEnv` isn't used for general PaperQA, so no let's not add this here.

Let's just have callers pass this info to `LFRQAPairwiseEvalEnv` directly

paperqa/agents/search.py
JoaquinPolonuer: https://github.com/Future-House/paper-qa/pull/846

paperqa/types.py
```diff
@@ -615,7 +615,7 @@ def validate_all_fields(cls, data: Mapping[str, Any]) -> dict[str, Any]:
         data = cls.misc_string_cleaning(data)
         data = cls.inject_clean_doi_url_into_data(data)
         data = cls.add_preprint_journal_from_doi_if_missing(data)
-        data = cls.populate_bibtex_key_citation(data)
+        # data = cls.populate_bibtex_key_citation(data)
```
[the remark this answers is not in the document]
JoaquinPolonuer: Yes, probably we do need it. Just not for this use case. 
Maybe in the tutorial we could indicate to comment this line? It would seem strange though.
Or maybe we just leave it as it is. For indexing small proportions of the dataset this shouldn't be a problem.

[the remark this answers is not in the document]
jamesbraza: So if we're going to build an index and share it with the team/community, I prefer it be built right (and not have outstanding hydrations of BibTeX or deferred embeddings).

When you built the index, did you have this part commented out?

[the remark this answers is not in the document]
JoaquinPolonuer: Yes. But does that matter? The issue is that these documents are not actual papers, but just a few sentences. We are also hardcoding `doi` as the document id, because they don't have dois.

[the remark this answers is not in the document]
mskarlin: I think we just need a flag that controls whether this is populated or not, and we can include it as an attribute.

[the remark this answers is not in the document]
JoaquinPolonuer: Like a `populate_bibtex` in settings?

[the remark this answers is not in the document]
JoaquinPolonuer: Isn't it controversial to have a setting that if set to `true` breaks the settings? (I really don't know)

paperqa/agents/search.py
```diff
@@ -453,6 +496,21 @@ async def maybe_get_manifest(
...
+        return manifest_entry.model_dump()
+    if manifest_fallback_location in manifest:
+        manifest_entry: DocDetails = manifest[manifest_fallback_location]
+        return manifest_entry.model_dump()
+    return {}
+
+
+processed = 0
+
+
```
JoaquinPolonuer, line 513: Lets put this on a different PR

benchmark_results_eda.ipynb
mskarlin, line 1: since this is only for EDA, should we just cut this from being checked into the repo?

gradable.py
```diff
@@ -0,0 +1,44 @@
+import asyncio
```
mskarlin, line 1: let's move the grading and running scripts into a subdirectory within `docs` for this benchmark in particular.

paperqa/agents/search.py
```diff
@@ -425,7 +469,18 @@ async def maybe_get_manifest(
         try:
             async with await anyio.open_file(filename, mode="r") as file:
                 content = await file.read()
-            records = [DocDetails(**r) for r in csv.DictReader(content.splitlines())]
+
+            csv_content = list(csv.DictReader(content.splitlines()))
+            records = [
+                DocDetails(**r)
+                for r in tqdm(
```
mskarlin, line 476: I think we need to either make tqdm optional here or cut it, only because this will trigger via the CLI and it messes up the interface in that case.

PR #846, opened 2025-02-03
Fixed pagination issue in search.py and added corresponding test

JoaquinPolonuer:


tests/test_agents.py
```diff
@@ -1059,7 +1059,34 @@ async def test_clinical_tool_usage(agent_test_settings) -> None:
...
+    results2 = await index.query(
+        query="test", 
+        top_n=page_size,
+        offset=page_size
+    )
+    all_results = await index.query(
+        query="test",
+        top_n=2*page_size,
+        offset=0
+    )
```
jamesbraza, line 1085:
```suggestion
    page1_results = await index.query(
        query="test", 
        top_n=page_size
    )
    page2_results = await index.query(
        query="test", 
        top_n=page_size,
        offset=page_size
    )
    page1and2_results = await index.query(
        query="test",
        top_n=2*page_size
    )
```

Let's be a bit more explicit in the variable names here

I find "results1" is not really as intuitive as "page1_results?

tests/test_agents.py
```diff
@@ -1061,6 +1061,25 @@ async def test_clinical_tool_usage(agent_test_settings) -> None:
     ), "No clinical trials were put into contexts"
 
 
+@pytest.mark.asyncio
+async def test_search_pagination(agent_test_settings: Settings) -> None:
+    """Test that pagination works correctly in SearchIndex.query()."""
+    index = await get_directory_index(settings=agent_test_settings)
```
jamesbraza, line 1067: One other comment is, this index build in CI is a bit slow.

Can you do this:

1. Make an async `pytest.fixture` with `session` scope that does this
2. Use it in `test_search_pagination`, `test_deepcopy_env`

Then this PR is good to go

jamesbraza: Per in-person convo, `agent_test_settings` being not session scoped messes this up.

It's too much work to undo this for now, so we can handle this in a separate PR. YAGNI

jamesbraza (approved):
LGTM, nice work and thanks!

PR #847, opened 2025-02-04
Allow `agent_query` to Use an Existing Search Index Without Rebuilding

psl-schaefer:
##### My Use Case:
- I have a collection of several hundred papers that I want to index once (i.e. chunking PDFs and embedding them in a vector database)
- I use a Jupyter notebook to configure all PaperQA settings (see [tutorial](https://github.com/psl-schaefer/expert_system/blob/main/tutorial.ipynb)), allowing me to query these documents without rebuilding the index. To achieve this, I always use the same `IndexSettings` with a fixed `name` attribute.
- To build the index, I want to use a custom function ([build_search_index.py](https://github.com/psl-schaefer/expert_system/blob/main/src/build_search_index.py)) that does not rely on external APIs such as Semantic Scholar and Crossref. SemanticScholar API "key issuance is currently on pause due to high demand", and I do not require metadata from these sources since I already have all PDFs stored in Zotero.

##### My Solution:
- The only modification needed in the PaperQA codebase is adding a boolean argument `rebuild_index` to `agent_query` which is passed down to `run_agent`. In `run_agent` this argument is passed down to `get_directory_index` as the `build` parameter.
    - If `rebuild_index=True` (the default), the index is fully rebuilt.
    - If `rebuild_index=False`, we use the search index as previously build
- This change allows me to build the index once using my custom function ([build_search_index.py](https://github.com/psl-schaefer/expert_system/blob/main/src/build_search_index.py)).
- Before running any queries with `agent_query`, I first create the search index using this custom function.
- For a detailed overview of the workflow, see the [tutorial](https://github.com/psl-schaefer/expert_system/blob/main/tutorial.ipynb).

##### Also
- What do you think about adding such a `build_search_index` (see [build_search_index.py](https://github.com/psl-schaefer/expert_system/blob/main/src/build_search_index.py)) function that does not rely on external APIs to the PaperQA code base?

### DIFF
--- a/paperqa/agents/main.py
+++ b/paperqa/agents/main.py
@@ -55,6 +55,7 @@ async def agent_query(
     settings: Settings,
     docs: Docs | None = None,
     agent_type: str | type = DEFAULT_AGENT_TYPE,
+    rebuild_index: bool = True,
     **runner_kwargs,
 ) -> AnswerResponse:
     if docs is None:
@@ -67,7 +68,9 @@ async def agent_query(
         storage=SearchDocumentStorage.JSON_MODEL_DUMP,
     )
 
-    response = await run_agent(docs, query, settings, agent_type, **runner_kwargs)
+    response = await run_agent(
+        docs, query, settings, agent_type, rebuild_index, **runner_kwargs
+    )
     agent_logger.debug(f"agent_response: {response}")
 
     agent_logger.info(f"[bold blue]Answer: {response.session.answer}[/bold blue]")
@@ -92,6 +95,7 @@ async def run_agent(
     query: str | MultipleChoiceQuestion,
     settings: Settings,
     agent_type: str | type = DEFAULT_AGENT_TYPE,
+    rebuild_index: bool = True,
     **runner_kwargs,
 ) -> AnswerResponse:
     """
@@ -103,6 +107,7 @@ async def run_agent(
         settings: Settings to use.
         agent_type: Agent type (or fully qualified name to the type) to pass to
             AgentType.get_agent, or "fake" to TODOC.
+        rebuild_index: If True, we force the index to be rebuild.
         runner_kwargs: Keyword arguments to pass to the runner.
 
     Returns:
@@ -120,7 +125,7 @@ async def run_agent(
     # Build the index once here, and then all tools won't need to rebuild it
     # only build if the a search tool is requested
     if PaperSearch.TOOL_FN_NAME in (settings.agent.tool_names or DEFAULT_TOOL_NAMES):
-        await get_directory_index(settings=settings)
+        await get_directory_index(settings=settings, build=rebuild_index)
 
     if isinstance(agent_type, str) and agent_type.lower() == FAKE_AGENT_TYPE:
         session, agent_status = await run_fake_agent(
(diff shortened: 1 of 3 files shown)


paperqa/agents/main.py
```diff
@@ -120,7 +125,7 @@ async def run_agent(
     # Build the index once here, and then all tools won't need to rebuild it
     # only build if the a search tool is requested
     if PaperSearch.TOOL_FN_NAME in (settings.agent.tool_names or DEFAULT_TOOL_NAMES):
-        await get_directory_index(settings=settings)
+        await get_directory_index(settings=settings, build=force_index_rebuild)
```
jamesbraza, line 128: If I understand this change, basically you're calling `agent_query ` multiple times with the same `IndexSettings.name`. And are finding it will rebuild the index every time?

We made this `agent_query` flow for CLI users, hence why there is the colorized logging. Internally, we just directly work with the core `PaperQAEnvironment`.

That being said, I think this change makes sense, but with some tweaks

Let's do this:

1. Make a new setting in `AgentSettings` called `rebuild_index`, defaulted to `True`, with a `description`
2. Access that setting here, instead of adding keyword arguments to `run_agent`/`agent_query`

tests/test_agents.py
```diff
@@ -1080,6 +1080,19 @@ async def test_search_pagination(agent_test_settings: Settings) -> None:
     ), "Second page should match second slice of all results"
 
 
+@pytest.mark.asyncio
+async def test_agent_empty_index(agent_test_settings: Settings):
+    """Test that empty index and `rebuild_index=False` lead to a RuntimeError."""
+    agent_test_settings.agent = AgentSettings(index=IndexSettings()) # empty index
+    with pytest.raises(RuntimeError):
```
jamesbraza, line 1087: Can you add a `match` parameter to this `pytest.raises`

tests/test_agents.py
```diff
@@ -1080,6 +1080,19 @@ async def test_search_pagination(agent_test_settings: Settings) -> None:
     ), "Second page should match second slice of all results"
 
 
+@pytest.mark.asyncio
+async def test_agent_empty_index(agent_test_settings: Settings):
```
jamesbraza, line 1084:
```suggestion
async def test_empty_index_without_index_rebuild(agent_test_settings: Settings):
```

Just being more explicit here

paperqa/agents/main.py
psl-schaefer: > If I understand this change, basically you're calling agent_query  multiple times with the same IndexSettings.name. And are finding it will rebuild the index every time?

Yes this is exactly what I am doing. And this is exactly the issue.

>  Internally, we just directly work with the core PaperQAEnvironment

Is there also documentation for that?

> Let's do this:
> 
> Make a new setting in AgentSettings called rebuild_index, defaulted to True, with a description
> Access that setting here, instead of adding keyword arguments to run_agent/agent_query

That sounds good, I will update the PR tomorrow, and also include a regex in the test.

jamesbraza: > Is there also documentation for that?

Yeah PaperQA's docs are quite lacking, there's not nearly enough. In our current docs, probably in https://github.com/Future-House/paper-qa#running-on-litqa-v2. It uses an environments framework designed for training agents to use it called aviary ([paper](https://arxiv.org/abs/2412.21154), [code](https://github.com/Future-House/aviary)), if you are curious.

Feel free to just ignore this part, I think use of `run_agent` or `agent_query` is quite fine too.

jamesbraza:
Thank @psl-schaefer, appreciate the contribution! Feel free to keep em coming :)

PR #848, opened 2025-02-04
Added the get_manifest_kwargs function to get more than doi and title…

JoaquinPolonuer:
… from manifest if available


tests/test_agents.py
```diff
@@ -242,6 +242,9 @@ async def test_get_directory_index_w_manifest(agent_test_settings: Settings) ->
 
         results = await index.query(query="who is Frederick Bates?")
         top_result = next(iter(results[0].docs.values()))
+        
+        # NOTE: This is failing because now the dockey comes from the manifest.
+        # In the doctdetails object we are setting the dockey to the doc_id
         assert top_result.dockey == md5sum(abs_paper_dir / "bates.txt")
```
JoaquinPolonuer, line 248: This test is failing, I will explain why.

When we build the manifest we do 

`records = [DocDetails(**r) for r in csv.DictReader(content.splitlines())]`

and when the DocDetails is constructed, we do

data["dockey"] = data["doc_id"]

So now we are getting everything we can from the manifest (that's the purpose of this PR) but then  `dockey = md5sum(path)` never runs in docs.py

That's why, in the test, the assertion `top_result.dockey == md5sum(abs_paper_dir / "bates.txt")` is false

paperqa/docs.py
```diff
@@ -280,6 +280,9 @@ async def aadd(  # noqa: PLR0912
         parse_config = all_settings.parsing
         if dockey is None:
             # md5 sum of file contents (not path!)
+
+            # NOTE: test_get_directory_index_w_manifest fails because now we are getting
```
jamesbraza, line 284: Let's avoid adding mentions of the external test suite into the core package

What makes you add this comment here? Is the test still failing?

tests/test_agents.py
```diff
@@ -242,6 +242,9 @@ async def test_get_directory_index_w_manifest(agent_test_settings: Settings) ->
 
         results = await index.query(query="who is Frederick Bates?")
         top_result = next(iter(results[0].docs.values()))
+
+        # NOTE: This is failing because now the dockey comes from the manifest.
```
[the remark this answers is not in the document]
JoaquinPolonuer: Yes, I'll explain everything here:

The manifest doesn't have a `dockey` column. But before building the index, we call the `maybe_get_manifest_function`.

That function has this snippet
```
            records = [DocDetails(**r) for r in csv.DictReader(content.splitlines())]
            file_loc_to_records = {
                str(r.file_location): r for r in records if r.file_location
            }
```
so the manifest that gets to the `process_file` function has everything that the `stub_data/stub_manifest.csv` file has + more things that come from the DocDetails constructor, where the `dockey` is populated with `doc_id`.

So now, we pass 
```
            manifest_kwargs = get_manifest_kwargs(
                manifest, manifest_fallback_location, file_location
            )
```
which has a `dockey` that was populated in the DocDetails constructor, that is why
```
        if dockey is None:
            # md5 sum of file contents (not path!)
            dockey = md5sum(path)
```
is not executed and then the test fails.

I hope I was clear now :)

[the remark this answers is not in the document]
jamesbraza: Got it, thanks! Will re-review now

paperqa/agents/search.py
```diff
@@ -455,6 +455,18 @@ async def maybe_get_manifest(
 FAILED_DOCUMENT_ADD_ID = "ERROR"
 
 
+def get_manifest_kwargs(
+    manifest: dict[str, Any], manifest_fallback_location: str, file_location: str
+) -> dict[str, Any]:
```
jamesbraza, line 460: Few comments:

1. Can you move this to be next to `maybe_get_manifest` in the file?
2. Can you put `file_location` first in the args, since it's primary over "fallback location"
3. Can you rename to something more explicit, like `fetch_kwargs_from_manifest`

paperqa/agents/search.py
```diff
@@ -478,23 +490,17 @@ async def process_file(
...
 
-            doi, title = None, None
-            if file_location in manifest:
-                manifest_entry = manifest[file_location]
-                doi, title = manifest_entry.doi, manifest_entry.title
-            elif manifest_fallback_location in manifest:
-                # Perhaps manifest used the opposite pathing scheme
-                manifest_entry = manifest[manifest_fallback_location]
-                doi, title = manifest_entry.doi, manifest_entry.title
+            manifest_kwargs = get_manifest_kwargs(
```
jamesbraza, line 493: If I recall, the point of this is to allow for no BibTeX in `DocDetails`.

Can you add or expand a unit test to check for this particular capability?

JoaquinPolonuer: The point is actually not having to call an llm to get the `citation`.
As it was before, we were only getting `doi` and `title` from manifest, so when doing `tmp_docs.aadd` this would happen
```
if citation is None:
    # Peek first chunk
    texts = read_doc(
        path,
        Doc(docname="", citation="", dockey=dockey),  # Fake doc
        chunk_chars=parse_config.chunk_size,
        overlap=parse_config.overlap,
        page_size_limit=parse_config.page_size_limit,
    )
    if not texts:
        raise ValueError(f"Could not read document {path}. Is it empty?")
    result = await llm_model.run_prompt(
        prompt=parse_config.citation_prompt,
        data={"text": texts[0].text},
        system_prompt=None,  # skip system because it's too hesitant to answer
    )
```

I think another option would have been to get doi, title, and citation from manifest, but talking to mike we decided to make this more general change.

paperqa/clients/__init__.py
```diff
@@ -201,7 +201,7 @@ async def upgrade_doc_to_doc_details(self, doc: Doc, **kwargs) -> DocDetails:
         # note we have some extra fields which may have come from reading the doc text,
         # but aren't in the doc object, we add them here too.
         extra_fields = {
-            k: v for k, v in kwargs.items() if k in {"title", "authors", "doi"}
+            k: v for k, v in kwargs.items() if k in set(DocDetails.model_fields)
         }
```
JoaquinPolonuer, line 205: @mskarlin this was necessary for the dockey to be equal to doc_id.
Do you know why it was like that previously and if its ok to change it?

paperqa/types.py
```diff
@@ -411,6 +411,8 @@ def lowercase_doi_and_populate_doc_id(cls, data: dict[str, Any]) -> dict[str, An
             data["doc_id"] = encode_id(uuid4())
 
         if data.get("overwrite_fields_from_metadata", True):
+            # Here we are setting the dockey to the doc_id
+            # but then we override it with md5sum(path) in the docs.py
```
jamesbraza, line 415: Is this still relevant?

tests/test_agents.py
```diff
@@ -242,7 +242,10 @@ async def test_get_directory_index_w_manifest(agent_test_settings: Settings) ->
 
         results = await index.query(query="who is Frederick Bates?")
         top_result = next(iter(results[0].docs.values()))
-        assert top_result.dockey == md5sum(abs_paper_dir / "bates.txt")
+
+        # note: we get every possible field from the manifest constructed in maybe_get_manifest,
+        # so the dockey will now be equal to the doc_id.
```
jamesbraza, line 247: This comment doesn't make sense because `doc_id`/`doc_key` isn't in the manifest

Let's modify this comment to reflect that

```suggestion
        # note: we get every possible field from the manifest constructed in maybe_get_manifest,
        # and then DocDetails construction sets the dockey to the doc_id.
```

tests/test_agents.py
```diff
@@ -242,7 +242,10 @@ async def test_get_directory_index_w_manifest(agent_test_settings: Settings) ->
 
         results = await index.query(query="who is Frederick Bates?")
         top_result = next(iter(results[0].docs.values()))
-        assert top_result.dockey == md5sum(abs_paper_dir / "bates.txt")
+
+        # note: we get every possible field from the manifest constructed in maybe_get_manifest,
```
jamesbraza, line 246: Let's get another test (or a `subtests.test` subtest) with a second manifest that leads to us not doing citation inference, something resembling the below

```csv
file_location,doi,title,citation
"bates.txt",,"Frederick Bates (Wikipedia article)",""
```

paperqa/types.py
```diff
@@ -590,19 +590,16 @@ def populate_bibtex_key_citation(  # noqa: PLR0912
                 data["bibtex"] = BibliographyData(
                     entries={data["key"]: new_entry}
                 ).to_string("bibtex")
-                # clear out the citation, since it will be regenerated
-                if data.get("overwrite_fields_from_metadata", True):
-                    data["citation"] = None
```
jamesbraza, line 595: Can we keep this but have `if data.get("overwrite_fields_from_metadata", True) and data.get("citation") is not None:`

I don't want to clobber this behavior in its entirety

JoaquinPolonuer: But if we do `data.get("citation") is not None`, either
1. The citation was None and we didn't set it to None
2. The citation had a value but we override it 
Anyway tests are probably failing because I removed that line :(

paperqa/types.py
```diff
@@ -52,6 +56,9 @@ class Doc(Embeddable):
             "flag to overwrite fields from metadata when upgrading to a DocDetails"
         ),
     )
+    overwrite_citation_from_metadata: bool = Field(
+        default=True, description=("flag to overwrite the citation from metadata")
```
jamesbraza, line 60: Few comments here:
- Can you remove the extra parens
- Can you clarify what you mean by "metadata" here? I would maybe say "manifest" not "metadata"
- Can you document why the default is `True`? "Defaults to True because ..."

paperqa/types.py
```diff
@@ -591,32 +598,43 @@ def populate_bibtex_key_citation(  # noqa: PLR0912
...
+            data |= {"overwrite_citation_from_metadata": False}
+
+        data = cls.lowercase_doi_and_populate_doc_id(data)
         data = cls.lowercase_doi_and_populate_doc_id(data)
         data = cls.remove_invalid_authors(data)
         data = cls.misc_string_cleaning(data)
         data = cls.inject_clean_doi_url_into_data(data)
         data = cls.add_preprint_journal_from_doi_if_missing(data)
         data = cls.populate_bibtex_key_citation(data)
+
```
jamesbraza, line 637: [PR nit] try to be conservative when adding one line whitespaces, it adds extra diff and the whitespace can break locality

If you do like this whitespace, feel free to keep it, more just sharing is all

paperqa/types.py
```diff
@@ -40,6 +40,10 @@
 logger = logging.getLogger(__name__)
 
 
+ENV_VAR_MATCH: Collection[str] = {"1", "true"}
+ENV_VAR_MISMATCH: Collection[str] = {"0", "false"}
```
jamesbraza, line 44: So we are using these now for both environment variables and manifest fields.

Do you mind renaming these to `VAR_MATCH_LOOKUP` and `VAR_MISMATCH_LOOKUP`, mainly since they're more general than just environment variables

paperqa/types.py
JoaquinPolonuer: So I created this flag because we already had `overwrite_fields_from_metadata`.

`overwrite_fields_from_metadata` would overwrite `doc_id`, `dockey` and `citation` (and maybe another field I don't remember) but I wanted to separate `citation` from that.

In some way I copied the  `overwrite_fields_from_metadata` flag and its default settings. Do you want me to rename this flag to `overwrite_fields_from_manifest`?

@mskarlin wtyt?

jamesbraza: Oh I see. Let's consolidate them into one field:

```python
    overwrite_fields_from_metadata: Collection[str] = Field(
        default_factory=lambda: {"key": "docname", "doc_id": "dockey"},
        description=(
            "(fill in)"
        ),
    )
```

Then for RAG QA and any relevant tests, we'd do `overwrite_fields_from_metadata={"key": "docname", "doc_id": "dockey", "citation"}`

JoaquinPolonuer: @mskarlin suggested the same thing, and I believe is the best approach.
I think the problem with that is that you would like to specify that in a `.csv` right? So specifying a set can be complicated.

paperqa/types.py
```diff
@@ -502,10 +515,15 @@ def overwrite_docname_dockey_for_compatibility_w_doc(
     ) -> dict[str, Any]:
         """Overwrite fields from metadata if specified."""
         overwrite_fields = {"key": "docname", "doc_id": "dockey"}
-        if data.get("overwrite_fields_from_metadata", True):
-            for field, old_field in overwrite_fields.items():
-                if data.get(field):
-                    data[old_field] = data[field]
+        for field, old_field in overwrite_fields.items():
```
jamesbraza, line 518: Can we have this be `DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA`, and remove `"citation"` from `DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA`

paperqa/types.py
```diff
@@ -40,17 +40,27 @@
 logger = logging.getLogger(__name__)
 
 
+VAR_MATCH_LOOKUP: Collection[str] = {"1", "true"}
+VAR_MISMATCH_LOOKUP: Collection[str] = {"0", "false"}
+
+DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA: set[str] = {
```
jamesbraza, line 46: Let's type hint `Collection[str]` so it's read-only

paperqa/types.py
JoaquinPolonuer: The problem is that many test rely on the fact that `"citation"` is in that set

paperqa/types.py
JoaquinPolonuer: To do that it's needed to set

`model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)`

or else I get 

```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for collections.abc.Collection[str]. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
```

Is that ok? Or we leave it as a `set`

paperqa/types.py
jamesbraza: Can't those tests then specify `DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA | {"citation"}` into the `Settings` then?

JoaquinPolonuer: We can use an intersection...

paperqa/types.py
jamesbraza: Then you'll need to do:

```python
default_factory=lambda: set(DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA)
```

paperqa/types.py
JoaquinPolonuer: Yes but I think we want to have the citation overwritten by default.
We are replacing the `overwrite_fields_from_metadata` here and that flag was used to overwrite the `citation`.
I mean, this would be changing a default behavior.

JoaquinPolonuer: Other approach would be

```python
"""Overwrite fields from metadata if specified."""
overwrite_fields = {"key": "docname", "doc_id": "dockey"}
fields_to_overwrite = data.get(
    "fields_to_overwrite_from_metadata",
    DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA,
)
for field in (overwrite_fields.keys() & fields_to_overwrite):
    if data.get(field):
        data[overwrite_fields[field]] = data[field]
return data
```

I think this makes sense

jamesbraza: Maybe I am not understanding, but why do we want to have the citation overwritten by default? Isn't that like a RAG QA manifest specific thing?

I think the best default for users is to infer citation

jamesbraza: Okay we talked in-person, I understand what you're saying here. Defer to your best judgement here, thanks for talking me through it 👌

jamesbraza (approved):
LGTM, nice!

PR #849, opened 2025-02-04
Support parallel indexing

JoaquinPolonuer:


paperqa/agents/search.py
```diff
@@ -295,10 +313,10 @@ async def add_document(
         async def _add_document() -> None:
             if not await self.filecheck(index_doc["file_location"], index_doc["body"]):
                 try:
-                    writer: IndexWriter = (await self.index).writer()
+                    writer: IndexWriter = await self.writer
                     writer.add_document(Document.from_dict(index_doc))  # type: ignore[call-arg]
-                    writer.commit()
-                    writer.wait_merging_threads()
+                    # writer.commit()
+                    # writer.wait_merging_threads()
```
JoaquinPolonuer, line 320: Will remove later

tests/test_agents.py
```diff
@@ -161,10 +161,11 @@ async def crashing_aadd(*args, **kwargs) -> str | None:
             ) as mock_aadd,
         ):
             index = await get_directory_index(settings=agent_test_settings)
+
+    assert len(await index.index_files) == num_source_files
     assert (
-        mock_aadd.await_count <= crash_threshold
+        mock_aadd.await_count != num_source_files
     ), "Should have been able to resume build"
```
JoaquinPolonuer, line 168: Probably the most controversial thing on this PR.
As we commit in batchs of 2, we may be building some documents again, but I change the test to see if

1. We end up indexing all docs
2. We don't reindex the whole thing

It's weaker than before, but I think it should be enough

.gitignore
```diff
@@ -308,3 +308,5 @@ tests/example2.*
 
 # Client data
 paperqa/clients/client_data/retractions.csv
+
+rag-qa-benchmarking/
```
jamesbraza, line 312: Let's save this for a separate RAG QA arena PR

main.py
```diff
@@ -0,0 +1,25 @@
+from paperqa import Settings, ask
+
+settings = Settings()
+settings.agent.index.name = "lfrqa_science_index_0.01"
+settings.agent.index.paper_directory = (
+    "rag-qa-benchmarking/lfrqa/science_docs_for_paperqa"
+)
+settings.agent.index.index_directory = (
+    "rag-qa-benchmarking/lfrqa/science_docs_for_paperqa_index"
```
jamesbraza, line 9: Let's save this for a separate RAG QA arena PR

paperqa/settings.py
```diff
@@ -405,7 +405,7 @@ class IndexSettings(BaseModel):
         description="Whether to recurse into subdirectories when indexing sources.",
     )
     concurrency: int = Field(
-        default=5,  # low default for folks without S2/Crossref keys
+        default=2,  # low default for folks without S2/Crossref keys
```
jamesbraza, line 408: Why are we changing the default in this PR?

paperqa/agents/search.py
```diff
@@ -525,9 +556,13 @@ async def process_file(
                 },
                 document=tmp_docs,
             )
-            # Save so we can resume the build without rebuilding this file if a
-            # separate process_file invocation leads to a segfault or crash
-            await search_index.save_index()
+
+            processed += 1
+            if processed == settings.agent.index.concurrency:
```
jamesbraza, line 561: let's make a separate index setting `batch_size` or something like this, and default it to `1`

paperqa/agents/search.py
```diff
@@ -326,6 +344,14 @@ async def _add_document() -> None:
             )
             raise
 
+    async def commit(self) -> None:
+        """Commit all pending changes to the index."""
```
jamesbraza, line 348:
```suggestion
    async def checkpoint(self) -> None:
        """Commit all pending changes, merge threads, and reset."""
```

This isn't just a `commit` any more, let's call it `checkpoint`

paperqa/agents/search.py
```diff
@@ -326,6 +344,14 @@ async def _add_document() -> None:
...
+        if self._writer:
+            self._writer.commit()
+            self._writer.wait_merging_threads()
+            self._searcher = None
```
jamesbraza, line 352: Can you add a comment why we are wiping the searcher?

paperqa/agents/search.py
```diff
@@ -270,6 +278,16 @@ async def mark_failed_document(self, path: str | os.PathLike) -> None:
         (await self.index_files)[str(path)] = FAILED_DOCUMENT_ADD_ID
         self.changed = True
 
+    async def release_lock(self) -> None:
```
jamesbraza, line 281: So this feels like a hack that we have added due to a lack of understanding the locks

Can we understand why we are encountering deadlocks, and do a more precise fix

paperqa/settings.py
JoaquinPolonuer: When I ran the test with concurrency set to 4 I got 429 from semantic scholar.

jamesbraza: Then let's make the particular test configure a lower value, without changing the default

paperqa/agents/search.py
JoaquinPolonuer: No need to wipe the searcher actually

paperqa/agents/search.py
```diff
@@ -546,11 +544,11 @@ async def process_file(
                 document=tmp_docs,
             )
 
-            processed += 1
-            if processed == settings.agent.index.batch_size:
+            processed_counter["n_processed_files"] += 1
```
JoaquinPolonuer, line 547: Is this the right way to do it?

jamesbraza: Yeah it's right, but can we call it something like `batched_save_counter` to be more explicit?

paperqa/agents/search.py
```diff
@@ -525,9 +543,13 @@ async def process_file(
...
                 document=tmp_docs,
             )
-            # Save so we can resume the build without rebuilding this file if a
-            # separate process_file invocation leads to a segfault or crash
-            await search_index.save_index()
+
+            processed_counter["n_processed_files"] += 1
+            if processed_counter["n_processed_files"] == settings.agent.index.batch_size:
+                await search_index.save_index()
+                print(f"Saved index after processing {processed_counter['n_processed_files']} files.")
```
jamesbraza: Can you remove the `print`

paperqa/agents/search.py
```diff
@@ -496,13 +514,15 @@ async def process_file(
                     fields=["title", "author", "journal", "year"],
                     settings=settings,
                 )
-            except (ValueError, ImpossibleParsingError):
+            except Exception:
+                # We handle any exception here because we want to save_index so we
+                # 1. can resume the build without rebuilding this file if a separate
```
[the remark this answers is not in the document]
JoaquinPolonuer: Hahah there was a 2. idk were it went.

paperqa/agents/search.py
```diff
@@ -674,6 +699,7 @@ async def get_directory_index(  # noqa: PLR0912
     )
     with progress_bar:
         async with anyio.create_task_group() as tg:
+            processed_counter: Counter = Counter()
```
jamesbraza, line 702:
```suggestion
            processed_counter = Counter()
```

No need to repeat the type hint, Python is not strongly typed

paperqa/readers.py
```diff
@@ -308,65 +309,74 @@ def read_doc(
...
     # next chunk the parsed text
 
-    # check if chunk is 0 (no chunking)
+    # check if chunk_chars is 0 (no chunking)
     if chunk_chars == 0:
+        # Offload reduce_content to a thread as well
         chunked_text = [
-            Text(text=parsed_text.reduce_content(), name=doc.docname, doc=doc)
+            Text(
+                text=await asyncio.to_thread(parsed_text.reduce_content),
```
jamesbraza, line 350: Let's only use `asyncio.to_thread` on code code with I/O

paperqa/settings.py
```diff
@@ -408,6 +408,10 @@ class IndexSettings(BaseModel):
         default=5,  # low default for folks without S2/Crossref keys
         description="Number of concurrent filesystem reads for indexing",
     )
+    batch_size: int = Field(
+        default=1,
```
jamesbraza, line 412: Let's at `ge=1` to be cool

tests/test_agents.py
```diff
@@ -161,10 +160,11 @@ async def crashing_aadd(*args, **kwargs) -> str | None:
             ) as mock_aadd,
         ):
             index = await get_directory_index(settings=agent_test_settings)
+
+    assert len(await index.index_files) == num_source_files
     assert (
-        mock_aadd.await_count <= crash_threshold
+        mock_aadd.await_count != num_source_files
```
jamesbraza, line 166: Can we do `< num_source_files` here? `!=` is less specific.

Also, can you beef up the comment `"Should have been able to resume build"` to better reflect the assertion?

paperqa/readers.py
```diff
@@ -308,65 +309,74 @@ def read_doc(
...
-        page_size_limit: optional limit on the number of characters per page
+    Offloads blocking I/O and CPU operations to a thread pool.
     """
     str_path = str(path)
-    parsed_text = None
 
     # start with parsing -- users may want to store this separately
     if str_path.endswith(".pdf"):
-        parsed_text = parse_pdf_to_pages(path, page_size_limit=page_size_limit)
+        parsed_text = await asyncio.to_thread(
```
jamesbraza, line 320: Can you add `TODO` comments next to `parse_pdf_to_pages` and `parse_text` to make them `async`

tests/test_agents.py
```diff
@@ -1103,3 +1103,29 @@ async def test_continuation(self) -> None:
...
         assert len(state.docs.docs) > trial_count, "Search was unable to continue"
+
+
+@pytest.mark.asyncio
+async def test_index_build_concurrency(agent_test_settings: Settings) -> None:
+
+    high_concurrency_settings = agent_test_settings.model_copy(deep=True)
+    high_concurrency_settings.agent.index.name = "high_concurrency"
+    high_concurrency_settings.agent.index.concurrency = 3
+    high_concurrency_settings.agent.index.batch_size = 3
```
jamesbraza, line 1114: Do we really need `batch_size = 3` in this test? I think `concurrency = 3` alone should be good enough

paperqa/agents/search.py
JoaquinPolonuer: I think this was because of the lint

jamesbraza: Ah I see. Just to be precise, it's not really "lint", it's the type checker `mypy` (which would be `pre-commit`)

I think then let's do this

```suggestion
            processed_counter: Counter[str] = Counter()
```

The reason is it's telling `mypy` what type to expect in the `Counter`: https://github.com/python/typeshed/blob/aac4394eb29d86797628b57d6f01f5e17f5ff83f/stdlib/collections/__init__.pyi#L277

tests/test_agents.py
JoaquinPolonuer: It seems so, but I don't really understand why.
I mean, isn't commiting every time the same as we did before?

paperqa/agents/search.py
```diff
@@ -461,8 +476,10 @@ async def process_file(
     manifest: dict[str, Any],
     semaphore: anyio.Semaphore,
     settings: Settings,
+    processed_counter: Counter,
```
jamesbraza, line 479:
```suggestion
    processed_counter: Counter[str],
```

paperqa/readers.py
```diff
@@ -308,10 +309,9 @@ def read_doc(
     overlap: int = 100,
     page_size_limit: int | None = None,
 ) -> list[Text] | ParsedText | tuple[list[Text], ParsedMetadata]:
-    """Parse a document and split into chunks.
+    """Parse a document and split into chunks without blocking the event loop.
```
jamesbraza, line 312: Let's revert this "without blocking the event loop"

This is really an implementation detail that is innate to all `async` methods, we don't need to document it in our docstring

paperqa/agents/search.py
JoaquinPolonuer: image

paperqa/agents/search.py
```diff
@@ -674,6 +699,7 @@ async def get_directory_index(  # noqa: PLR0912
     )
     with progress_bar:
         async with anyio.create_task_group() as tg:
+            processed_counter = Counter()
```
jamesbraza, line 702:
```suggestion
            processed_counter: Counter[str] = Counter()
```

When `mypy` makes these requests, it's because `mypy` detects there's a risk of a type error, so it wants to be sure

tests/test_agents.py
```diff
@@ -1103,3 +1103,52 @@ async def test_continuation(self) -> None:
...
+    with patch.object(
+        SearchIndex,
+        "save_index",
+        side_effect=high_batch_size_increment_save_index_count,
+    ):
+        start_time = time.perf_counter()
+        await get_directory_index(settings=high_concurrency_settings)
+        high_concurrency_duration = time.perf_counter() - start_time
```
jamesbraza, line 1127:
```suggestion
    high_concurrency_settings = agent_test_settings.model_copy(deep=True)
    high_concurrency_settings.agent.index.name = "high_concurrency"
    high_concurrency_settings.agent.index.concurrency = 3
    high_concurrency_settings.agent.index.batch_size = 3
    with patch.object(
        SearchIndex,
        "save_index",
        side_effect=SearchIndex.save_index,
        autospec=True
    ) as mock_save_index:
        start_time = time.perf_counter()
        await get_directory_index(settings=high_concurrency_settings)
        high_concurrency_duration = time.perf_counter() - start_time
    high_batch_size_save_index_count = mock_save_index.awaited_count
```

The pass-through method is less code and more full stack (includes saving in the delta time)

paperqa/agents/search.py
```diff
@@ -326,6 +333,13 @@ async def _add_document() -> None:
             )
             raise
 
+    async def checkpoint(self) -> None:
```
jamesbraza, line 336: Can you call `checkpoint` in `delete_document` too

tests/test_agents.py
```diff
@@ -140,7 +140,6 @@ async def crashing_aadd(*args, **kwargs) -> str | None:
 
     # 1. Try to build an index, and crash halfway through
     with (
-        pytest.raises(ExceptionGroup, match="unhandled"),
```
jamesbraza, line 143: Let's bring this one back, because we need some way of knowing failure in the case of an API outage.

Also, let's make `crashing_aadd` more clear why it's there:

```python
# This could be something like a 429 from OpenAI
raise RuntimeError("Unexpected crash.")
```

paperqa/agents/search.py
JoaquinPolonuer: Yes, but for that `delete_document` will have to be async and not be a static method, is that ok?

jamesbraza: Oh yeah `delete_document` doesn't need to be a `staticmethod`

paperqa/agents/search.py
```diff
@@ -326,19 +333,24 @@ async def _add_document() -> None:
             )
             raise
 
-    @staticmethod
+    async def checkpoint(self) -> None:
+        """Commit all pending changes, merge threads, and reset."""
```
jamesbraza, line 337: "and reset" -- this is vague, can you expand the docstring a bit

paperqa/agents/search.py
```diff
@@ -326,19 +333,24 @@ async def _add_document() -> None:
...
+        if self._writer:
+            self._writer.commit()
+            self._writer.wait_merging_threads()
+            self._writer = None
```
jamesbraza, line 341: I am hesitant about the `if not self._writer` case where `commit` does nothing, it feels like a footgun.

Can we do something like `if not self._writer: raise RuntimeError("Please do XYZ before calling commit.")`

Or better yet, can we redesign the control flows so this case does not exist?

There is a design philosophy: the software works, because it must work

paperqa/agents/search.py
```diff
@@ -496,16 +511,21 @@ async def process_file(
...
                     f"Error parsing {file_location}, skipping index for this file."
                 )
                 await search_index.mark_failed_document(file_location)
-                # Save so we can resume the build without rebuilding this file if a
-                # separate process_file invocation leads to a segfault or crash
                 await search_index.save_index()
                 if progress_bar_update:
                     progress_bar_update()
+
+                if not isinstance(e, ValueError | ImpossibleParsingError):
```
jamesbraza, line 527: Can you attempt to document each of these with a code comment:

```python
# We silence ValueError because ABC
# We silence ImpossibleParsingError because XYZ
```

JoaquinPolonuer: I honestly don't know why we chose those two, do you know?

jamesbraza: It's because they are considered both (1) non-dangerous and (2) non-retryable indexing failures:
- `ImpossibleParsingError`: our parser layer has failed, so we can never succeed. Thus we mark it as a failure and continue
- `ValueError`: I forget about this one, maybe just put `TODOC` for this one

paperqa/agents/search.py
JoaquinPolonuer: I think we actually don't need it at all.
Either we `save_index` instantly after adding a document or we use the `changed` attribute

paperqa/agents/search.py
```diff
@@ -326,19 +333,24 @@ async def _add_document() -> None:
             )
             raise
 
-    @staticmethod
+    async def checkpoint(self) -> None:
+        """Commit all pending changes, merge threads, and reset writer, as writers are not valid after commit."""
+        writer: IndexWriter = await self.writer
+        writer.commit()
+        writer.wait_merging_threads()
+        self._writer = None
```
jamesbraza, line 341: This pattern here where you have:

- Handling of `_writer` in `self.writer`
- Use the public `self.writer` here
- Also assign `_writer`

It's awkward and breaks encapsulation. Can you revisit this logic a bit? Try to think of ways you can avoid manipulating both private and public stuff here

paperqa/agents/search.py
```diff
@@ -496,16 +511,24 @@ async def process_file(
...
-                # Save so we can resume the build without rebuilding this file if a
-                # separate process_file invocation leads to a segfault or crash
                 await search_index.save_index()
                 if progress_bar_update:
                     progress_bar_update()
+
+                # TODOC: We silence ValueError because ...
+                # We silence ImpossibleParsingError because we won't be able to
+                # parse the file if it failed one time
+                if not isinstance(e, ValueError | ImpossibleParsingError):
```
jamesbraza, line 530:
```suggestion
                if not isinstance(e, ValueError | ImpossibleParsingError):
                    # ImpossibleParsingError: parsing failure, don't retry
                    # ValueError: TODOC
```

[style nit] comments go to the right of or inside `if`, not outside

paperqa/agents/search.py
JoaquinPolonuer: We can do 
```python
        self._writer.commit()
        self._writer.wait_merging_threads()
        self._writer = None
```
As there has to be an active writer to call this method.

JoaquinPolonuer: The only problem is that the pre-commit will fail because `self._writer` can be have type `None`

jamesbraza (approved):
LGTM, excellent work here and thanks for working through all the comments

JoaquinPolonuer:
Thank you!!

PR #856, opened 2025-02-11
Move doc details line

JoaquinPolonuer:


paperqa/agents/search.py
```diff
@@ -429,35 +429,36 @@ async def query(
 
 def fetch_kwargs_from_manifest(
     file_location: str, manifest: dict[str, Any], manifest_fallback_location: str
-) -> dict[str, Any]:
-    manifest_entry: DocDetails | None = manifest.get(file_location) or manifest.get(
+) -> dict[Any, Any]:
```
jamesbraza, line 432: Why do we need to do this? `dict[str, Any]` should be compatible with `model_dump`

paperqa/agents/search.py
```diff
@@ -429,35 +429,36 @@ async def query(
...
     if manifest_entry:
-        return manifest_entry.model_dump()
+        return DocDetails(**manifest_entry).model_dump()
     return {}
 
 
 async def maybe_get_manifest(
     filename: anyio.Path | None = None,
-) -> dict[str, DocDetails]:
+) -> dict[Any, dict[Any, Any]]:
```
jamesbraza, line 443: I think this should just be `dict[str, dict]` but more preferably `dict[str, dict[str, Any]]`

paperqa/agents/search.py
```diff
@@ -430,34 +430,35 @@ async def query(
 def fetch_kwargs_from_manifest(
     file_location: str, manifest: dict[str, Any], manifest_fallback_location: str
 ) -> dict[str, Any]:
-    manifest_entry: DocDetails | None = manifest.get(file_location) or manifest.get(
+    manifest_entry: dict[Any, Any] | None = manifest.get(file_location) or manifest.get(
```
jamesbraza, line 433:
```suggestion
    manifest_entry: dict[str, Any] | None = manifest.get(file_location) or manifest.get(
```

paperqa/agents/task.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         self._rewards = rewards
 
     async def validate_sources(
-        self, manifest_or_index: dict[str, DocDetails] | SearchIndex | None = None
+        self, manifest_or_index: dict[Any, dict[Any, Any]] | SearchIndex | None = None
```
jamesbraza, line 99:
```suggestion
        self, manifest_or_index: dict[str, dict[str, Any]] | SearchIndex | None = None
```

jamesbraza (approved):

PR #858, opened 2025-02-11
added escape_backlashes function with test

JoaquinPolonuer:


tests/test_paperqa.py
```diff
@@ -424,6 +424,29 @@ def test_llm_parse_json_newlines() -> None:
     }
 
 
+def test_llm_parse_json_double_backslashes() -> None:
```
jamesbraza, line 427: We already have `test_llm_parse_json`, can you just add a case to the `pytest.mark.parameterize` there?

JoaquinPolonuer: The problem is that all the tests there are of the type 
python`llm_parse_json(example) == {"example": "json"}`

jamesbraza: I see. There's only one case there so far, so can you just add an `expected: dict[str, Any]` to the parameterization

JoaquinPolonuer: So it think we either

1. Create this as a separate test
2. Instead of replacing // for //// (which makes the string // appear in the parsed json), we remove the //

JoaquinPolonuer: Ah nice, I didn't know that. Thanks

jamesbraza (approved):

PR #864, opened 2025-02-13
Pin fh-llm-client

maykcaldas:
We are pinning fh-llm-client to 0.1.0 so that its next release won't be imported. This is to avoid breaking changes for now

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "PyMuPDF>=1.24.12",  # For pymupdf.set_messages addition
     "aiohttp>=3.10.6",  # TODO: remove in favor of httpx, pin for aiohttp.ClientConnectionResetError
     "anyio",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "PyMuPDF>=1.24.12",  # For pymupdf.set_messages addition
     "aiohttp>=3.10.6",  # TODO: remove in favor of httpx, pin for aiohttp.ClientConnectionResetError
     "anyio",
-    "fh-llm-client>=0.0.11",  # Pin for LLMModel.run_prompt allowing Iterable
+    "fh-llm-client<0.1.0",  # Pin to avoid breaking changes until lfrqa is migrated
```
jamesbraza, line 28:
```suggestion
    "fh-llm-client<0.1.0",  # Pin to avoid breaking changes
```

Let's avoid being LFRQA specific here, because really the whole repo would be broken

jamesbraza (approved):

Issue #796, opened 2025-01-08
`gemini-1.5-flash` can't call tools without empty arguments


jamesbraza:
Running the below code on Jan 8th 2025 with Python 3.12 and `litellm==1.53.3`:

```python
import litellm

OPENAI_MODEL = "gpt-4o-2024-11-20"
GEMINI_MODEL = "gemini/gemini-1.5-flash"


response = litellm.completion(
    model=GEMINI_MODEL,
    messages=[
        {"role": "user", "content": "Select a tool to print 'hi'."},
    ],
    tool_choice="required",
    tools=[
        {
            "type": "function",
            "function": {
                "name": "print_hi",
                "description": "Print 'hi'.",
                "parameters": {"type": "object", "properties": {}, "required": []},
            },
        }
    ],
)
assert len(response.choices) == 1, "Expected one response choice."
print(response.choices[0].message.tool_calls)
```

It works fine with OpenAI `gpt-4o-2024-11-20`, but does not work with Google `gemini/gemini-1.5-flash`:

```none
litellm.exceptions.BadRequestError: litellm.BadRequestError: VertexAIException BadRequestError - {
  "error": {
    "code": 400,
    "message": "* GenerateContentRequest.tools[0].function_declarations[0].parameters.properties: should be non-empty for OBJECT type\n",
    "status": "INVALID_ARGUMENT"
  }
}
```

So Gemini does not yet support empty tool `parameters`.

Unfortunately, PaperQA's `gen_answer` and `reset` tools have empty parameters. Thus, `gemini/gemini-1.5-flash` can't be used for tool selections (our agent LLM).

vietvudanh:
Can we just drop the key `properties` when it's an empty object?

jamesbraza:
@vietvudanh if you can get some variant of the above snippet to run, perhaps we can implement it.

I just tried removing `properties` like you suggest, so `"parameters": {"type": "object", "required": []},`, and still got the same `INVALID_ARGUMENT` error.

vietvudanh:
> [@vietvudanh](https://github.com/vietvudanh) if you can get some variant of the above snippet to run, perhaps we can implement it.
> 
> I just tried removing `properties` like you suggest, so `"parameters": {"type": "object", "required": []},`, and still got the same `INVALID_ARGUMENT` error.

I mean to remove `"parameters"`. As in when call `litellm.completion`, this works:
```python

# this works
litellm.completion(
  ...
  tools=[
          {
              "type": "function",
[...]

PR #871, opened 2025-02-17
Add support and instructions for gemini

JoaquinPolonuer:
Closes https://github.com/Future-House/paper-qa/issues/796


README.md
```diff
@@ -116,6 +116,7 @@ So we arbitrarily call version 5 and onward PaperQA2,
 and versions before it as PaperQA1 to denote the significant change in performance.
 We recognize that we are challenged at naming and counting at FutureHouse,
 so we reserve the right at any time to arbitrarily change the name to PaperCrow.
+https://github.com/Future-House/paper-qa/pull/870
```
jamesbraza, line 119: Did you mean to link this PR here?

README.md
```diff
@@ -432,6 +433,25 @@ answer_response = ask(
 )
 ```
 
+If you want to use Gemini, you will have to set your `VERTEX_PROJECT` and `VERTEX_LOCATION` and be logged in using the Google Cloud CLI. You can ensure this by executing:
+
+```bash
+gcloud auth application-default login
+```
+
+and install the optional dependency `paper-qa[gemini]`
```
jamesbraza, line 442: Let's put this first, since it's required before you can run `gcloud auth`

JoaquinPolonuer: Are you sure? `paper-qa[gemini]` will only install the python `google-auth` library. The gcloud cli is installed [here](https://cloud.google.com/sdk/docs/install)

README.md
```diff
@@ -432,6 +432,21 @@ answer_response = ask(
...
+
+```python
+from paperqa import Settings, ask
+
+answer_response = ask(
+    "What manufacturing challenges are unique to bispecific antibodies?",
+    settings=Settings(
+        llm="gemini-1.5-pro",
+        summary_llm="gemini-1.5-pro",
+        agent=AgentSettings(agent_llm="gemini-1.5-pro"),
```
JoaquinPolonuer, line 445: Why do we not have this line for anthropic and openai in the readme?

jamesbraza: Our docs need much love. Model configuration is our greatest source of issues

Feel free to give more love

paperqa/agents/env.py
```diff
@@ -55,6 +56,7 @@ def settings_to_tools(  # noqa: PLR0912
     llm_model = llm_model or settings.get_llm()
     summary_llm_model = summary_llm_model or settings.get_summary_llm()
     embedding_model = embedding_model or settings.get_embedding_model()
+    agent_llm_model = agent_llm_model or settings.get_agent_llm()
```
JoaquinPolonuer, line 59: Will remove this if we use the function in Settings

paperqa/agents/main.py
```diff
@@ -333,19 +338,25 @@ def __init__(
...
             await self.on_agent_action_callback(*args)
 
-    async def after_env_reset(self, traj_id: str, *_) -> None:  # noqa: ARG002
+    async def after_env_reset(
+        self, traj_id: str, obs: list[Message], tools: list[Tool]  # noqa: ARG002
+    ) -> None:
         if self.on_env_reset_callback is not None:
             await self.on_env_reset_callback(self.env.state)
+        if self.settings:
+            self.settings.adapt_tools(tools)
```
JoaquinPolonuer, line 359: Is it ok to have settings in this class?

Also, there is something weird in this implementation: This will run `adapt_tools` every time an `after_env_reset` callback runs. I'm not sure if I like it.

The problem with passing the `adapt_tools` function as a callback (which I think would make more sense) is that the `after_env_reset` callback is defined to only receive the state.

Thoughts on this?

paperqa/settings.py
```diff
@@ -932,6 +932,12 @@ async def make_ldp_agent(
             )
         raise NotImplementedError(f"Didn't yet handle agent type {agent_type}.")
 
+    def adapt_tools(self, tools):
```
jamesbraza, line 935: I like that the name makes it clear the input `tools` will be mutated, that's good

But the name can be a bit more verbose to reflect:
- Something only happens for a subset of LLM types
- It relates to the agent

Perhaps name `adjust_tools_for_agent_llm`

paperqa/settings.py
```diff
@@ -932,6 +932,12 @@ async def make_ldp_agent(
...
+        if "gemini" in self.agent.agent_llm.lower():
```
jamesbraza, line 936: Can you link the LiteLLM issue https://github.com/BerriAI/litellm/issues/7634 , because in the future we can remove this logic if LiteLLM supports

paperqa/agents/main.py
jamesbraza: > The problem with passing the `adapt_tools` function as a callback (which I think would make more sense) is that the `after_env_reset` callback is defined to only receive the state.

I am not following, `after_env_reset` receives `obs`, `tools`, `traj_id`, etc. We don't pass state at all?

> Is it ok to have settings in this class?

It's a breaking change in the signature, now seeing this it may be cleaner just to make a second focused callback:

```python
class LDPAdjustToolsForAgentCallback(Callback):
    def __init__(self, settings: Settings):
        self._settings = settings

    async def after_env_reset(
        self, traj_id: str, obs: list[Message], tools: list[Tool]
    ) -> None:
        self._settings.adjust_tools_for_agent_llm(tools)
```

paperqa/agents/main.py
```diff
@@ -339,17 +344,15 @@ def __init__(
         self.on_agent_action_callback = on_agent_action_callback
         self.on_env_step_callback = on_env_step_callback
 
-    async def after_agent_get_asv(self, traj_id: str, *args) -> None:  # noqa: ARG002
-        if self.on_agent_action_callback is not None:
-            await self.on_agent_action_callback(*args)
```
JoaquinPolonuer, line 344: Sorry didn't mean to remove this. Will fix.

paperqa/settings.py
```diff
@@ -932,6 +932,14 @@ async def make_ldp_agent(
             )
         raise NotImplementedError(f"Didn't yet handle agent type {agent_type}.")
 
+    def adjust_tools_for_agent_llm(self, tools):
```
jamesbraza, line 935: Do you mind type hinting this one

paperqa/settings.py
JoaquinPolonuer: I stole your comment from aviary, is that ok?

jamesbraza: Yeah sure it's fine

paperqa/settings.py
```diff
@@ -932,6 +932,14 @@ async def make_ldp_agent(
             )
         raise NotImplementedError(f"Didn't yet handle agent type {agent_type}.")
 
+    def adjust_tools_for_agent_llm(self, tools: list[Tool]):
```
jamesbraza, line 935:
```suggestion
    def adjust_tools_for_agent_llm(self, tools: list[Tool]) -> None:
```

jamesbraza (commented):
Can you make a unit test of this

jamesbraza (approved):
Yaaay nice work 🥳

JoaquinPolonuer:
> Yaaay nice work 🥳

Thank you! 😊

PR #881, opened 2025-02-23
Supporting `.md` in indexing

jamesbraza:
This PR adds support for `.md` to index folders
- Adds https://en.wikipedia.org/wiki/Gravity_hill to our stub data
- Adds test cases for it


mskarlin (approved):
Igtm

PR #892, opened 2025-02-26
Support to parse nested quotes in JSONs during the summary step 

mskarlin:
Claude 3.7 likes to use these and we fail otherwise.


jamesbraza (approved):

mskarlin:
I'm seeing an unrelated test failure that might be stemming from openAI's prompt caching -- if upping the retries doesn't work I'm going to swap to an anthropic model in the test

PR #893, opened 2025-02-27
Added tutorial showing how to configure paperqa `Settings`

maykcaldas:
The main motivation for that is the number of open issues with user asking why paperqa still requires an `OPENAI_API_KEY` even though they set `Settings.llm` to some other provider. This is a quick tutorial showing why paperqa keeps asking OpenAI key in such cases.

TODO:
- [x] Check if gitbook is able to render a jupyter notebook

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,13 @@ repos:
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
+  - repo: https://github.com/mwouts/jupytext
+    rev: v1.16.7
+    hooks:
+      - id: jupytext
+        args: [--sync, --pipe, black]
+        additional_dependencies: [black]
+        files: ^docs/.*\.ipynb$
   - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 25.1.0
     hooks:
@@ -35,13 +42,6 @@ repos:
     rev: v3.4.2
     hooks:
       - id: prettier
-  - repo: https://github.com/mwouts/jupytext
-    rev: v1.16.7
-    hooks:
-      - id: jupytext
-        args: [--sync, --pipe, black]
-        additional_dependencies: [black]
-        files: ^docs/.*\.ipynb$
   - repo: https://github.com/pappasam/toml-sort
     rev: v0.24.2
     hooks:
@@ -78,7 +78,7 @@ repos:
     rev: 1.19.1
     hooks:
       - id: blacken-docs
-        exclude: \.md$ # markdown files are being blackened by jupytext
+        exclude: \.md$ # The generated markdown files are being blackened by jupytext
   - repo: https://github.com/srstevenson/nb-clean
     rev: 4.0.1
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,6 +42,7 @@ repos:
     rev: v3.4.2
     hooks:
       - id: prettier
+        exclude: ^docs/.*\.md$
   - repo: https://github.com/pappasam/toml-sort
     rev: v0.24.2
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,13 @@ repos:
     rev: v3.4.2
     hooks:
       - id: prettier
+  - repo: https://github.com/mwouts/jupytext
+    rev: v1.16.7
+    hooks:
+      - id: jupytext
+        args: [--sync, --pipe, black]
+        additional_dependencies: [black]
+        files: ^docs/.*\.ipynb$
   - repo: https://github.com/pappasam/toml-sort
     rev: v0.24.2
     hooks:
@@ -71,6 +78,12 @@ repos:
     rev: 1.19.1
     hooks:
       - id: blacken-docs
+        exclude: \.md$ # markdown files are being blackened by jupytext
+  - repo: https://github.com/srstevenson/nb-clean
+    rev: 4.0.1
+    hooks:
+      - id: nb-clean
+        args: [--preserve-cell-outputs, --remove-empty-cells]
   - repo: https://github.com/jsh9/markdown-toc-creator
     rev: 0.0.10
     hooks:
--- a/docs/tutorials/settings_tutorial.ipynb
+++ b/docs/tutorials/settings_tutorial.ipynb
@@ -22,10 +22,7 @@
     "!echo \"OPENAI_API_KEY=<your-openai-api-key>\" > .env\n",
     "!echo \"ANTHROPIC_API_KEY=<your-anthropic-api-key>\" >> .env\n",
     "\n",
-    "!uv pip install -q nest-asyncio\n",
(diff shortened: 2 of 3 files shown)


docs/tutorials/settings_tutorial.md
```diff
@@ -0,0 +1,301 @@
+# Setup
```
jamesbraza, line 1: So there are tools to convert Jupyter notebooks into Markdown. Do you mind if we consolidate into Jupyter?

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,437 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(\"You have set the following environment variables:\")\n",
+    "print(f\"OPENAI_API_KEY: {os.environ['OPENAI_API_KEY']}\")\n",
+    "print(f\"ANTHROPIC_API_KEY: {os.environ['ANTHROPIC_API_KEY']}\")"
```
jamesbraza, line 52: I feel like someone out there is going to run this, forget to clear outputs, share the notebook, and thus leak these secrets.

What do you think of removing this cell?

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,437 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `Settings` class is used to configure the PaperQA settings. \n",
+    "Official documentation can be found [here](https://github.com/Future-House/paper-qa?tab=readme-ov-file#settings-cheatsheet) and the open source code can be found [here](https://github.com/Future-House/paper-qa/blob/main/paperqa/settings.py).\n",
+    "\n",
+    "Here is a basic example of how to use the `Settings` class. We will be unnecessarily verbose for the sake of clarity. Please notice that most of the settings are optional and the defaults are good for most cases. Refer to the [descriptions of each setting]((https://github.com/Future-House/paper-qa/blob/main/paperqa/settings.py)) for more information.\n",
```
jamesbraza, line 91: Extra closing parenthesis

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,437 @@
...
+    "        overlap=250,\n",
+    "        citation_prompt=citation_prompt,\n",
+    "        structured_citation_prompt=structured_citation_prompt,\n",
+    "    ),\n",
+    "    prompts=PromptSettings(\n",
+    "        summary=summary_prompt,\n",
+    "        qa=qa_prompt,\n",
+    "        select=select_paper_prompt,\n",
+    "        pre=None,\n",
+    "        post=None,\n",
```
jamesbraza, line 188: This is already a lot of parameters to read, can we not restate defaulted args?

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,437 @@
...
+    "`Paperqa` returns a `PQASession` object, which contains not only the answer but also all the information gatheres to answer the questions."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "response.__dict__"
```
jamesbraza, line 346: It feels unexpected to use `__dict__` (we don't even use this in our own code), do we need this cell, or can we use `print` with a message?

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,437 @@
...
+    "    print(f\"Content: {ctx.context}\")\n",
+    "    print(f\"Score: {ctx.score}\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
jamesbraza, line 413: We have a blank cell here, can we add `nb-clean` to `pre-commit` to auto-remove these:

```yaml
  - repo: https://github.com/srstevenson/nb-clean
    rev: 4.0.1
    hooks:
      - id: nb-clean
        args: [--preserve-cell-outputs, --remove-empty-cells]
```

docs/tutorials/settings_tutorial.md
maykcaldas: I even prefer. I (claude) created the markdown because I don`t think the gitbook can render jupyter notebooks

.pre-commit-config.yaml
```diff
@@ -71,6 +78,12 @@ repos:
     rev: 1.19.1
     hooks:
       - id: blacken-docs
+        exclude: \.md$ # markdown files are being blackened by jupytext
```
maykcaldas, line 81: This isn't actually true. Only the `.md` files created from an `.ipynb` will be blackened. Need to fix this

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "`Paperqa` returns a `PQASession` object, which contains not only the answer but also all the information gatheres to answer the questions. We recommend printing the `PQASession` object (`print(response)`) to understand the information it contains."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(response.session)"
```
jamesbraza, line 348: Why do we have this standalone `print` before we print `print("Let's examine the PQASession object returned by paperqa:\n")`?

Wouldn't it make sense to have it after that print?

docs/tutorials/settings_tutorial.ipynb
```diff
@@ -0,0 +1,431 @@
...
+    "settings = Settings(\n",
+    "    llm=llm_openai,\n",
+    "    llm_config={\n",
+    "        \"model_list\": [\n",
+    "            {\n",
+    "                \"model_name\": llm_openai,\n",
+    "                \"litellm_params\": {\n",
+    "                    \"model\": llm_openai,\n",
+    "                    \"temperature\": 0.1,\n",
+    "                    \"max_tokens\": 512,\n",
```
jamesbraza, line 146: Why are we putting max 512 tokens? This seems like a risky value

jamesbraza (approved):
Nice work!

docs/tutorials/settings_tutorial.ipynb
maykcaldas: It does. The cell before this one is a markdown cell with:
```
# The output

`Paperqa` returns a `PQASession` object, which contains not only the answer but also all the information gatheres to answer the questions. We recommend printing the `PQASession` object (`print(response.session)`) to understand the information it contains. Let's check the `PQASession` object:
```

docs/tutorials/settings_tutorial.ipynb
maykcaldas: I just wanted any number. Because `sumary_llm` and `agent_llm` doesn't set a max_tokens, context and the answer are created correctly still

PR #895, opened 2025-02-28
Added embeddings to claude and gemine examples in the readme

maykcaldas:
- Also changed the gemini model to use googleAI studio's model

Motivation: OpenAI is the default for embeddings. Hence, an `OPENAI_API_KEY` would still be required for those examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,9 @@ By default, PaperQA2 uses OpenAI's `gpt-4o-2024-11-20` model for the
 `summary_llm`, `llm`, and `agent_llm`.
 Please see the [Settings Cheatsheet](#settings-cheatsheet)
 for more information on these settings.
+PaperQA2 also defaults to using OpenAI's `text-embedding-3-small` model for the `embedding` setting.
+If you don't have an OpenAI API key, you can use a different embedding model.
+More information about embedding models can be found [in the "Embedding Model" section](#embedding-model).
 
 We use the [`lmi`](https://github.com/Future-House/ldp/tree/main/packages/lmi) package for our LLM interface,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -435,6 +435,7 @@ answer_response = ask(
         llm="claude-3-5-sonnet-20240620",
         summary_llm="claude-3-5-sonnet-20240620",
         agent=AgentSettings(agent_llm="claude-3-5-sonnet-20240620"),
+        embedding="hybrid-st-multi-qa-MiniLM-L6-cos-v1",
```
jamesbraza, line 438: Why are we doing this? It seems like a really random model

Can you at least add a comment or expand `README` to mention what this is for

maykcaldas: Just because if we don't set `embedding`, `pqa` will still use `openai` embeddings. 

Because of the "To use Claude, make sure you set the `ANTHROPIC_API_KEY`", I wanted to make it independent from OpenAI.

Added more context and explanation

README.md
```diff
@@ -435,6 +440,7 @@ answer_response = ask(
         llm="claude-3-5-sonnet-20240620",
         summary_llm="claude-3-5-sonnet-20240620",
         agent=AgentSettings(agent_llm="claude-3-5-sonnet-20240620"),
+        embedding="st-multi-qa-MiniLM-L6-cos-v1",
```
jamesbraza, line 443:
```suggestion
        # SEE: https://huggingface.co/sentence-transformers/multi-qa-MiniLM-L6-cos-v1
        embedding="st-multi-qa-MiniLM-L6-cos-v1",
```

Just linking the model so people can see it's open source

jamesbraza (approved):

PR #896, opened 2025-02-28
Add reasoning from LLMResult to PQASession

maykcaldas:
If a reasoning is not available, an empty string is saved -- no reasoning


paperqa/docs.py
```diff
@@ -794,6 +794,7 @@ async def aquery(  # noqa: PLR0912
                     name="answer",
                 )
             answer_text = cast(str, answer_result.text)
+            answer_reasoning = answer_result.reasoning_content or ""
```
jamesbraza, line 797: So here, empty reasoning is different from no reasoning. Can we just do `answer_reasoning = answer_result.reasoning_content` and have `PQASession.answer_reasoning: str | None`

maykcaldas: Will `None` going to be jsonified well when saving the session in `paperqa-server`?

jamesbraza: Yeah `None` is valid JSON. If you're worried about it, could add a `model_dump(mode="json")` somewhere in tests here to be sure

paperqa/types.py
```diff
@@ -126,6 +126,7 @@ class PQASession(BaseModel):
     id: UUID = Field(default_factory=uuid4)
     question: str
     answer: str = ""
+    answer_reasoning: str | None = None
```
jamesbraza, line 129: Can you add a `description` here just stating that if the model supports reasoning, we include it here

jamesbraza (approved):

PR #898, opened 2025-03-02
Updated tooling versions and configs

jamesbraza:
Modernized for March 2025


paperqa/agents/env.py
```diff
@@ -72,7 +72,7 @@ def settings_to_tools(  # noqa: PLR0912
             )
             for pname in ("min_year", "max_year"):
                 tool.info.get_properties()[pname]["description"] = cast(
-                    str, tool.info.get_properties()[pname]["description"]
+                    "str", tool.info.get_properties()[pname]["description"]
```
ludomitch, line 75: I'm not sure I like having strings instead of types. Is it to reduce the imports? What's the benefit?

jamesbraza: It's [`TC006`](https://docs.astral.sh/ruff/rules/runtime-cast-value/), check it's justification there and lmk what you think. From my perspective, I think it's an alright rule

mskarlin (approved):
thanks for these updates

PR #911, opened 2025-03-12
Support for answer iterations in the gen_answer flow

mskarlin:
If we are answering after a prior answer already exists in the session, we currently don't expose that to the `gen_answer` tool  / `aquery`. To support continuations, where a user follows up to a query, exposing the prior answer if very helpful (because the agent can't pass it directly), so this implements that feature.

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -782,11 +782,18 @@ class Docs(BaseModel):
                     Message(role="system", content=prompt_config.system),
                     Message(
                         role="user",
-                        content=prompt_config.qa.format(
+                        content=prompt_config.configured_qa_prompt.format(
                             context=context_str,
                             answer_length=answer_config.answer_length,
                             question=session.question,
                             example_citation=prompt_config.EXAMPLE_CITATION,
+                            answer_iteration_context=(
+                                prompt_config.iteration_prompt.format(
+                                    prior_answer=session.answer
+                                )
+                                if session.answer
+                                else ""
+                            ),
                         ),
                     ),
                 ]
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -778,9 +778,9 @@ class Docs(BaseModel):
             answer_reasoning = None
         else:
             with set_llm_session_ids(session.id):
-                iteration_context = ""
+                prior_answer_prompt = ""
                 if prompt_config.iteration_prompt and session.answer:
-                    iteration_context = prompt_config.iteration_prompt.format(
+                    prior_answer_prompt = prompt_config.iteration_prompt.format(
                         prior_answer=session.answer
                     )
                 messages = [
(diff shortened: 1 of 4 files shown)


paperqa/settings.py
```diff
@@ -265,6 +267,9 @@ class PromptSettings(BaseModel):
 
     summary: str = summary_prompt
     qa: str = qa_prompt
+    qa_without_iteration: str = qa_without_iteration_prompt
+    iteration_prompt: str = answer_iteration_prompt
```
jamesbraza, line 271: Can you add a `description` stating what this is?

Also, how do you feel about `use_qa_iterations` somehow being an inference on `iteration_prompt`'s value being non-default?

paperqa/settings.py
```diff
@@ -362,6 +379,12 @@ def check_context_inner(cls, v: str) -> str:
             raise ValueError("Context inner prompt must have name and text")
         return v
 
+    @property
+    def configured_qa_prompt(self) -> str:
```
jamesbraza, line 383: Why not just `qa_prompt`?

mskarlin: sure

paperqa/settings.py
mskarlin: good point -- actually I'm just gonna cut out the `qa_without_iteration` entirely, I was trying to hedge, but it's not needed.

nadolskit (approved):

paperqa/prompts.py
```diff
@@ -21,6 +21,15 @@
 # 1. Lead to complete tool being called with has_successful_answer=False
 # 2. Can be used for unit testing
 CANNOT_ANSWER_PHRASE = "I cannot answer"
+
+answer_iteration_prompt = (
+    "You are iterating on a prior answer, with a potentially different context:\n\n"
+    "{prior_answer}\n\n"
+    "Create a new answer only using keys and data from the included context, do not "
```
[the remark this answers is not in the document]
mskarlin: yes! good catch

paperqa/prompts.py
```diff
@@ -21,6 +21,15 @@
 # 1. Lead to complete tool being called with has_successful_answer=False
 # 2. Can be used for unit testing
 CANNOT_ANSWER_PHRASE = "I cannot answer"
+
+answer_iteration_prompt = (
```
jamesbraza, line 25: Any thoughts on naming this `answer_iteration_prompt_template`

paperqa/docs.py
```diff
@@ -778,6 +778,11 @@ async def aquery(  # noqa: PLR0912
             answer_reasoning = None
         else:
             with set_llm_session_ids(session.id):
+                iteration_context = ""
```
jamesbraza, line 781: [naming nit] this may be confused with `Context` (the type here)

Can we name something different?

jamesbraza (approved):

paperqa/settings.py
```diff
@@ -265,6 +266,13 @@ class PromptSettings(BaseModel):
 
     summary: str = summary_prompt
     qa: str = qa_prompt
+    iteration_prompt: str | None = Field(
```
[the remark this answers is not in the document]
mskarlin: ok renaming

sremo (approved):

PR #912, opened 2025-03-13
Only show the top contexts to the agent for the specific question

mskarlin:
If multiple gather evidence iterations are performed in a question, the agent sees the top contexts for all submitted questions currently. We should only show the agent the top contexts for the particular question being asked of gather_evidence at that time. This PR fixes that and adds a test.

### DIFF
--- a/paperqa/agents/tools.py
+++ b/paperqa/agents/tools.py
@@ -251,8 +251,15 @@ class GatherEvidence(NamedTool):
 
         status = state.status
         logger.info(status)
+        # only show top n contexts for this particular question to the agent
         sorted_contexts = sorted(
-            state.session.contexts, key=lambda x: x.score, reverse=True
+            [
+                c
+                for c in state.session.contexts
+                if (c.question is None or c.question == question)
+            ],
+            key=lambda x: x.score,
+            reverse=True,
         )
 
         top_contexts = "\n".join(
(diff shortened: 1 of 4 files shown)


paperqa/agents/tools.py
```diff
@@ -251,8 +251,15 @@ async def gather_evidence(self, question: str, state: EnvironmentState) -> str:
 
         status = state.status
         logger.info(status)
+        # only show top n contexts for this particular question to the agent
         sorted_contexts = sorted(
-            state.session.contexts, key=lambda x: x.score, reverse=True
+            [
+                c
+                for c in state.session.contexts
+                if (c.question is None or c.question == question)
```
maykcaldas, line 259: Should a `Context` with `c.question == None` exist?

maykcaldas (approved):
There's [this test](https://github.com/Future-House/paper-qa/blob/0e519b10fd6537dc22bd5daf3a1477c3680d8bc7/tests/test_agents.py#L917) which uses `Context` and `question` is not set. 

Specifically for the test, it doesn't matter. But do we want to add the question to the context just to be consistent?

Besides that, it lgtm!

paperqa/agents/tools.py
mskarlin: I left it optional for users -- this was just for type safety, but I don't think it can happen practically.

PR #916, opened 2025-03-20
Pulled in latest `tantivy` for `Document.from_dict` typing fix

jamesbraza:
Pulling in https://github.com/quickwit-oss/tantivy-py/pull/346


maykcaldas (approved):
Every `# type: ignore` that is deleted makes me happy

PR #872, opened 2025-02-19
Migrated LitQA and LFRQA to aviary

maykcaldas:
Code from LitQA and LFRQA were moved to aviary: https://github.com/Future-House/aviary/pull/190

This PR cleans up paperqa

PR #923, opened 2025-03-29
Removed LFRQA prompts

jamesbraza:
https://github.com/Future-House/aviary/pull/191 correctly propagated these prompts to `aviary`, but https://github.com/Future-House/paper-qa/pull/872 missed their removal here.


maykcaldas (approved):
Nice catch!

PR #929, opened 2025-04-01
Gave instructions to correct citation on answer

maykcaldas:
The LLM sometimes broke the pattern by using citation styles other than the citation key, which caused problems on the platform downstream. This PR gives more instructions on how the citation should be included in the answer.

A few examples of wrong citations:

1 - Task: What is the molecule known to have the greatest solubility in water?
Answer: (...) Similarly, studies on aqueous solubility in drug discovery (e.g., Ishikawa and Hashimoto, 2011) discuss extremely low solubility values for particular organic compounds, but these systems are not directly comparable to the gas-phase solubility measurements in Hildebrand’s work. (...)
Issue: *Ishikawa and Hashimoto, 2011* is not a citation key

2 - Task: What is the molecule known to have the greatest solubility in water?
Answer: (...) Therefore, based solely on the provided context, ammonia (NH₃) is the molecule known to have the greatest solubility in water (hildebrand1916solubility pages 14–17).
Issue: * Hildebrand’s 1916 paper (pages 14–17) * and *hildebrand1916solubility* should be [this paper](https://pubs.acs.org/doi/abs/10.1021/ja02265a002). It should have used "*hildebrand1916solubility. pages 17-19*". The difference is the period.

We need to use whatever the citation key is and avoid the LLM from changing it


tests/test_agents.py
```diff
@@ -973,6 +973,31 @@ async def test_clinical_tool_usage(agent_test_settings) -> None:
     ), "No clinical trials were put into contexts"
 
 
+@pytest.mark.asyncio
+async def test_citation_formatting(agent_test_settings):
+    docs = Docs()
+    response = await run_agent(
+        docs,
+        query=("What is XAI and how does it work?"),
```
jamesbraza, line 981: [nit] extra parens

paperqa/prompts.py
```diff
@@ -38,14 +38,19 @@
...
-    "relevant, use them in the answer. This answer will go directly onto "
-    "Wikipedia, so do not add any extraneous information.\n\n"
+    "it via citation keys at the end of sentences, like {example_citation}. "
+    "You can only use this formatting for citations. "
+    "For instance, if the citation key is 'example2012example pages 3-4', "
+    "you cannot use 'Example2012Example', 'Example2012 (pages 3-4)', "
+    "'Example et. al. (2012)', 'example2012example pages 3-4, pages 5-6' "
+    "or any adapted version of these citation keys "
+    "as a citation key in your answer. Use the citation keys from the contexts below. "
+    "Only cite from the context below and only use the valid keys. "
```
jamesbraza, line 48: These two lines feel redundant to me

paperqa/prompts.py
```diff
@@ -38,14 +38,19 @@
...
-    "it via citation keys at the end of sentences, "
-    "like {example_citation}. Only cite from the context "
-    "below and only use the valid keys. Write in the style of a "
-    "Wikipedia article, with concise sentences and coherent paragraphs. "
-    "The context comes from a variety of sources and is only a summary, "
-    "so there may inaccuracies or ambiguities. If quotes are present and "
-    "relevant, use them in the answer. This answer will go directly onto "
-    "Wikipedia, so do not add any extraneous information.\n\n"
+    "it via citation keys at the end of sentences, like {example_citation}. "
+    "You can only use this formatting for citations. "
```
jamesbraza, line 42: Can you say `"You can only use this formatting for citations, do not apply it to X or Y."` (and add what X or Y are)

jamesbraza (commented):
In the PR description, can you provide the output that made you discover this issue, for posterity

jamesbraza (approved):
🙏

tests/test_agents.py
```diff
@@ -973,6 +973,30 @@ async def test_clinical_tool_usage(agent_test_settings) -> None:
     ), "No clinical trials were put into contexts"
 
 
+@pytest.mark.asyncio
+async def test_citation_formatting(agent_test_settings):
```
jamesbraza, line 977: What do you think of collapsing this test into a subtest of `test_agent_types`, and it would actually:

- Decrease CI runtime (since we're already running the agent in those tests)
- Increase the coverage of the assertions, since more agents types are used there

tests/test_agents.py
```diff
@@ -973,6 +973,30 @@ async def test_clinical_tool_usage(agent_test_settings) -> None:
...
+    response = await run_agent(
+        docs,
+        query="What is XAI and how does it work?",
+        settings=agent_test_settings,
+    )
+
+    name_year_regex = r"\b([A-Z][a-zA-Z]*\d{4})(?:\s+(pages\s+)?(\d+-\d+))?\b"
+    assert not re.search(
+        name_year_regex, response.session.answer
+    ), "Answer contains citation with name and year instead of citation key"
```
jamesbraza, line 988: This assertion fails on current `main` branch, nice work here!! 🚀

PR #947, opened 2025-05-07
Add ipykernel as dev dependency

tobiasdiez:
It's handy to run the example notebooks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ requires-python = ">=3.11"
 [project.optional-dependencies]
 dev = [
     "fhlmi>=0.27",  # For LLMResult reflecting actual model
+    "ipykernel>=6.29",  # For running Jupter notebooks, and pin to keep recent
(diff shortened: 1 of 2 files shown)


jamesbraza:
Hi @tobiasdiez thanks for this PR, down to merge it but just for understanding, can you elaborate on why `ipykernel` is useful?

jamesbraza:
Okay I made a fresh `.venv` and tried to run a Jupyter Notebook in Cursor, and I got:

> Running cells with '.venv (Python 3.13.2)' requires the ipykernel and pip package.

I see now, yes this is useful indeed.

tobiasdiez:
> Okay I made a fresh `.venv` and tried to run a Jupyter Notebook in Cursor, and I got:
> 
> > Running cells with '.venv (Python 3.13.2)' requires the ipykernel and pip package.
> 
> I see now, yes this is useful indeed.

That was also exactly how I discovered it.

PR #948, opened 2025-05-07
Add error handling for index file loading in SearchIndex

tobiasdiez:
I got the following error while playing with paper-qa:

```
File D:\Programming\paper-qa\paperqa\agents\search.py:262, in SearchIndex.index_files(self)
    259         async with await anyio.open_file(file_index_path, "rb") as f:
    260             content = await f.read()
    261             self._index_files = pickle.loads(  # noqa: S301
--> 262                 zlib.decompress(content)
    263             )
    264 return self._index_files

error: Error -5 while decompressing data: incomplete or truncated stream
```
Since it was a bit hard to locate the offending file, I added logging so that one can see where the error is.

(Not sure why the error occurred in the first place, it disappeared after deleting the index files.)

### DIFF
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -266,6 +266,7 @@ class SearchIndex:
                         logger.exception(
                             f"Failed to load index file {file_index_path}."
                         )
+                        raise
         return self._index_files
 
     @staticmethod
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -258,9 +258,12 @@ class SearchIndex:
             if await file_index_path.exists():
                 async with await anyio.open_file(file_index_path, "rb") as f:
                     content = await f.read()
-                    self._index_files = pickle.loads(  # noqa: S301
-                        zlib.decompress(content)
-                    )
+                    try:
+                        self._index_files = pickle.loads(  # noqa: S301
+                            zlib.decompress(content)
+                        )
+                    except Exception:
+                        logger.exception(f"Failed to load index file {file_index_path}")
         return self._index_files
 
     @staticmethod
--- a/paperqa/agents/search.py
+++ b/paperqa/agents/search.py
@@ -263,7 +263,9 @@ class SearchIndex:
                             zlib.decompress(content)
                         )
                     except Exception:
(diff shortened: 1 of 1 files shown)


paperqa/agents/search.py
```diff
@@ -258,9 +258,12 @@ async def index_files(self) -> dict[str, str]:
...
                     content = await f.read()
-                    self._index_files = pickle.loads(  # noqa: S301
-                        zlib.decompress(content)
-                    )
+                    try:
+                        self._index_files = pickle.loads(  # noqa: S301
+                            zlib.decompress(content)
+                        )
+                    except Exception:
+                        logger.exception(f"Failed to load index file {file_index_path}")
```
jamesbraza, line 266:
```suggestion
                        logger.exception(f"Failed to load index file {file_index_path}.")
```

Adding a period can help reveals empty string issues

paperqa/agents/search.py
```diff
@@ -258,9 +258,12 @@ async def index_files(self) -> dict[str, str]:
...
                 async with await anyio.open_file(file_index_path, "rb") as f:
                     content = await f.read()
-                    self._index_files = pickle.loads(  # noqa: S301
-                        zlib.decompress(content)
-                    )
+                    try:
+                        self._index_files = pickle.loads(  # noqa: S301
+                            zlib.decompress(content)
+                        )
+                    except Exception:
```
jamesbraza, line 265: We need to re-`raise`, as suppressing the `Exception` (what your PR does) changes the behavior

tobiasdiez: Isn't it 'nicer' to just ignore the corrupt index file than fail?

jamesbraza: These `incomplete or truncated stream` correspond with empty or missing files, which means degraded performance of the index. The degradation's impact depends on how valuable the missing contents were.

Let's say the missing file contains the relevant information to answer, so you get a false negative: PaperQA would get the answer incorrect, but just because the relevant information was gone from the index.

So it's fine to add this logging of the impacted file, but we don't want to allow degraded performance in this part of the stack.

tobiasdiez: Okay, added the reraising now.

PR #962, opened 2025-06-11
Restoring `typeguard` functionality

jamesbraza:
To get `typeguard` to work you need to not import the target package into `conftest.py`. We were doing that, so to not do that:
- Using `lmi` imports where possible
- Lazily importing the rest

Then, I resolved all type errors that were exposed.


paperqa/docs.py
```diff
@@ -60,7 +60,7 @@ async def empty_callback(result: LLMResult) -> None:
     pass
 
 
-async def print_callback(result: LLMResult) -> None:
+async def print_callback(result: str | LLMResult) -> None:
```
jamesbraza, line 63: Arguably we should just remove this `print_callback` altogether, not sure why we have it

whitead (approved):

paperqa/docs.py
whitead: think it's a demo in README

jamesbraza: It seems unused everywhere from a CTRL-F, so I went ahead and removed in https://github.com/Future-House/paper-qa/pull/964

PR #965, opened 2025-06-11
Fixed `chunk_code_text`'s `UnboundLocalError` crash on empty code file

jamesbraza:
If a code file is empty (e.g. a `py.typed` file), the `Text`-creation loop inside `chunk_code_text` will not be entered, so we hit an `UnboundLocalError`:

```none
UnboundLocalError: cannot access local variable 'i' where it is not associated with a value
```

This PR:

- Fixes that crash, with test coverage
- Improves variable names and documents the purpose of conditional logic

### DIFF
--- a/paperqa/readers.py
+++ b/paperqa/readers.py
@@ -222,7 +222,7 @@ def chunk_code_text(
     """Parse a document into chunks, based on line numbers (for code)."""
     text_buffer = ""
     texts: list[Text] = []
-    last_line_i = 0
+    line_i = last_line_i = 0
 
     if not isinstance(parsed_text.content, str | list):
         raise NotImplementedError(
--- a/paperqa/readers.py
+++ b/paperqa/readers.py
@@ -245,7 +245,10 @@ def chunk_code_text(
             )
             text_buffer = text_buffer[chunk_chars - overlap :]
             last_line_i = line_i
-    if len(text_buffer) > overlap or not texts:
+    if (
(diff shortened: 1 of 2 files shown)


paperqa/readers.py
```diff
@@ -220,36 +220,39 @@ def chunk_code_text(
...
+    if (
+        len(text_buffer) > overlap  # Save meaningful amount of content as a final text
+        or not texts  # Contents were smaller than one chunk, save it anyways
+    ):
         texts.append(
             Text(
-                text=split[:chunk_chars],
-                name=f"{doc.docname} lines {last_line}-{i}",
+                text=text_buffer[:chunk_chars],
+                name=f"{doc.docname} lines {last_line_i}-{line_i}",
```
maykcaldas, line 255: If the text is empty, won't we have a `Text` with `Text.name` like "Document lines 0-0"? Should we treat this?

nadolskit (approved):

paperqa/readers.py
jamesbraza: You're right about this. Though I didn't mind the lines 0-0, it's explicit/traceable. Also, it avoids making a special case

Do you think we should just get rid of `lines 0-0` in the case of empty text?

maykcaldas: I was thinking of not even making a `Text` in this case. But I think it's fine. You're right that they're more explicit. I think it's good to have it like that

PR #972, opened 2025-06-13
add client response errors to our catching stack

mskarlin:
s2 has started often returning 403s which are not currently caught otherwise.


paperqa/clients/client_models.py
```diff
@@ -117,6 +117,12 @@ async def query(self, query: dict) -> DocDetails | None:
                 f" {client_query.doi if isinstance(client_query, DOIQuery) else client_query.title} in"
                 f" {self.__class__.__name__}."
             )
+        except aiohttp.ClientResponseError:
```
jamesbraza, line 121: Can you add a comment explaining this suppression

jamesbraza (approved):

PR #977, opened 2025-06-19
Allow for non-mutated user specified doc-ids

mskarlin:
This was a deep bug. On round-trip serialization of a `DocDetails` object, if a user doesn't have a `doi` field to create a deterministic `doc_id`, then when serializing/deserializing (during validation), the `doc_id` was mutating. This prevented users from ever setting their own `doc_id` value. 

This round trip serialization happens during the `gather_evidence` step, when the existing `doc_id` for each `DocDetails` object in `Docs` would then diverge from the `doc_id` for the associated `Context` objects generated. So for each call to `gather_evidence`, if a `DocDetails` did not have a DOI, a random `doc_id` was created for each `Text.doc` object within each `Context`. This PR fixes this behavior.

As an aside, my tests were failing because the `RetractionDataPostProcessor` test file (`tests/stub_data/test_retractions.csv`) was more than 30 days since the creation date. So this triggered a cache break which tried to rebuild the file. This then triggered a network call which broke the VCR cassette cache. I turned this behavior off by default, as I don't think this is needed under normal circumstances.

### DIFF
--- a/paperqa/clients/journal_quality.py
+++ b/paperqa/clients/journal_quality.py
@@ -45,12 +45,14 @@ class JournalQualityPostProcessor(MetadataPostProcessor[JournalQuery]):
         # remember, if both have docnames (i.e. key) they are
         # wiped and re-generated with resultant data
         return doc_details + DocDetails(
+            doc_id=doc_details.doc_id,  # ensure doc_id is preserved
+            dockey=doc_details.dockey,  # ensure dockey is preserved
             source_quality=max(
                 [
                     self.data.get(query.journal.casefold(), DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
                     self.data.get("the " + query.journal.casefold(), DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
                 ]
-            )
+            ),
         )
 
     def query_creator(self, doc_details: DocDetails, **kwargs) -> JournalQuery | None:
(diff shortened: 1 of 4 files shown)


nadolskit (approved):

paperqa/clients/journal_quality.py
```diff
@@ -45,12 +45,14 @@ async def _process(
         # remember, if both have docnames (i.e. key) they are
         # wiped and re-generated with resultant data
         return doc_details + DocDetails(
+            doc_id=doc_details.doc_id,  # ensure doc_id is preserved
+            dockey=doc_details.dockey,  # ensure dockey is preserved
```
jamesbraza, line 49: Perhaps let's improve these comment a bit, imo in their current form they're just mirroring the code

A good comment should mention why -- perhaps explain why was the word "preservation" used

```python
 DocDetails(
    # Propagate doc_id and dockey to preserve them
    # so DocDetails construction doesn't generate new ones
    doc_id=doc_details.doc_id,
    dockey=doc_details.dockey,
```

Or maybe remove the comment entirely

paperqa/types.py
```diff
@@ -431,14 +431,15 @@ def lowercase_doi_and_populate_doc_id(cls, data: dict[str, Any]) -> dict[str, An
                 if doi.startswith(url_prefix_to_remove):
                     doi = doi.replace(url_prefix_to_remove, "")
             data["doi"] = doi.lower()
-            data["doc_id"] = encode_id(doi.lower())
-        else:
+            if "doc_id" not in data or not data["doc_id"]:  # keep user defined doc_ids
```
jamesbraza, line 434: `if not data.get("doc_id")` may be simpler, sometimes the double negatives are harder to comprehend

PR #983, opened 2025-07-01
Allow extras to be kept in user output

mskarlin:
We previously filtered these out when the embeddings, etc. were removed.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -960,6 +960,7 @@ def test_answers_are_striped() -> None:
                     ),
                 ),
                 score=3,
+                extra_field="extra_value",
             )
         ],
     )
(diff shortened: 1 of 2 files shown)


tests/test_agents.py
```diff
@@ -969,6 +970,7 @@ def test_answers_are_striped() -> None:
     assert not response.session.contexts[0].text.text
     assert response.session.contexts[0].text.doc is not None
     assert response.session.contexts[0].text.doc.embedding is None
+    assert response.session.contexts[0].extra_field is not None  # type: ignore[attr-defined]
```
jamesbraza, line 973: Can you add an assertion message here as informal docs, something like `, "Extra fields should be propagated so we can XYZ"`

jamesbraza (approved):

PR #984, opened 2025-07-02
Refactor to expose PDF parsing settings

jamesbraza:
This PR enables users to bring their own PDF parser if desired.

It also documents `ParsedText` and `Text` a bit, since they're related.

### DIFF
--- a/paperqa/agents/__init__.py
+++ b/paperqa/agents/__init__.py
@@ -11,8 +11,7 @@ from aviary.utils import MultipleChoiceQuestion
 from pydantic_settings import CliSettingsSource
 from rich.logging import RichHandler
 
-from paperqa.readers import setup_pymupdf_python_logging
-from paperqa.settings import Settings, get_settings
+from paperqa.settings import ParsingSettings, Settings, get_settings
 from paperqa.utils import pqa_directory, run_or_ensure, setup_default_logs
 from paperqa.version import __version__
 
@@ -92,10 +91,12 @@ def configure_log_verbosity(verbosity: int = 0) -> None:
 def configure_cli_logging(verbosity: int | Settings = 0) -> None:
     """Suppress loquacious loggers according to the settings' verbosity level."""
     setup_default_logs()
-    setup_pymupdf_python_logging()
     set_up_rich_handler()
     if isinstance(verbosity, Settings):
+        verbosity.parsing.configure_pdf_parser()
         verbosity = verbosity.verbosity
+    else:
+        ParsingSettings.model_fields["configure_pdf_parser"].default()
     configure_log_verbosity(verbosity)
     if verbosity > 0:
         print(f"PaperQA version: {__version__}")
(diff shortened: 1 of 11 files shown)


paperqa/agents/__init__.py
```diff
@@ -91,9 +92,13 @@ def configure_log_verbosity(verbosity: int = 0) -> None:
 def configure_cli_logging(verbosity: int | Settings = 0) -> None:
     """Suppress loquacious loggers according to the settings' verbosity level."""
     setup_default_logs()
+    setup_pymupdf_python_logging()
```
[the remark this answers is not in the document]
jamesbraza: Oh nice catch, resolved

tests/conftest.py
```diff
@@ -31,15 +31,17 @@ def _load_env() -> None:
...
 def _setup_default_logs() -> None:
     # Lazily import from paperqa so typeguard doesn't throw:
     # > /path/to/.venv/lib/python3.12/site-packages/typeguard/_pytest_plugin.py:93:
     # > InstrumentationWarning: typeguard cannot check these packages because they
     # > are already imported: paperqa
+    from paperqa.settings import ParsingSettings
     from paperqa.utils import setup_default_logs
 
     setup_default_logs()
+    ParsingSettings.model_fields["configure_pdf_parser"].default()
```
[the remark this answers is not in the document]
jamesbraza: I prefer this method, thank you for the suggestion though

paperqa/types.py
```diff
@@ -96,9 +96,19 @@ def matches_filter_criteria(self, filter_criteria: Mapping[str, Any]) -> bool:
...
+
+    text: str = Field(description="Processed text content of the chunk.")
+    name: str = Field(
+        description=(
+            "Human-readable identifier for the chunk"
+            " (e.g., 'Wiki2023 chunk 1', 'sentence1')."
+        )
+    )
+    doc: Doc | DocDetails = Field(
+        union_mode="left_to_right",
```
whitead, line 109: TIL

paperqa/types.py
```diff
@@ -286,10 +296,22 @@ class ParsedMetadata(BaseModel):
 
 
 class ParsedText(BaseModel):
-    """Parsed text (pre-chunking)."""
+    """All text from a document read, before any processing such as chunking."""
```
whitead: Isn't parsing a category of processing? Would say just before chunking

jamesbraza: Sure, reworded just now

paperqa/settings.py
```diff
@@ -202,6 +207,17 @@ class ParsingSettings(BaseModel):
             " summarization."
         ),
     )
+    parse_pdf: PDFParserFn = Field(
```
mskarlin, line 210: Could we pop in an assert on a round trip serialization of our settings objects, since we do so much of this via servers, I'd like a test that fails if anyone breaks this.

mskarlin (approved):
this lgtm nice work

paperqa/settings.py
jamesbraza: Yeah, just added round-trip serialization assertions to `test_settings_default_instantiation`

PR #988, opened 2025-07-03
Implemented `Text.__eq__`

jamesbraza:
This enables one to think less, and resolves a `TODO`.

### DIFF
--- a/paperqa/types.py
+++ b/paperqa/types.py
@@ -123,7 +123,7 @@ class Text(Embeddable):
         )
 
     def __hash__(self) -> int:
-        return hash(self.text)
+        return hash((self.name, self.text))
 
 
 class Context(BaseModel):
--- a/paperqa/types.py
+++ b/paperqa/types.py
@@ -110,6 +110,18 @@ class Text(Embeddable):
         description="Source document this text chunk originates from.",
     )
 
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        # We ignore the embedding since the embedding can:
+        # - Be lazily acquired, or not used (depending on settings)
+        # - Get ditched when serializing a text for an HTTP request
+        return (
+            self.name == other.name
+            and self.text == other.text
+            and self.doc == other.doc
+        )
+
     def __hash__(self) -> int:
         return hash(self.text)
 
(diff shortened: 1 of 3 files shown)


paperqa/types.py
```diff
@@ -110,6 +110,18 @@ class Text(Embeddable):
         description="Source document this text chunk originates from.",
     )
 
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, type(self)):
```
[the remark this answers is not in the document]
jamesbraza: YAGNI on this

paperqa/types.py
```diff
@@ -110,6 +110,18 @@ class Text(Embeddable):
...
+        # - Be lazily acquired, or not used (depending on settings)
+        # - Get ditched when serializing a text for an HTTP request
+        return (
+            self.name == other.name
+            and self.text == other.text
+            and self.doc == other.doc
+        )
+
     def __hash__(self) -> int:
         return hash(self.text)
```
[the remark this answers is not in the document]
jamesbraza: Thanks for the suggestion, I added `name` just now.

I am going to YAGNI on `doc` inclusion since that would require lots of thought

paperqa/types.py
```diff
@@ -110,8 +110,20 @@ class Text(Embeddable):
...
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        # We ignore the embedding since the embedding can:
+        # - Be lazily acquired, or not used (depending on settings)
+        # - Get ditched when serializing a text for an HTTP request
+        return (
+            self.name == other.name
+            and self.text == other.text
+            and self.doc == other.doc
```
whitead, line 122: Is this one implemented?

whitead: Wait - why doesn't this infinitely recurse? 

if you hit the doc == doc line, then that calls the text == text in the docs, which will call the doc == doc line, etc.

jamesbraza: > Is this one implemented?

There is no `Doc.__eq__` right now, just `Doc.__hash__`. So then `Doc` uses Pydantic's `__eq__`.

> Wait - why doesn't this infinitely recurse?

 I think you may be confusing `Docs` with `Doc`, as `Doc`/`DocDetails` doesn't include a `Text`, so that's why there's no recursion. Also, note tests pass with this change, so we don't hit `RecusionError`.

PR #990, opened 2025-07-03
Change contexts to use ids rather than text

mskarlin:
Gives each context a deterministically generated ID, which is then used for references. 

This replaces the need for complex citation string formatting because the IDs are simple and easy to identify via regexp. The answer which uses the context ids rather than the text names is added into a new `Session` attribute, `raw_answer`, this can be used to find the exact context corresponding to each key. `answer` and `formatted_answer` are left as they were before, but rather than relying on the LLM to populate the citation strings, we deterministically substitute the text name for each citation ID, after de-duping.

Before this change, if `gather_evidence` was run twice with different questions, we had no way to map a citation to the correct summary. 

This PR also fixes a latent bug I found in the tests where auto-generated dockeys weren't able to be overridden with the metadata based dockey if using `aadd`. This only showed in the tests because I needed to regenerate the cassette for a few tests.


paperqa/docs.py
```diff
@@ -708,6 +719,61 @@ def query(
...
+            # now we replace eligible parentheticals with the deduped names
+            deduped_names = {
+                id_to_name_map.get(key, "") for key in get_citation_keys(parentheical)
+            }
+
+            # replace the parenthetical with the deduped names
+            if deduped_names:
+                formatted_without_references = formatted_without_references.replace(
+                    parentheical,
+                    f"{', '.join(deduped_names)}",
```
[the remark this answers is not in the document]
mskarlin: No I think I want this behavior because the content within the parenthesis should be replaced, it's fine if it's done globally

paperqa/docs.py
```diff
@@ -267,9 +267,11 @@ async def aadd(  # noqa: PLR0912
         """Add a document to the collection."""
         all_settings = get_settings(settings)
         parse_config = all_settings.parsing
+        dockey_from_hash = False
```
jamesbraza: Can we rename this to `dockey_is_hash`? Or does that not make sense

paperqa/docs.py
```diff
@@ -692,6 +703,61 @@ def query(
...
+
+        # strip out any leftover hallucinated citations
+        included_keys = get_citation_keys(answer)
+        for hallucinated_key in set(included_keys) - set(id_to_name_map):
+            formatted_without_references = formatted_without_references.replace(
+                hallucinated_key, ""
+            )
+
+        formatted_with_references = (
+            f"Question: {session.question}\n\n{formatted_without_references}\n"
```
jamesbraza: Why the trailing newline, can we remove them?

paperqa/docs.py
```diff
@@ -692,6 +703,61 @@ def query(
             )
         )
 
+    @staticmethod
+    def format_answers(
+        session: PQASession, answer: str, post: str | None = None
```
jamesbraza: Can you update `post` variable name to be more verbose?

I know it has meaning elsewhere, but in this `staticmethod` it has lost that meaning.

Also, why aren't callers handling the `answer`/`post` merging?

paperqa/docs.py
```diff
@@ -692,6 +703,61 @@ def query(
...
+            )
+
+        formatted_with_references = (
+            f"Question: {session.question}\n\n{formatted_without_references}\n"
+        )
+
+        if bib:
+            formatted_with_references += f"\nReferences\n\n{bib}\n"
+
+        return (formatted_without_references, formatted_with_references, bib)
```
jamesbraza: [nit] extra parens

paperqa/types.py
```diff
@@ -144,6 +166,10 @@ class PQASession(BaseModel):
     id: UUID = Field(default_factory=uuid4)
     question: str
     answer: str = ""
+    raw_answer: str = Field(
+        default="",
+        description=("Raw answer from the LLM, including context IDs."),
```
jamesbraza: [nit] extra parens

paperqa/utils.py
```diff
@@ -150,6 +150,7 @@ def extract_score(text: str) -> int:
     return 5
 
 
+# no longer used, but kept for backwards compatibility
```
jamesbraza: Do you mind starting a deprecation on this for PQA v6?

paperqa/utils.py
```diff
@@ -170,6 +171,10 @@ def get_citenames(text: str) -> set[str]:
     return set(values)
 
 
+def get_citation_keys(text: str) -> set[str]:
```
whitead: So does citation key = pqac-XXX?
then what is the name for Example2012Foo pg 15.?

Do we need to rename any other functions?

mskarlin: Yea -- but they are internal the flow of generating the answer. So from an end users perspective answer / formatted answer look the same. The name you have above would be the `Context.text.name`  where as the citation key  generated in the `raw_answer` would be from `Context.id`

paperqa/utils.py
mskarlin: Because it's now an unused free function, I don't know if a deprecation notice makes sense, I think maybe I'll just cut it.

paperqa/docs.py
mskarlin: Yea lemme rename, it should really be `dockey_is_autogenerated`.

paperqa/prompts.py
```diff
@@ -25,23 +25,22 @@
...
+    "- (pqac-d79ef6fa, pqac-0f650d59) \n"
+    "- (pqac-d79ef6fa) \n"
     "## Invalid citation examples: \n"
-    "- Example2024Example pages 3-4 and pages 4-5 \n"
-    "- Example2024Example (pages 3-4) \n"
-    "- Example2024Example pages 3-4, pages 5-6 \n"
-    "- Example2024Example et al. (2024) \n"
-    "- Example's work (pages 17–19) \n"  # noqa: RUF001
-    "- (pages 17–19) \n"  # noqa: RUF001
+    "- (pqac-d79ef6fa and pqac-0f650d59) \n"
```
jamesbraza, line 38: Can you review in the `README.md` or `docs/` if there's anything we need to update?

paperqa/docs.py
mskarlin: good point -- i'll make the caller do it

paperqa/prompts.py
mskarlin: yea i'll give it a once over -- for the end user, they shouldn't notice this change using our standard objects.

paperqa/utils.py
mskarlin: To your point -- I'm gonna rename this `get_citation_ids`

paperqa/docs.py
```diff
@@ -691,6 +702,57 @@ def query(
             )
         )
 
+    @staticmethod
+    def format_answers(session: PQASession, answer: str) -> tuple[str, str, str]:
+        """Format the answer for display with and without references/question."""
+        formatted_without_references = f"{answer}"
```
jamesbraza:
```suggestion
        formatted_without_references = answer
```

I think this is an unnecessary string cast

paperqa/docs.py
```diff
@@ -691,6 +702,57 @@ def query(
...
+
+        id_to_name_map = {c.id: c.text.name for c in session.contexts}
+        name_to_citation_map = {
+            c.text.name: c.text.doc.formatted_citation for c in session.contexts
+        }
+        name_bib = {}
+
+        for parenthetical in re.findall(r"\(([^)]*)\)", formatted_without_references):
```
jamesbraza: Bonus points for a simple regex 101 link, and feel free to ignore

paperqa/docs.py
```diff
@@ -691,6 +702,57 @@ def query(
             )
         )
 
+    @staticmethod
+    def format_answers(session: PQASession, answer: str) -> tuple[str, str, str]:
```
jamesbraza: Can we make this a method on `PQASession`? It seemingly makes more sense there, and below the usage would be:

```diff
--- self.format_answers(session, answer_text)
+++ session.format_answers(answer_text)
```

Also, feel free to ignore this one

jamesbraza: Can you rename this to `format_answer` -- given there is just one answer argument?

And if there's 1+ answers, can you document in the docstring why it's "answers" and not "answer"

paperqa/prompts.py
jamesbraza: I just gave a once-over, it looks good. Resolving this one

paperqa/settings.py
```diff
@@ -284,8 +284,8 @@ def get_formatted_variables(s: str) -> set[str]:
 class PromptSettings(BaseModel):
     model_config = ConfigDict(extra="forbid", validate_assignment=True)
 
-    # MLA parenthetical in-text citation, SEE: https://nwtc.libguides.com/citations/MLA#s-lg-box-707489
```
jamesbraza, line 287: Okay actually we still do use MLA parenthetical in-text citation format, can you preserve this link?

```python
# citations are inserted with Context.id as follows,
# using MLA parenthetical in-text citation styling
# SEE: https://nwtc.libguides.com/citations/MLA#s-lg-box-707489
```

paperqa/types.py
```diff
@@ -143,10 +147,28 @@ class Context(BaseModel):
     text: Text
     score: int = 5
 
+    CONTENT_ENCODING_LENGTH: ClassVar[int] = 500
+    CONTENT_ID_HASHSIZE: ClassVar[int] = 8
```
jamesbraza:
```suggestion
    CONTEXT_ENCODING_LENGTH: ClassVar[int] = 500  # chars
    ID_HASH_LENGTH: ClassVar[int] = 8  # chars
```

Okay a few problems here:
- You have "length" and "hashsize" -- but actually they're both character units -- let's standardize
- "content" is interpretable to apply to the `question`, `context`, `text`, etc and is confusing.

Let's rename to more intuitive versions:

- `CONTEXT_ENCODING_LENGTH`
- `ID_HASH_LENGTH`

tests/test_paperqa.py
```diff
@@ -72,42 +71,6 @@ async def docs_fixture(stub_data_dir: Path) -> Docs:
     return docs
 
 
-def test_get_citations() -> None:
```
jamesbraza, line 75: Why are we removing this test? It seems useful

mskarlin: the underlying function is deleted as it's deprecated and not used anywhere anymore

jamesbraza: Ah okay gotchu

PR #993, opened 2025-07-07
Fixing `exclude_text_filter`'s mix up of `name` and `Text`

jamesbraza:
Previously, `Docs.aget_evidence`'s `exclude_text_filter`:

1. Centered on `name` up front
2. Later excluded on `Text`

This is flawed in two ways:

1. It's possible that `name` can be a substring -- so this was brittle
2. It was broken (confusing `str` with `Text`)

This PR fixes both issues

### DIFF
--- a/paperqa/docs.py
+++ b/paperqa/docs.py
@@ -7,7 +7,7 @@ import re
 import tempfile
 import urllib.request
 import warnings
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Collection, Sequence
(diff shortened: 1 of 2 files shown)


jamesbraza:
I realized this `exclude_text_filter` bug led to the evidence gathering process:

- Having a backdoor to 2+ `Context` using the same `Text`, which is actually useful
- Has performance now coupled with this bug, due to pulling in more `matches`

So I am going to close this for further thought.

PR #1003, opened 2025-07-10
Exposing environment on-reset hook for `Docs` preparation

jamesbraza:
The base class will clear docs, but it's possible subclasses would want to do something else (e.g. clear then add media)

### DIFF
--- a/paperqa/agents/env.py
+++ b/paperqa/agents/env.py
@@ -235,7 +235,13 @@ class PaperQAEnvironment(Environment[EnvironmentState]):
             embedding_model=self._embedding_model,
         )
 
-    def make_initial_state(self) -> EnvironmentState:
+    async def _reset_docs(self) -> None:
+        """Hook to reset the docs when creating the initial state."""
+        self._docs.clear_docs()
+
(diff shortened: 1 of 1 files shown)


paperqa/agents/env.py
```diff
@@ -264,8 +270,7 @@ async def reset(self) -> tuple[list[Message], list[Tool]]:
         # NOTE: don't build the index here, as sometimes we asyncio.gather over this
         # method, and our current design (as of v5.0.10) could hit race conditions
         # because index building does not use file locks
-        self._docs.clear_docs()
-        self.state, self.tools = self.make_initial_state(), self.make_tools()
+        self.state, self.tools = (await self.make_initial_state()), self.make_tools()
```
maykcaldas, line 273: Is this the only place where make_initial_state is used (maybe the environment constructor)?
Just making sure it is awaited everywhere

maykcaldas (approved):

paperqa/agents/env.py
jamesbraza: In this repo yes, other repos there will need to be changes

PR #1015, opened 2025-07-18
Adding variable timeouts for clients

mskarlin:
Our default timeouts were not long enough for Crossref. This ups the default and makes it a configurable variable.

### DIFF
--- a/paperqa/clients/crossref.py
+++ b/paperqa/clients/crossref.py
@@ -38,7 +38,9 @@ logger = logging.getLogger(__name__)
 
 CROSSREF_HOST = "api.crossref.org"
 CROSSREF_BASE_URL = f"https://{CROSSREF_HOST}"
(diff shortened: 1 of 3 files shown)


nadolskit (approved):

paperqa/clients/crossref.py
```diff
@@ -38,7 +38,9 @@
 
 CROSSREF_HOST = "api.crossref.org"
 CROSSREF_BASE_URL = f"https://{CROSSREF_HOST}"
-CROSSREF_API_REQUEST_TIMEOUT = 5.0
+CROSSREF_API_REQUEST_TIMEOUT = float(
+    os.environ.get("CROSSREF_API_REQUEST_TIMEOUT", "10.0")
```
jamesbraza, line 42: I like the configurability, but why bump from 5-sec to 10-sec default here? Isn't that going to make things worse?

jamesbraza (approved):

PR #1022, opened 2025-07-23
Removed `JSONType` for Pydantic's equivalent `JsonValue`

jamesbraza:
https://github.com/Future-House/paper-qa/pull/962 added in a JSON type alias, but as seen in [this CI run](https://github.com/Future-House/paper-qa/actions/runs/16453474103/job/46504464218):

```none
tests/test_cli.py::test_cli_ask
  /home/runner/work/paper-qa/paper-qa/paperqa/agents/search.py:91: TypeHintWarning: Skipping type check against 'JSONType'; this looks like a string-form forward reference imported from another module
    return json.loads(data)
```

We appear to have issues with it. Let's just use Pydantic's built-in one, so there's less code here

### DIFF
--- a/paperqa/agents/models.py
+++ b/paperqa/agents/models.py
@@ -5,27 +5,17 @@ import logging
 import time
 from contextlib import asynccontextmanager
 from enum import StrEnum
-from typing import Any, ClassVar, Protocol, TypeAlias, cast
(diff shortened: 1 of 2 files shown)


whitead (approved):
-10 lines. I like

Issue #1005, opened 2025-07-14
Dockey confilct because of same Doi


lucky0218:
In the default settings, if two papers have the same Doi and the user doesn't set the doc_id, they'll have the same dockey, leading to one paper unable to be found.

`
elif "doc_id" not in data or not data["doc_id"]:  # keep user defined doc_ids
      data["doc_id"] = encode_id(uuid4())
        if "dockey" in data.get(
            "fields_to_overwrite_from_metadata",
            DEFAULT_FIELDS_TO_OVERWRITE_FROM_METADATA,
        ) and ("dockey" not in data or not data["dockey"]):
            data["dockey"] = data["doc_id"]
`

It may happen on the circumstance of two papers having the same Doi (e.g. one paper is normal thesis, the other one is its supplement information, like those published on ACS)

jamesbraza:
Thanks for the report. From my understanding of DOIs, there is an assumption of uniqueness.

Can you point out where one DOIs corresponds with two separate PDF papers?

jamesbraza:
> Can you point out where one DOIs corresponds with two separate PDF papers?

> > It may happen on the circumstance of two papers having the same Doi (e.g. one paper is normal thesis, the other one is its supplement information, like those published on ACS)

Nevermind to this, I should have read the full issue. I will get back to you on this shortly.

jamesbraza:
Hi @lucky0218 I guess what we can do is have the `doc_id` be a composite of fields from `paperqa.DocDetails`.

Do you mind sharing the `DocDetails` you get for the main text and the supp?  Mainly, I want to diff the two `paperqa.DocDetails` and see what fields we can use to make a composite key.

For example, does the `title` change between the two?

lucky0218:
Of course not!

[...]

PR #1029, opened 2025-07-24
Adding `Doc.content_hash` and integrating for unique `DocDetails.doc_id`

jamesbraza:
This PR:
- Enables `DocDetails.doc_id` deduplication across files sharing the beginning (e.g. a main text and a supplemental information whose metadata is inferred to be identical)
    - Adds a `content_hash` field on `Doc` (which we already collect during `Docs.aadd`)
    - Expands `DocDetails.doc_id` to be a composite key of DOI and content hash
    - Adds test coverage here in `test_duplicate` that does not work on current `main`
- Substantially documents `DocDetails` metadata upgrades and `file_location`

Closes https://github.com/Future-House/paper-qa/issues/1005

### DIFF
--- a/tests/cassettes/test_docs_lifecycle.yaml
+++ b/tests/cassettes/test_docs_lifecycle.yaml
@@ -9,6 +9,71 @@ interactions:
         connection:
           - keep-alive
         host:
+          - api.crossref.org
+        user-agent:
+          - python-httpx/0.28.1
+      method: GET
+      uri: https://api.crossref.org/works?mailto=example@papercrow.ai&query.title=National+Flag+of+Canada+Day&rows=1
+    response:
+      body:
+        string: !!binary |
+          H4sIAAAAAAAA/61UTY/aMBD9K5HPcWKHJJBcWVWqtP1QRS8lHEwygEUSp7azgBD/veOwXSjsblWp
+          OcXjj3nz3sw7EmOF7Q3JidoSnzRgjFgDtYcOMLZTektraezV1hNoI1WLuzxgAbvskPxIVqIEi68d
+          Tz6xyoqaajB97UIZHycZ84m00OByfiSyrWAPlbtXCQu0E9odnM8jFkV+7EeLhX/esbJxcFycspiy
+          aMbTnCV5kvzA/G4Xy2g6kvM0nmQ8wW/CGWLQsAINbQm0VH1rSY4Aun6JJW1A44vfWznUYw+eWnkz
+          pVVrlfcVQRt8uMQVtJZWqhGyHXA+/80RWamVMY1AhvC41bK0Ay0rURs4XdJUtNPSpb6vkqc+j/wR
+          Xyzw/MOXj45TFoz4aFKE2XjC4zhKI56MsGKeuULPqiyV2tJyIzqLNTgcgO++SiOf+JnP03se+YSy
+          jPJ0xljOJnmc3vKYjMYsSZMoZczx2A36kpglNB6leNaoXpcuNHUsIM0Yk4a+8F3R5eGKcyttjafn
+          hKeB9wk5AyvbtWc3gKuuBu8RxCr3Zrie9lqq3nhTYcCJ8lk4XkXtfajF2gWmohWV8B7EgWBdHaaU
+          +wtzQz82y0Fdnmbps4yoGmj6AmMKNSy1GECc33Nv1bLdDp35/dsjXt9Y25m8CIvQnhsjqGCt+wPS
+          HpSqKcInCbsidHLc6FWEb+sY7Jv6qreeNbWwt0W4sX/sXUbtSTmlpQtXyK3oulqW4txww13ayBaL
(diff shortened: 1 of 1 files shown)


src/paperqa/utils.py
```diff
@@ -210,6 +210,14 @@ def encode_id(value: str | bytes | UUID, maxsize: int | None = 16) -> str:
     return hashlib.md5(value).hexdigest()[:maxsize]  # noqa: S324
 
 
+def compute_unique_doc_id(doi: str | None, content_hash: str | None) -> str:
+    if doi:
+        value_to_encode: str = doi.lower() + (content_hash or "")
+    else:
+        value_to_encode = content_hash or str(uuid4())
```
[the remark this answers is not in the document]
jamesbraza: We recompute `doc_id` if there's changes in `DocDetails.__add__`, so I believe we're safe

src/paperqa/types.py
```diff
@@ -1011,6 +1038,14 @@ def __add__(self, other: DocDetails | int) -> DocDetails:  # noqa: PLR0912
...
                 else:
                     merged_data[field] = max(self_value, other_value)
+            elif field == "content_hash" and (  # noqa: PLR0916
+                # Hashes are both present but differ
+                (self_value and other_value and self_value != other_value)
+                # One hash is explicitly disabled (not autopopulated)
+                or (self_value is None or other_value is None)
+            ):
+                # We don't know which to pick, so just discard the value
+                merged_data[field] = None
```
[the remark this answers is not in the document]
jamesbraza: I prefer to keep it in one place, a private method would break LOC

src/paperqa/docs.py
```diff
@@ -265,18 +265,20 @@ async def aadd(  # noqa: PLR0912
         """Add a document to the collection."""
         all_settings = get_settings(settings)
         parse_config = all_settings.parsing
+        content_hash = md5sum(path)
         dockey_is_content_hash = False
         if dockey is None:
-            # md5 sum of file contents (not path!)
```
jamesbraza, line 271: I moved this comment to `Doc.content_hash`'s `description`, fyi

src/paperqa/clients/__init__.py
```diff
@@ -197,16 +197,16 @@ async def bulk_query(
         )
 
     async def upgrade_doc_to_doc_details(self, doc: Doc, **kwargs) -> DocDetails:
-
-        # note we have some extra fields which may have come from reading the doc text,
-        # but aren't in the doc object, we add them here too.
-        extra_fields = {
```
jamesbraza, line 227: The usage of "extra" here is an overloaded term:

- Extra in a Pydantic context (and we have `DocDetails.other`, to add to the confusion)
- Extra in a query context

Paired with vagueness such as "some extra fields", I decided to rename and reworded this whole section

tests/test_paperqa.py
```diff
@@ -648,16 +651,44 @@ async def test_duplicate(stub_data_dir: Path) -> None:
...
+    assert len(set(docs.docs.keys())) == 3, "Expected a third document to be added"
+    shorter_flag_day = docs.docs["flag_day_shorter"]
+    assert isinstance(shorter_flag_day, DocDetails)
+    assert shorter_flag_day.doi == common_doi
+    assert all(
+        x in shorter_flag_day.citation.lower() for x in ("wikipedia", "flag")
+    ), "Expected citation to be inferred"
+    assert shorter_flag_day.content_hash
+    assert flag_day.content_hash != shorter_flag_day.content_hash
+    assert flag_day.doc_id != shorter_flag_day.doc_id
```
jamesbraza, line 1045: This whole PR basically enables this singular assertion to pass

src/paperqa/types.py
```diff
@@ -623,10 +638,10 @@ def lowercase_doi_and_populate_doc_id(cls, data: dict[str, Any]) -> dict[str, An
                 if doi.startswith(url_prefix_to_remove):
                     doi = doi.replace(url_prefix_to_remove, "")
             data["doi"] = doi.lower()
-            if "doc_id" not in data or not data["doc_id"]:  # keep user defined doc_ids
-                data["doc_id"] = encode_id(doi.lower())
-        elif "doc_id" not in data or not data["doc_id"]:  # keep user defined doc_ids
-            data["doc_id"] = encode_id(uuid4())
+            if not data.get("doc_id"):  # keep user defined doc_ids
+                data["doc_id"] = compute_unique_doc_id(doi, data.get("content_hash"))
```
mskarlin, line 871: we talked about this in person, but let's move this into an optional feature

mskarlin (commented):
I need to read this in more detail and think about it

src/paperqa/types.py
jamesbraza: I piloted this today some more btw. Moving `compute_unique_doc_id` into something like `Settings.details_id_factory` requires us to pass a throwaway `Settings` reference to all `DocDetails` constructions, which is both awkward and error prone.

Perhaps a less awkward route is having an `bool` environment variable `PQA_USE_OLD_ID_FACTORY` that gets used inside `compute_unique_doc_id`

jamesbraza:
May god bless us all

PR #1034, opened 2025-07-25
Journal quality fixes

mskarlin:
We had some common journals that didn't show up either because of strange characters or non canonical name usage. This fixes and adds test for these outliers. 

Note a raw data update could be done but it looks to be almost identical -- none of these issues are fixed with a data refresh.

### DIFF
--- a/src/paperqa/clients/client_data/journal_quality.csv
+++ b/src/paperqa/clients/client_data/journal_quality.csv
@@ -5143,6 +5143,8 @@ annual review of neuroscience,2
 annual review of nuclear and particle science,1
 annual review of nutrition,3
 annual review of pathology-mechanisms of disease,2
+annual review of pathology: mechanisms of disease,2
+annual review of pathology,2
 annual review of pharmacology and toxicology,3
 annual review of physical chemistry,1
 annual review of physiology,3
@@ -6057,6 +6059,15 @@ biochimica et biophysica acta: molecular basis of disease,1
 biochimica et biophysica acta: molecular cell research,1
 biochimica et biophysica acta: proteins and proteomics,1
 biochimica et biophysica acta: reviews on cancer,1
+biochimica et biophysica acta (bba) - bioenergetics,1
+biochimica et biophysica acta (bba) - biomembranes,1
+biochimica et biophysica acta (bba) - gene regulatory mechanisms,1
+biochimica et biophysica acta (bba) - general subjects,1
+biochimica et biophysica acta (bba) - molecular and cell biology of lipids,1
+biochimica et biophysica acta (bba) - molecular basis of disease,1
+biochimica et biophysica acta (bba) - molecular cell research,1
+biochimica et biophysica acta (bba) - proteins and proteomics,1
+biochimica et biophysica acta (bba) - reviews on cancer,1
 biochimie,1
 biochip journal,1
 bioconjugate chemistry,1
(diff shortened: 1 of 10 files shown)


src/paperqa/clients/journal_quality.py
```diff
@@ -18,6 +18,10 @@
 
 
 class JournalQualityPostProcessor(MetadataPostProcessor[JournalQuery]):
+
+    # these will be deleted from any journal names before querying
+    PHRASES_TO_REMOVE: ClassVar[list[str]] = ["amp;"]
```
jamesbraza: Can you add the prefix `CASEFOLD_` to this variable name

src/paperqa/clients/journal_quality.py
```diff
@@ -49,8 +59,9 @@ async def _process(
             dockey=doc_details.dockey,  # ensure dockey is preserved
             source_quality=max(
                 [
-                    self.data.get(query.journal.casefold(), DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
-                    self.data.get("the " + query.journal.casefold(), DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
+                    self.data.get(journal_query, DocDetails.UNDEFINED_JOURNAL_QUALITY),  # type: ignore[union-attr]
```
jamesbraza: Can you remove the `list` here, and just have it be flat inside `max`

src/paperqa/clients/client_data/journal_quality.csv
```diff
@@ -31738,7 +31749,6 @@ proceedings of international conference on the advancement of steam,0
 selected papers of internet research,0
 bat research news,1
 imerides endymasiologias - praktika,1
-scientific reports,0
```
whitead, line 31741: why did you delete it?

src/paperqa/clients/client_data/journal_quality.csv
```diff
@@ -17342,7 +17353,7 @@ proceedings of the lfg conference,1
 proceedings of the linnean society of new south wales,1
 proceedings of the london mathematical society,3
 proceedings of the national academy of sciences india section b: biologicalsciences,1
-proceedings of the national academy of sciences of the united states of america,3
```
whitead, line 17345: Will a complete journal name (including of the united states of amercia) now fail?

Will the abbreviation work? PNAS?

mskarlin: yea and no, it won't. We don't use the full name anywhere as far as I can tell -- ideally this is some partial matching or jaccard overlap calculation, but we'd need to do some more testing of the precision / recall because so many of the journal titles have overlapping names

tests/test_clients.py
```diff
@@ -711,3 +711,35 @@ async def test_arxiv_doi_is_used_when_available() -> None:
...
+        ("10.1038/s41598-018-27044-6", 1),
+        ("10.1073/pnas.1205508109", 3),
+        ("10.1186/1471-2148-11-4", 2),
+        ("10.1016/j.semcdb.2016.08.024", 1),
+        ("10.1146/annurev.pathol.4.110807.092311", 2),
+        ("10.1016/j.bbcan.2023.188947", 1),
+    ],
+)
+@pytest.mark.asyncio
+async def test_tricky_journal_quality_results(doi: str, score: int) -> None:
```
jamesbraza, line 729: Can you either:
- Add a docstring just stating the meaning of "tricky"
- Add a comment (or `pytest.param` with `id="reason"`) next to each test case stating what's tricky

src/paperqa/clients/client_data/journal_quality.csv
mskarlin: there were two, one was 0 and the other was 1

jamesbraza (approved):
Nice work

src/paperqa/clients/client_data/journal_quality.csv
mskarlin: I'll leave in the above -- and make a new line for it -- also i'll just add pnas below it

PR #1035, opened 2025-07-26
Allow users to skip citation stripping from evidence

mskarlin:
Users may have text chunks which reference other papers, and that's a necessary part of the context. For example, in a multi-hop question asking about how one work extends a prior work. We strip these citations out currently, and the resulting gen_answer call not understand what paper is being referenced.

### DIFF
--- a/src/paperqa/settings.py
+++ b/src/paperqa/settings.py
@@ -124,6 +124,10 @@ class AnswerSettings(BaseModel):
         default=False,
         description="Whether to group contexts by question when generating answers.",
     )
+    skip_evidence_citation_strip: bool = Field(
+        default=False,
+        description="Whether to skip stripping citations from evidence.",
+    )
 
     @model_validator(mode="after")
     def _deprecated_field(self) -> Self:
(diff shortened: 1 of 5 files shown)


whitead:
But aren't the majority of these numeric instead of parenthetic citations? We probably need to actually reconcile the citations with the known references to do this properly.

mskarlin:
Certainly not the majority in the dataset I'm working with -- but if we wanted to reconcile them, it's tough as there will be papers that aren't in any Text / Doc object obviously. I need to think on this more.

src/paperqa/settings.py
```diff
@@ -124,6 +124,10 @@ class AnswerSettings(BaseModel):
         default=False,
         description="Whether to group contexts by question when generating answers.",
     )
+    skip_evidence_citation_strip: bool = Field(
+        default=False,
+        description="Whether to skip stripping citations from evidence.",
```
jamesbraza, line 129: Can you add a brief sentence here stating why one would want to do this? Otherwise basically people will need to read the source or dig up this PR

```none
Whether to skip stripping citations from evidence. Opting out of stripping is useful in multi-hop tasks where evidence may cite other works.
```

jamesbraza (approved):
Code and behaviors look great, note I defer to you and Andrew here per your discussion on full-blown citation reconciliation

whitead:
@mskarlin The majority of articles I read have numeric citations, but maybe this will be fine. Guess we can at least benchmark it.

mskarlin:
Merging this since it's a default OFF setting

PR #1039, opened 2025-07-29
Move context str generation into configurable function

mskarlin:
Users may want custom control over how contexts are generated and munged before inserting into the final answer. This was previously locked into the `aquery` code, but this exposes it as a configurable function. 

We may not be the best, but by god, we'll be the most configurable.

### DIFF
--- a/src/paperqa/docs.py
+++ b/src/paperqa/docs.py
@@ -29,7 +29,7 @@ from paperqa.llms import (
 )
 from paperqa.prompts import CANNOT_ANSWER_PHRASE
 from paperqa.readers import read_doc
-from paperqa.settings import MaybeSettings, get_settings
+from paperqa.settings import MaybeSettings, Settings, get_settings
 from paperqa.types import Context, Doc, DocDetails, DocKey, PQASession, Text
 from paperqa.utils import (
     citation_to_docname,
@@ -710,7 +710,7 @@ class Docs(BaseModel):  # noqa: PLW1641  # TODO: add __hash__
             )
         )
 
-    async def aquery(  # noqa: PLR0912
+    async def aquery(
         self,
         query: PQASession | str,
         settings: MaybeSettings = None,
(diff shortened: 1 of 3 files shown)


src/paperqa/settings.py
```diff
@@ -128,6 +160,11 @@ class AnswerSettings(BaseModel):
         default=False,
         description="Whether to skip stripping citations from evidence.",
     )
+    context_str_fn: ContextStrFn = Field(
```
jamesbraza: Can you add to the settings table in the README

jamesbraza: When I hear `context_str_fn`, I just immediately think of `Context.__str__`. What do you think of naming this a  `context_serializer`

src/paperqa/settings.py
```diff
@@ -63,8 +73,30 @@
...
 
 
+@runtime_checkable
+class ContextStrFn(Protocol):
+    """Protocol for generating a context string from settings and context."""
+
+    def __call__(
+        self,
+        settings: "Settings",
+        contexts: list[Context],
```
jamesbraza: Can we have this be `Sequence[Context]` to avoid mutability issues and also allow things like `tuple[Context, ...]`

src/paperqa/docs.py
```diff
@@ -905,3 +844,79 @@ async def aquery(  # noqa: PLR0912
         session.populate_formatted_answers_and_bib_from_raw_answer()
 
         return session
+
+
+def default_context_str_fn(
```
jamesbraza: What do you think of making this a method of `Settings` or a `classmethod` of `Context`?

A method of `Settings` may also help with circular imports

tests/test_paperqa.py
```diff
@@ -566,6 +566,30 @@ async def test_query(docs_fixture) -> None:
     await docs_fixture.aquery("Is XAI usable in chemistry?", settings=settings)
 
 
+@pytest.mark.asyncio
+async def test_custom_context_str_fn(docs_fixture) -> None:
+
+    def custom_context_str_fn(
```
jamesbraza: Can you add an assertion checking this `isinstance(x, ContextStrFn)`

tests/test_paperqa.py
```diff
@@ -566,6 +566,30 @@ async def test_query(docs_fixture) -> None:
...
+    settings = Settings(
+        prompts={"answer_iteration_prompt": None},
+        answer={"context_str_fn": custom_context_str_fn},
+    )
+
+    session = await docs_fixture.aquery(
+        "Is XAI usable in chemistry?", settings=settings
+    )
+    assert (
+        session.context == "TEST OVERRIDE"
```
jamesbraza, line 592: Can you add an assertion on `session.contexts` too

src/paperqa/docs.py
mskarlin: sure it's just so much core core that now lives in a settings object, which is a little suspicious

src/paperqa/settings.py
mskarlin: but the input will always specifically be a list since the calling is hard coded

jamesbraza: It's sort of like "defensive driving" -- just better tees things up for future features (we may use `context_str_fn` for other things someday), and it encourages callers to not mutate `contexts`

src/paperqa/docs.py
jamesbraza: Yeah I know what you mean, sure let's keep it separate 👌

tests/test_paperqa.py
```diff
@@ -566,6 +567,34 @@ async def test_query(docs_fixture) -> None:
...
+    ), "Expected custom context string to be returned."
+
+    assert session.contexts, "Expected contexts to be present in session."
```
jamesbraza: Will any of the `Context` have the string `"TEST OVERRIDE"` in it? I think I was hoping for more of an assertion here besides the presence of contexts

Imo it's confusing that `Session.context` would be `"TEST OVERRIDE"` but then none of the `Context` have it, but maybe I am misunderstanding

src/paperqa/settings.py
mskarlin: i suppose that's true -- done!

src/paperqa/settings.py
```diff
@@ -791,6 +815,11 @@ class Settings(BaseSettings):
         exclude=True,
         frozen=True,
     )
+    custom_context_serializer: ContextSerializer | None = Field(
+        default=None,
+        description="Function to turn settings and contexts into an answer context str.",
```
jamesbraza: Why not just have this be `custom_context_serializer: ContextSerializer = Field(default=context_serializer, ...)`? We have the method here, let's just use that over `None` as a default

Alternately, can you update the description to mention it's auto-populated:

```python
description="Optional function ..., and if left unspecified it's auto-populated.",
```

jamesbraza: Imo "answer context str" in the `description` is a bit of a word salad and not that meaningful -- can you attempt to generalize or ground the wording a bit

tests/test_paperqa.py
mskarlin: So the “context” in the final session object is just a single string that was used to generate your answer. It’s what I’m referring to in the description in the table “answer context string” you will have many more contexts than what shows up in your context string. It’s a formatted and filtered version. I just added that assertion to make you happy but I’m not actually sure what to assert on the contexts given the function of this code.

mskarlin: The idea here wouldn’t be to ever override your contexts you just specify how to use them to generate your final prompt. This is what’s blocking me atm.

src/paperqa/settings.py
mskarlin: IMO answer context string is quite precise and exactly what it is — it was the variable name for it in the code and it’s in the answer prompt.

tests/test_paperqa.py
jamesbraza: Alright I was misunderstanding here a bit, now I see. Yeah basically the `contexts` are an input to `ContextSerializer`, but may not matter at all in the end.

Understanding this now, I think we should probably remove `assert session.contexts` from the test -- it's sort of besides the point. Alternately what you could do is this:

```python
    spy_context_fn = MagicMock(return_value="TEST OVERRIDE")

    def custom_context_str_fn(
        settings: Settings,  # noqa: ARG001
        contexts: list[Context],  # noqa: ARG001
        question: str,  # noqa: ARG001
        pre_str: str | None = None,  # noqa: ARG001
    ) -> str:
        return spy_context_fn(contexts)

    ...

    assert (
        session.context == "TEST OVERRIDE"
    ), "Expected custom context string to be returned."
    spy_context_fn.assert_called_once_with(session.contexts)
```

src/paperqa/settings.py
```diff
@@ -63,8 +74,21 @@
 _EnvironmentState: TypeAlias = Any
 
 
+@runtime_checkable
+class ContextSerializer(Protocol):
+    """Protocol for generating a context string from settings and context."""
+
+    def __call__(
```
jamesbraza: Wdyt of making this `async` 🤔 

Mainly, in theory one could call an LLM inside of this

src/paperqa/settings.py
mskarlin: I think i'm going to just comment, because `context_serializer` is a method on `Settings` I can't make it a default without instantiation. I could do it after the fact with a `model_post_init` which is kind of messy.

src/paperqa/settings.py
mskarlin: yup, done

PR #1046, opened 2025-08-05
Image reader and image support in `gather_evidence`

jamesbraza:
This PR is the first step in PaperQA becoming multimodal:

1. Adds reader support for images
2. Converts `Docs` to also store images (with text and metadata) as a `ParsedMedia`
3. Expands `gather_evidence` tool to include images in the `Context`-generation prompt
4. Adds tests of the prior three steps


src/paperqa/types.py
```diff
@@ -431,15 +446,83 @@ class ParsedMetadata(BaseModel):
...
+        return UUID(int=uuid_int)
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, ParsedMedia):
+            return NotImplemented
+        return (
+            self.index == other.index
+            and self.data == other.data
+            and self.text == other.text
+            and json.dumps(self.info) == json.dumps(self.info)
```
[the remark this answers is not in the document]
jamesbraza: Nice catch, just implemented this locally

src/paperqa/types.py
```diff
@@ -431,15 +446,83 @@ class ParsedMetadata(BaseModel):
...
+    info: dict[str, JsonValue | tuple[float, ...] | bytes] = Field(
+        default_factory=dict,
+        description=(
+            "Optional image metadata. This may come from image definitions sourced from"
+            " the PDF, or attributes of custom pixel maps."
+        ),
+    )
+
+    def __hash__(self) -> int:
+        return hash((self.index, self.data, self.text, json.dumps(self.info)))
```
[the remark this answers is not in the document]
jamesbraza: Nice catch, just implemented this locally

src/paperqa/types.py
```diff
@@ -431,15 +446,83 @@ class ParsedMetadata(BaseModel):
...
+        with Path(path).open("wb") as f:
+            f.write(self.data)
+
+
 class ParsedText(BaseModel):
     """All text from a document read, before chunking."""
 
-    content: dict[str, str] | str | list[str] = Field(
+    content: (
+        dict[str, str] | str | list[str] | dict[str, tuple[str, list[ParsedMedia]]]
```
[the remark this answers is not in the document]
jamesbraza: YAGNI on this, I don't want to create more types

src/paperqa/types.py
```diff
@@ -448,6 +531,8 @@ class ParsedText(BaseModel):
             "\n- `dict[str, str]` (e.g. page number -> page text) for PDFs."
             "\n- `str` for text files."
             "\n- `list[str]` for line-by-line parsings."
+            "\n- `dict[str, tuple[str, list[ParsedImage]]]` (e.g. page number"
```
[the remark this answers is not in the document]
jamesbraza: Nice catch, just implemented this locally

src/paperqa/prompts.py
```diff
@@ -1,20 +1,26 @@
...
     " question. Stay detailed; report specific numbers, equations, or direct quotes"
     ' (marked with quotation marks). Reply "Not applicable" if the excerpt is'
     " irrelevant. At the end of your response, provide an integer score from 1-10 on a"
     " newline indicating relevance to question. Do not explain your score.\n\nRelevant"
     " Information Summary ({summary_length}):"
 )
+# This prompt template integrates with `text` variable of the above `summary_prompt`
+text_with_tables_prompt_template = (
+    "{text}\n\n------------\n\nMarkdown tables from {citation}."
+    " If the markdown is garbled, refer to the images"
```
mskarlin:
```suggestion
    " If the markdown is poorly formatted, defer to using the images"
```

src/paperqa/readers.py
```diff
@@ -48,28 +64,34 @@ def chunk_pdf(
             f" {doc.dockey}, either empty or corrupted."
         )
 
-    for page_num, page_text in parsed_text.content.items():
+    def make_kwargs(lower_page: str, upper_page: str) -> dict[str, Any]:
+        media: list[ParsedMedia] = []
+        for pg_num in range(int(lower_page), int(upper_page) + 1):
+            pg_contents = cast(dict, parsed_text.content)[str(pg_num)]
```
mskarlin: could we make this an input to avoid using encapsulation, makes the flow more explicit when called below

src/paperqa/readers.py
```diff
@@ -30,6 +32,20 @@ def __call__(
     ) -> ParsedText: ...
 
 
+async def parse_image(path: str | os.PathLike, **_) -> ParsedText:
```
mskarlin: I think we need some defensive error handling here given how often PDFs throw weird things at us -- could we do a size check (like how PDFs sometimes come out to billions of characters), and wrap calls to this with a `ImpossibleParsingError` if something goes awry reading the bytes.

tests/test_paperqa.py
```diff
@@ -1321,6 +1329,96 @@ async def test_chunk_metadata_reader(
...
+    assert parsed_image.to_id() == parsed_image.to_id(), "Expected same ID"
+    assert len({parsed_image, parsed_image}) == 1, "Expected shared hash"
+    assert not parsed_image.text, "Expected no text for later assertions to make sense"
+    assert (
+        parsed_image.info.get("type") != "table"
+    ), "Expected no table for later assertions to make sense"
+
+    # Test how self-comparisons work
+    ((_, (parsed_image2,)),) = cast(
+        dict, (await parse_image(png_path)).content
```
mskarlin, line 1353: Can you test parsing several in an async gather to ensure we don't hit any race conditions during the file-reading.

mskarlin (approved):

src/paperqa/prompts.py
jamesbraza: Nice suggestion, implemented just now

src/paperqa/readers.py
jamesbraza: Yeah np, I just decomposed out a `_make_chunk` function

tests/test_paperqa.py
jamesbraza: Yeah I went ahead just now and added a `validate_image` helper function, and used it in testing with an `asyncio.gather` to confirm we can read in 10 images at once, checking the concurrently-read images are both identical and valid

src/paperqa/readers.py
jamesbraza: Alright I added an optional `validator` argument to `parse_image`, which will raise an `ImpossibleParsingError` if the validation fails.

Also, let's say one didn't use a validator, then the invalid images will then degrade `gather_evidence`. So I also created an `AnswerSettings.evidence_text_only_fallback` setting to allow context creation without images. Users can opt into this setting if desired.

PR #1077, opened 2025-08-26
Fixing `gen_answer` failover leaving `raw_answer` blank

jamesbraza:
https://github.com/Future-House/paper-qa/pull/1073 was suboptimal in that if the `EmptyDocsError` route was taken, the `PQASession`'s `raw_answer`/`answer_reasoning` would be left unpopulated without `patch`ing in-place. This PR:
- Keeps the specialization in the returned tool response message
- While allowing the code to populate `raw_answer`/`answer_reasoning` in a normal flow

### DIFF
--- a/src/paperqa/agents/tools.py
+++ b/src/paperqa/agents/tools.py
@@ -309,9 +309,6 @@ class GenerateAnswer(NamedTool):
         Args:
             state: Current state.
         """
-        if not state.docs.docs:
-            raise EmptyDocsError("Not generating an answer due to having no papers.")
-
         logger.info(f"Generating answer for '{state.session.question}'.")
 
         if f"{self.TOOL_FN_NAME}_initialized" in self.settings.agent.callbacks:
--- a/src/paperqa/docs.py
+++ b/src/paperqa/docs.py
@@ -742,7 +742,7 @@ class Docs(BaseModel):  # noqa: PLW1641  # TODO: add __hash__
         contexts = session.contexts
         if answer_config.get_evidence_if_no_contexts and not contexts:
             session = await self.aget_evidence(
-                query=session,
+                session,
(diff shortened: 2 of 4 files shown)


src/paperqa/docs.py
```diff
@@ -774,9 +774,10 @@ async def aquery(
             pre_str=pre_str,
         )
 
-        if len(context_str.strip()) < 10:  # noqa: PLR2004
```
jamesbraza, line 777: Given empty `context_str` is 11-chars, we couldn't hit this case actually

src/paperqa/docs.py
```diff
@@ -774,9 +774,10 @@ async def aquery(
...
+        if len(context_str.strip()) <= EMPTY_CONTEXTS:
             answer_text = (
-                f"{CANNOT_ANSWER_PHRASE} this question due to insufficient information."
+                f"{CANNOT_ANSWER_PHRASE} this question due to"
+                f" {'having no papers' if not self.docs else 'insufficient information.'}."
```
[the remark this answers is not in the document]
jamesbraza: This is not string concat, it's a multiline string

src/paperqa/agents/tools.py
```diff
@@ -309,9 +309,6 @@ async def gen_answer(self, state: EnvironmentState) -> str:
         Args:
             state: Current state.
         """
-        if not state.docs.docs:
-            raise EmptyDocsError("Not generating an answer due to having no papers.")
```
jamesbraza, line 313: Note that `state.session.raw_answer` was left untouched here. This is fine in normal `gen_answer` use, but is not okay in the truncation's `gen_answer` failover. So this is a bug this PR fixes

mskarlin (approved):
lgtm

PR #1078, opened 2025-09-03
S2 metadata retrieval: fallback to exact title match if authors do not match

sidnarayanan:
Sometimes the author check can have false negatives, so this implements a fallback to exact title match. To avoid introducing false positives, it's only applied if the titles are at least 10 words long.

### DIFF
--- a/src/paperqa/clients/semantic_scholar.py
+++ b/src/paperqa/clients/semantic_scholar.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 from collections.abc import Collection
 from datetime import datetime
 from enum import IntEnum, auto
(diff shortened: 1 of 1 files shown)


tests/test_clients.py
```diff
@@ -99,7 +99,7 @@
             "formatted_citation": (
                 "Andres M. Bran, Sam Cox, Oliver Schilter, Carlo Baldassari, Andrew D."
                 " White, and Philippe Schwaller. Augmenting large language models with"
-                " chemistry tools. Nature Machine Intelligence, 6:525-535, May 2024."
+                " chemistry tools. Nature Machine Intelligence, 6:535, May 2024."
```
sidnarayanan, line 102: This test was failing on main, not caused by this PR.

whitead: This test flip-flops, depending on phase of the moon. Kind of annoying.

src/paperqa/clients/semantic_scholar.py
```diff
@@ -250,27 +253,36 @@ async def s2_title_search(
...
+        and title_similarity < HIGH_TITLE_SIMILARITY_THRESHOLD
+        and not s2_authors_match(authors, data=result)
     ):
+        raise DOINotFoundError(
+            f"Could not find DOI for {title} - author and title disagreement."
+        )
+
+    # If we made it here, either authors were not provided or they matched, so apply
+    # a possibly weakter threshold
+    if title_similarity < title_similarity_threshold:
```
whitead: Previously, we checked for some disagreement for authors and/or relatively weak title disagreement. Both would fail.

Now, if we have authors - we have a very strict requirement (must be exact match) on title and an author match. Or, we can have no authors and then we have a looser title check. Or, we had an author match and can have a looser title match.

I think my issue is that when we have no authors, I would prefer a tighter title match rather than a looser. Maybe you could break it out into an author/no author if statement. Then you can apply author match + weak or no author match + strict in author branch and strict title in no author branch

sidnarayanan: > if we have authors - we have a very strict requirement (must be exact match) on title and an author match.

If we have authors and we have author match, we fall back to `title_similarity_threshold`. That's how this test case passes (I added a period to break title match): https://github.com/Future-House/paper-qa/blob/5ad5894c4a7fca3383ba0688b6bfc99ce7bfc400/tests/test_clients.py#L520-L524

But your point is taken on no authors - will rewrite this. And try to make it a little more readable.

sidnarayanan: Ok - implemented the tighter threshold if no authors, added a test case

whitead (approved):

PR #1089, opened 2025-09-16
Context id updates

mskarlin:
We had some scenarios where our formatted answer could miss context-ids. This implements a more universal matching solution along with tests.

### DIFF
--- a/src/paperqa/types.py
+++ b/src/paperqa/types.py
@@ -388,9 +388,7 @@ class PQASession(BaseModel):
                 for key in get_citation_ids(parenthetical)
                 if id_to_name_map.get(key)
             )
-            # replace the parenthetical with the deduped names
             if deduped_names:
-
                 formatted_without_references = formatted_without_references.replace(
                     parenthetical,
                     f"({', '.join(deduped_names)})",
--- a/src/paperqa/types.py
+++ b/src/paperqa/types.py
@@ -41,6 +41,7 @@ from paperqa.utils import (
     encode_id,
     format_bibtex,
     get_citation_ids,
+    get_parentheses_substrings,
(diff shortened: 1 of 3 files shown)


src/paperqa/types.py
```diff
@@ -377,19 +378,18 @@ def populate_formatted_answers_and_bib_from_raw_answer(
...
+        for parenthetical in get_parenthetical_substrings(formatted_without_references):
             # now we replace eligible parentheticals with the deduped names
-            deduped_names = {
-                id_to_name_map.get(key, "") for key in get_citation_ids(parenthetical)
-            }
-
-            # replace the parenthetical with the deduped names
+            # while we preserve order and deduplicate
+            deduped_names: dict[str, None] = dict.fromkeys(
+                id_to_name_map.get(key)  # type: ignore[misc]
```
jamesbraza:
```suggestion
                id_to_name_map[key]
```

Can't we do this, since we did `if id_to_name_map.get(key)` below

src/paperqa/utils.py
```diff
@@ -154,8 +154,31 @@ def extract_score(text: str) -> int:
...
+    Finds the all nested parenthetical substrings.
+
+    Args:
+        text: The input string to analyze.
+
+    Returns:
+        A list of parenthetical substrings.
+    """
+    substrings = []
+    open_paren_indices = []
```
jamesbraza, line 168: Maybe use `collections.deque` to make it clear this is queue

jamesbraza (approved):
Nice work

src/paperqa/types.py
mskarlin: yea will change

PR #1092, opened 2025-09-19
Removed other from doc details for user

whitead:
These are huge and arbitrary with no guarantees on structure.

### DIFF
--- a/src/paperqa/types.py
+++ b/src/paperqa/types.py
@@ -358,7 +358,7 @@ class PQASession(BaseModel):
                     # Similar to the explanation in `map_fxn_summary`'s internals
                     # on why we drop embeddings, drop embeddings here too because
                     # embeddings aren't displayed to front end users
-                    doc=c.text.doc.model_dump(exclude={"embedding"}),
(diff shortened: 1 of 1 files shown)


mskarlin (approved):

src/paperqa/types.py
```diff
@@ -358,7 +358,7 @@ def filter_content_for_user(self) -> None:
                     # Similar to the explanation in `map_fxn_summary`'s internals
                     # on why we drop embeddings, drop embeddings here too because
                     # embeddings aren't displayed to front end users
-                    doc=c.text.doc.model_dump(exclude={"embedding"}),
+                    doc=c.text.doc.model_dump(exclude={"embedding", "other"}),
```
jamesbraza, line 364: Do you mind adding a code comment saying why we drop `other`

Also, should we do the same thing in `core.map_fxn_summary`?

whitead: added more comments. Nah - we don't want to delete them there, because it's fine to lug them around. Could have purpose. But this method is supposed to provide a smaller object with fixed attributes. The `other` can have all kinds of gunk.

PR #1098, opened 2025-09-22
Adopting `prek` over `pre-commit`, `setup-uv`'s `python-version`

jamesbraza:
This PR upgrades our developer tooling:
- Modernizes our `setup-uv` parameters
    - `UV_PROJECT_ENVIRONMENT` --> `activate-environment` parameter
    - `uv python pin` --> `python-version` parameter
- Having `pre-commit` config use `uv`-managed venv for `mypy`, decoupling from local venv
- Moves from `pre-commit` to its Rust port `prek` for speed (and 🌲)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Clean up paper-qa build # Work around https://github.com/hynek/build-and-inspect-python-package/issues/174
         if: matrix.python-version == '3.11'
         run: rm -r ${{ steps.build-paper-qa.outputs.dist }}
-      - run: uv sync --python-preference=only-managed
+      - run: uv sync
       - run: uv run pylint src packages
       - uses: suzuki-shunsuke/github-action-renovate-config-validator@v1.1.1
   test:
@@ -84,7 +84,7 @@ jobs:
         with:
           enable-cache: true
           python-version: ${{ matrix.python-version }}
-      - run: uv sync --python-preference=only-managed
+      - run: uv sync
       - run: uv run pytest -n auto
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -23,11 +23,10 @@ jobs:
       - uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
-      - run: echo "UV_PROJECT_ENVIRONMENT=$(python -c "import sysconfig; print(sysconfig.get_config_var('prefix'))")" >> $GITHUB_ENV
```
jamesbraza, line 26: This `UV_PROJECT_ENVIRONMENT` tricked `uv` into installing into system Python.

Using `activate-environment` from `setup-uv` eliminates the need for this complexity

.github/workflows/tests.yml
```diff
@@ -23,11 +23,10 @@ jobs:
...
-      - run: uv python pin ${{ matrix.python-version }} # uv requires .python-version to match OS Python: https://github.com/astral-sh/uv/issues/11389
-      - run: uv sync --python-preference only-system
-      - run: git checkout .python-version # For clean git diff given `pre-commit run --show-diff-on-failure`
```
jamesbraza, line 29: We avoid this by just specifying the Python version to `setup-uv` directly

.pre-commit-config.yaml
```diff
@@ -8,7 +8,8 @@ repos:
         exclude: |
           (?x)^(
             src/paperqa/clients/client_data.*|
-            tests/stub_data.*
+            tests/stub_data.*|
+            uv\.lock
```
maykcaldas, line 13: Should we exclude uv.lock here? I've never seen it getting that large

maykcaldas (approved):

.pre-commit-config.yaml
jamesbraza: I faced this in https://github.com/Future-House/paper-qa/pull/1099 too so I moved this change there.

It's just getting big, an `ls -la` is showing its size as `598k` right now

PR #1100, opened 2025-09-24
Upstreaming prompt change

whitead:
Allowing JSON mode to bail out on summarizing, like we allow in non-json mode

### DIFF
--- a/tests/cassettes/test_get_reasoning[deepseek-reasoner].yaml
+++ b/tests/cassettes/test_get_reasoning[deepseek-reasoner].yaml
@@ -45,20 +45,20 @@ interactions:
     response:
       body:
         string: !!binary |
-          H4sIAAAAAAAAAwAAAP//jJLfa9swEMff/VeIe06G4yZxyNvYoHRQBoPRsjrYsnRJ1MqSkM7pj5D/
-          fchO4nTrYC96uM99T/e9u33CGCgJSwZiy0k0To+/fJu336/udj/vF/OX1+tfdfb17fbp+vamztM3
-          GEWFrR9R0En1SdjGaSRlTY+FR04Yq07y2WKep4ts2oHGStRRtnE0ntpxlmbT8WQyztKjcGuVwABL
-          9pAwxti+e2OLRuILLFk6OkUaDIFvEJbnJMbAWx0jwENQgbghGA1QWENouq6rqnoM1hRmXxjGCiBF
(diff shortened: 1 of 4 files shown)


src/paperqa/prompts.py
```diff
@@ -113,10 +113,12 @@
     " Respond with the following JSON format:"
     '\n\n{{\n  "summary": "...",\n  "relevance_score": "..."\n  "used_images"\n}}'
     "\n\nwhere `summary` is relevant information from the text - {summary_length} words."
-    " `relevance_score` is an integer 1-10 for the relevance of `summary` to the question."
+    " `relevance_score` is an integer 0-10 for the relevance of `summary` to the question."
```
jamesbraza, line 117: Should we add the `0-10` to the `summary_prompt` above as well?

jamesbraza (approved):
Nice

src/paperqa/prompts.py
jamesbraza, line 117: Should we add this `0-10` thing to the JSON snippet as well

```python
    '  "relevance_score": 0-10\n'  # putting numbers here, instead of "...", to get them as integer type not string
```

src/paperqa/prompts.py
```diff
@@ -113,10 +113,12 @@
...
     " `used_images` is a boolean flag indicating"
     " if any images present in a multimodal message were used,"
     " and if no images were present it should be false."
+    "\n\nThe excerpt may or may not contain relevant information. If not, leave `summary` empty, "
```
jamesbraza: Can you update `test_paperqa.py`'s `test_json_evidence` too

src/paperqa/prompts.py
```diff
@@ -113,10 +113,12 @@
     " Respond with the following JSON format:"
     '\n\n{{\n  "summary": "...",\n  "relevance_score": "..."\n  "used_images"\n}}'
```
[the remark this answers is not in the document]
jamesbraza: Fixed the missing comma in https://github.com/Future-House/paper-qa/pull/1107, I will fix `used_images` in another PR

src/paperqa/prompts.py
jamesbraza: Since we use "not applicable" phrasing here, we don't need 0-10

I documented this in https://github.com/Future-House/paper-qa/pull/1107

jamesbraza: Done in https://github.com/Future-House/paper-qa/pull/1107

src/paperqa/prompts.py
jamesbraza: Done in https://github.com/Future-House/paper-qa/pull/1107

PR #1103, opened 2025-09-24
Gather evidence fails gracefully on timeout

sidnarayanan:
We basically never want to see this as a response to a gather evidence tool call:
```
Encountered exception during tool call: litellm.Timeout: APITimeoutError - Request timed out. Error_str: Request timed out.
```

We were previously bubbling up timeout errors all the way up to the tool call. Even one context failure would prevent all other successes from making it through. This PR catches and returns None if we timeout.

We may want to extend this to e.g. the BadRequestErrors, but sticking to Timeouts for now.

### DIFF
--- a/src/paperqa/core.py
+++ b/src/paperqa/core.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 from collections.abc import Callable, Sequence
-from typing import Any
+from typing import Any, ClassVar
 
 import litellm
 from aviary.core import Message
@@ -124,12 +124,40 @@ def llm_parse_json(text: str) -> dict[str, JsonValue]:
     return data
 
 
-class LLMBadContextJSONError(ValueError):
-    """Retryable exception for when the LLM gives back bad JSON."""
+class LLMContextError(ValueError):
+    retryable: ClassVar[bool]
+    error_log: ClassVar[str]
 
     def __init__(self, message: str, llm_results: list[LLMResult]) -> None:
         super().__init__(message)
         self.llm_results = llm_results  # House so we can cost track across retries
+
+
+class LLMBadContextJSONError(LLMContextError):
+    """Retryable exception for when the LLM gives back bad JSON."""
+
+    retryable = True
+    error_log = (
+        "Abandoning this context."
+        " Your model may not be capable of supporting JSON output"
+        " or our parsing technique could use some work. Try"
+        " a different model or specify `Settings(prompts={'use_json': False})`."
+        " Or, feel free to just ignore this message, as many contexts are"
+        " concurrently made and we're not attached to any one given context."
+    )
+
+
+class LLMContextTimeoutError(LLMContextError):
+    """Non-retryable exception for when the LLM call times out."""
+
+    retryable = False
+    error_log = (
+        "Timeout when creating a context, abandoning it."
+        " If you see this error frequently, consider increasing the timeout in"
+        " Settings(summary_llm_config=...). Or, feel free to just ignore this message,"
+        " as many contexts are concurrently made and we're not attached to any one"
+        " given context."
+    )
 
 
 async def _map_fxn_summary(  # noqa: PLR0912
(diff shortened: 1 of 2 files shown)


src/paperqa/core.py
```diff
@@ -204,41 +232,48 @@ async def _map_fxn_summary(  # noqa: PLR0912
...
+                    ],
+                    callbacks=callbacks,
+                    name="evidence:" + text.name,
+                )
+                used_text_only_fallback = True
+        except litellm.Timeout as exc:
+            raise LLMContextTimeoutError(
+                f"LLM call to create a context timed out on text named {text.name!r}: {exc}",
+                llm_results=llm_results,
+            ) from exc
```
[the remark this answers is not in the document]
sidnarayanan: Yeah, this is a good point, except we make two `call_single` calls inside this outer block. So it's either one outer block or two inner try/excepts 🤷‍♂️ . This felt better.

src/paperqa/core.py
```diff
@@ -124,14 +124,42 @@ def fraction_replacer(match: re.Match) -> str:
...
         super().__init__(message)
         self.llm_results = llm_results  # House so we can cost track across retries
 
 
+class LLMBadContextJSONError(LLMContextError):
+    """Retryable exception for when the LLM gives back bad JSON."""
+
+    retryable = True
+    error_log = (
+        "Abandoning this context."
```
jamesbraza:
```suggestion
        "Abandoning this context creation."
```

src/paperqa/core.py
```diff
@@ -124,14 +124,42 @@ def fraction_replacer(match: re.Match) -> str:
     return data
 
 
-class LLMBadContextJSONError(ValueError):
-    """Retryable exception for when the LLM gives back bad JSON."""
+class LLMContextError(ValueError):
+    retryable: ClassVar[bool]
+    error_log: ClassVar[str]
```
jamesbraza:
```suggestion
    error_log: ClassVar[str]  # NOTE: this is not used in `str(exc)`
```

Or something similar (e.g. a rename), just clarifying this is not what the LLM sees upon retries, but what a human sees in logs

src/paperqa/core.py
```diff
@@ -204,41 +232,48 @@ async def _map_fxn_summary(  # noqa: PLR0912
...
+                    messages=[
+                        Message(role="system", content=system_prompt),
+                        Message(content=message_prompt),
+                        *append_msgs,
+                    ],
+                    callbacks=callbacks,
+                    name="evidence:" + text.name,
+                )
+                used_text_only_fallback = True
+        except litellm.Timeout as exc:
```
jamesbraza, line 271: The fact `litellm` didn't call this a `TimeoutError` annoys me 😆

src/paperqa/core.py
```diff
@@ -204,41 +232,48 @@ async def _map_fxn_summary(  # noqa: PLR0912
...
+            raise LLMContextTimeoutError(
+                f"LLM call to create a context timed out on text named {text.name!r}: {exc}",
```
jamesbraza:
```suggestion
                f"LLM call to create a context timed out on text named {text.name!r}."
```

I think because (1) we do `from exc` and (2) we use `logger.exception` below, means we don't need to duplicate it in the string

tests/test_paperqa.py
```diff
@@ -2616,3 +2622,49 @@ def test_pqa_context_id_parsing(raw_text: str, cleaned_text: str) -> None:
     )
     session.populate_formatted_answers_and_bib_from_raw_answer()
     assert session.answer == cleaned_text
+
+
+@pytest.mark.skipif(os.environ.get("ANTHROPIC_API_KEY") is None, reason="No API key")
```
jamesbraza:
```suggestion
```

Let's just always run this test

sidnarayanan: You sure? I was copying what we did for `test_anthropic` above

jamesbraza: From my experience having disabled aspects of CI can lead to hidden failures

I am okay if you'd like to keep it though

PR #1106, opened 2025-09-25
`gather_evidence` tool response filtering irrelevant `Context`, with updated threshold of `0`

jamesbraza:
We were seeing `gather_evidence` responses like this:

```none
Added 22 pieces of evidence, 0 of which were relevant. Best evidence(s):

1. 

2. 

3. 

4. 

5. 

Status: Paper Count=2 | Relevant Papers=0 | Current Evidence=0 | Current Cost=$0.0086"
```

There's more to the story here, but this PR at least:

- Documents when we'd have an empty `Context.context`
- Updates the relevant evidence threshold from `5` to `0`
- Filters the `gather_evidence` tool response for zero-scored `Context`


src/paperqa/core.py
```diff
@@ -276,14 +276,14 @@ async def _map_fxn_summary(  # noqa: PLR0912
 
         llm_results.append(llm_result)
         context = llm_result.text or ""
-        try:
-            result_data = parser(context) if parser else {}
```
jamesbraza, line 280: This little refactor removes the extra `if parser else {}`, making the code slightly less stateful

whitead (approved):
Nice work getting all these cleaned-up

PR #1107, opened 2025-09-25
Minor prompt adjustments and docs

jamesbraza:
Resolving my own PR comments

### DIFF
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -533,11 +533,11 @@ async def test_json_evidence(docs_fixture: Docs) -> None:
         " based on the excerpt. The excerpt may be irrelevant."
         " Do not directly answer the question - only summarize relevant information."
         " Respond with the following JSON format:"
-        '\n\n {{\n"summary": "...",\n"author_name": "...",\n"relevance_score": "..."}}'
-        "\n\n where `summary` is relevant information from text - about 100 words words,"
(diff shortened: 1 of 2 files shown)


tests/test_paperqa.py
```diff
@@ -529,14 +529,15 @@ async def test_json_evidence(docs_fixture: Docs) -> None:
...
-        " question (integer out of 10)."
+        "Provide a summary of the excerpt that could help answer the question"
+        " based on the excerpt. The excerpt may be irrelevant."
+        " Do not directly answer the question - only summarize relevant information."
+        " Respond with the following JSON format:"
+        '\n\n{{\n  "summary": "...",\n  "author_name": "...",\n  "relevance_score": 1-10\n}}'
+        "\n\nwhere `summary` is relevant information from text - about 100 words words,"
+        " `author_name` specifies the author,"
+        " and `relevance_score` is an integer 1-10 for the relevance of `summary`"
+        " to the question."
```
whitead: This conflicts with prior - should be 0-10

whitead (approved):

tests/test_paperqa.py
jamesbraza: This prompt clobbers the prior and it doesn't mention the `0` score case. What I can do is just propagate the `0` score case here

jamesbraza: Alright it's done, agreed it's nice just to standardize

PR #1141, opened 2025-10-16
Making search tools concurrent

sidnarayanan:
It is safe for search tools to run concurrently with each other, so this PR enables that.

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -289,8 +289,8 @@ async def test_get_directory_index_w_manifest(agent_test_settings: Settings) -> 
     assert set((await absolute_index.index_files).keys()) == {
         str(abs_paper_dir / f) for f in EXPECTED_STUB_DATA_FILES
     }, (
-        "Incorrect index files, should be absolute to deny sharing indexes across "
-        "machines"
+        "Incorrect index files, should be absolute to deny sharing indexes across"
+        " machines"
     )
     for index in (relative_index, absolute_index):
         assert index.fields == [


src/paperqa/agents/env.py
```diff
@@ -69,7 +69,8 @@ def settings_to_tools(  # noqa: PLR0912
             tool = Tool.from_function(
                 PaperSearch(
                     settings=settings, embedding_model=embedding_model
-                ).paper_search
+                ).paper_search,
+                concurrency_safe=tool_type.CONCURRENCY_SAFE,
```
jamesbraza: Can you add comments to the code explaining why we've decided to take on the risk?

The `Docs` functions underlying search and gathering evidence in general are not concurrency safe

tests/test_agents.py
```diff
@@ -1073,7 +1071,6 @@ async def test_continuation(self) -> None:
 
 @pytest.mark.asyncio
 async def test_index_build_concurrency(agent_test_settings: Settings) -> None:
-
```
jamesbraza, line 1076: [pr nit] why the random whitespace changes in random tests?

Mainly unrelated whitespace changes degrade the `git blame`

sidnarayanan: Fixed, was ruff format vs black

src/paperqa/agents/env.py
sidnarayanan: After looking at it in person, we confirmed `aadd_text` is concurrency-safe, and that's the only Docs manipulation by the search tools. Documented.

tests/test_agents.py
jamesbraza: After reading up on it, yeah it's https://github.com/astral-sh/ruff/issues/9745. Dang too bad

jamesbraza (approved):
Nice work, LGTM!

PR #1153, opened 2025-10-27
Deduplicating media on `Context` creation

jamesbraza:
I realized we were passing the same images for `Context` creation. This PR:

1. Adds an `info_hash` to `ParsedMedia` to allow for loosened deduplication without affecting stored info
    - Now: can store high-resolution bbox, but have low-resolution bbox with deduplication
2. Uses ordered set-based deduplication for the `Context` creation prompt

This PR also makes a custom PDF for testing deduplication capabilities within multimodal PaperQA.

### DIFF
--- a/src/paperqa/prompts.py
+++ b/src/paperqa/prompts.py
@@ -16,7 +16,7 @@ summary_prompt = (
 text_with_tables_prompt_template = (
     "{text}\n\n------------\n\nMarkdown tables from {citation}."
     " If the markdown is poorly formatted, defer to the images"
-    "\n\n{tables}"
+    "\n\n------------\n\n{tables}"
 )
 
 summary_json_prompt = (
--- a/src/paperqa/core.py
+++ b/src/paperqa/core.py
@@ -240,7 +240,7 @@ async def _map_fxn_summary(  # noqa: PLR0912
                 text_with_tables_prompt_template.format(
                     text=cleaned_text,
                     citation=citation,
-                    tables="\n\n".join(media_text),
+                    tables="\n\n----\n\n".join(media_text),
                 )
                 if media_text
                 else cleaned_text


packages/paper-qa-docling/src/paperqa_docling/reader.py
```diff
@@ -169,6 +170,18 @@ def parse_pdf_to_pages(  # noqa: PLR0912
...
+                        k: (
+                            v
+                            if k != "bbox"
+                            # Deduplicate bboxes based on whole pixels
+                            else tuple(round(x) for x in cast(tuple, v))
+                        )
+                        for k, v in media_metadata.items()
+                    },
+                    sort_keys=True,
+                )
```
sidnarayanan: nit: should this be called a hash if it's a json serialization? Why not hash the string or call it `info_serialized`?

src/paperqa/prompts.py
```diff
@@ -16,7 +16,7 @@
 text_with_tables_prompt_template = (
     "{text}\n\n------------\n\nMarkdown tables from {citation}."
```
sidnarayanan, line 17: Should we use Markdown hrule `---` everywhere instead a mix of variable-lengths? I've been trying to standardize on 3.

jamesbraza: That is a great idea and agreed.

I will just do this revert this change for this PR, it's unrelated to media deduplication and I was cargo culting it in 😈

packages/paper-qa-docling/src/paperqa_docling/reader.py
jamesbraza: Yeah it's a good point. Actually what I care is that `info_hash` is hashable. It can be an `int` output from `hash`, or it can be a string.

To disambiguate, I just renamed this to `info_hashable`

src/paperqa/prompts.py
jamesbraza: Alright moved this to https://github.com/Future-House/paper-qa/pull/1156, with more changes. Resolving this one

PR #1160, opened 2025-10-29
Exposed `reader_config` setting to set DPI and simplify settings

jamesbraza:
This PR allows users to:
- Configure settings like DPI
- Simplifies overall setting count in v6


tests/test_paperqa.py
```diff
@@ -3047,3 +3047,58 @@ async def test_timeout_resilience() -> None:
...
+    """Test that deprecated settings trigger warnings and are migrated to reader_config."""
+    with pytest.warns(DeprecationWarning, match="chunk_size.*deprecated"):
+        settings1 = Settings(parsing=ParsingSettings(chunk_size=2000))
+    assert settings1.parsing.reader_config["chunk_chars"] == 2000
+    with pytest.warns(DeprecationWarning, match="overlap.*deprecated"):
+        settings2 = Settings(parsing=ParsingSettings(overlap=50))
+    assert settings2.parsing.reader_config["overlap"] == 50
+    with pytest.warns(DeprecationWarning, match="pdfs_use_block_parsing.*deprecated"):
+        settings3 = Settings(parsing=ParsingSettings(pdfs_use_block_parsing=True))
+    assert settings3.parsing.reader_config["use_block_parsing"]
```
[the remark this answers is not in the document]
jamesbraza: A truthy check is fine for now

src/paperqa/docs.py
```diff
@@ -278,17 +278,15 @@ async def aadd(  # noqa: PLR0912
...
                 page_size_limit=parse_config.page_size_limit,
-                use_block_parsing=parse_config.pdfs_use_block_parsing,
                 parse_images=False,  # Peeking is text only
                 # We only use the first chunk, so let's peek just enough pages for that.
                 # Usually pages 1 - 2 give that,
                 # but in the event page 2 is blank (true for some PDFs),
                 # we read pages 1 - 3 to be safe
                 page_range=(1, 3),
                 parse_pdf=parse_config.parse_pdf,
+                **parse_config.reader_config,
```
maykcaldas, line 289: Do we want this to act like a kwargs to keep the freedom or should we make a pydantic validation for `reader_config`?

maykcaldas: That should be ok. It's just that I don't know the valid keywords for the reader

jamesbraza: Yeah I follow what you're saying. I am having it as general kwargs because each reader (PyPDF, PyMuPDF, Docling) has different parameters/capabilities (mirroring the different libraries)

src/paperqa/readers.py
```diff
@@ -345,8 +345,8 @@ async def read_doc(  # noqa: PLR0912
     doc: Doc,
     parsed_text_only: bool = False,
     include_metadata: bool = False,
-    chunk_chars: int = 3000,
```
jamesbraza, line 348: These were lagging our defaults in `Settings`, so this re-syncs them

PR #1162, opened 2025-10-31
'Full page screenshot' enrichment prompt template

jamesbraza:
This PR adds a prompt template for full-page screenshots


src/paperqa/prompts.py
```diff
@@ -198,3 +198,21 @@
     "\n\n{context_text}Describe the media,"  # Allow for empty context_text
     " or if uncertain on a description please state why:"
 )
+full_page_enrichment_prompt_template = (
```
maykcaldas, line 201: Is this being used somewhere? Or am I missing something?

jamesbraza: Yeah it's dead code in this PR, but when I make a benchmarks PR in the future it will be used. The main point of this PR was having variable names be more scalable

PR #1168, opened 2025-11-01
Make OpenAlex work

whitead:
It turns out openalex never worked if you query on DOI. OpenAlex refers to DOIs always with the URL prefix (e.g., https://doi.org/10xxx/xxxxx). So when it returned articles, they had DOIs that didn't match the requested DOIs and we rejected them as failed. This fixes and adds a low-effort unit test that just checks if openalex is working.

---

> [!NOTE]
> Normalize OpenAlex DOIs by removing the https://doi.org/ prefix before comparison and add a VCR test to verify DOI queries work via OpenAlexProvider.
> 
> - **Clients/OpenAlex**:
>   - Normalize `doi` by removing `https://doi.org/` prefix in API results before matching against requested DOI.
>   - Handle title-search path by selecting first result and validating non-empty results.
> - **Tests**:
>   - Add `OpenAlexProvider` to tests and new async test `test_does_openalex_work` querying by DOI with `fields=["open_access"]`.
>   - Record VCR cassette for the DOI request to `OpenAlex`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0da01b24c89e6b80f41927176df38fa3580a3bf7. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/cassettes/test_does_openalex_work[10.1021-acs.jctc.5b00178-True].yaml
+++ b/tests/cassettes/test_does_openalex_work[10.1021-acs.jctc.5b00178-True].yaml
@@ -59,44 +59,44 @@ interactions:
           V9EJRNSPYz8M+hD5IY1DSoYhIgFsjU9lyfZsKzlGuGtUkJl9Cqm8DyEw4tPmKHxA7IJ02eV3cNQF
           tYUAdAJO8OBccbGfzfcAYjgsDEjcWwbRMKbBwPCFoDYjZ/LSiXTkxcT3IZb3YAqwSaokGFZjdB0k
           tF9L+c63aheuKICPH9rnABj6PhBwIzwwZoVRV/Hz3pxr5F2+uQfNVNpwqp02R4ztJXW1c265GWsK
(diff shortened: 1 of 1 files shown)


tests/test_clients.py
```diff
@@ -793,3 +793,24 @@ async def test_tricky_journal_quality_results(doi: str, score: int) -> None:
...
+@pytest.mark.asyncio
+async def test_does_openalex_work(doi: str):
+    async with httpx_aiohttp.HttpxAiohttpClient() as http_client:
+        openalex_client = DocMetadataClient(
+            http_client,
+            metadata_clients=[OpenAlexProvider],
+        )
+        openalex_details = await openalex_client.query(
+            doi=doi,
+            fields=["open_access"],
```
jamesbraza, line 815: Why do we include `fields=["open_access"]` in this test? Can we either:
- Remove it if unnecessary
- Document and/or assert upon the results

jamesbraza (approved):
Nice catch

tests/test_clients.py
whitead: added docstring and real test

jamesbraza: Nice and thanks. Already approved the PR

Issue #1176, opened 2025-11-04
PDF parsing error (tried with `uv add git+https://github.com/Future-House/paper-qa.git`)


takeruhukushima:
I'm trying to do with .pptx ,.xlsx,.docx parsing.
Then encountered PDF parsing error.
another version (not installed from github) PDF pars is Successful.
Would you mind helping me figure out what's going on?

```Python
import asyncio
import os
from dotenv import load_dotenv
from paperqa import Settings, ask
from paperqa.settings import AgentSettings
from datetime import datetime
import re

load_dotenv()

def clean_answer_text(text):
    """回答テキストから余計な情報を除去する関数"""
    if not text:
        return ""
    
    # 文字列に変換
    text = str(text)
    
    # "Question: ..." の行を除去
    text = re.sub(r'^Question:.*\n', '', text)
    
    # "References" 以降の部分を除去
    text = re.sub(r'\n\nReferences.*$', '', text, flags=re.DOTALL)
    
    # 複数の改行を整理
    text = re.sub(r'\n\s*\n', '\n\n', text)
    text = re.sub(r'^\s+|\s+$', '', text)
    
    return text.strip()

async def main():
    gemini_api_key = os.getenv("GEMINI_API_KEY")
    if not gemini_api_key:
        gemini_api_key = os.getenv("GOOGLE_API_KEY")

    if not gemini_api_key:
        print("エラー: 環境変数 'GEMINI_API_KEY' または 'GOOGLE_API_KEY' が設定されていません。")
        print(".envファイルを作成し、APIキーを記述してください。")
        return

    question_to_ask_initially = "What is PaperQA2?" # 初期質問 (ループで上書き)
    papers_directory = "./input"
    output_directory = "./out"

    gemini_llm_name = "gemini/gemini-2.5-flash"
    gemini_embedding_name = "gemini/text-embedding-004"

    os.makedirs(output_directory, exist_ok=True)

    print("PaperQA (Gemini版) チャットへようこそ！")
    print(f"論文は '{papers_directory}' ディレクトリから読み込みます。")
    print(f"回答は '{output_directory}' ディレクトリにMarkdownファイルとして保存されます。")
    print("質問を入力してください (終了するには 'exit' または 'quit' と入力)。")
    print("-" * 30)

    base_settings = Settings(
        llm=gemini_llm_name,
        summary_llm=gemini_llm_name,
        embedding=gemini_embedding_name,
[...]

PR #1200, opened 2025-11-13
Updating readers for full-page parsing

jamesbraza:
This PR upgrades our readers a bit:
- As shown in https://github.com/Future-House/paper-qa/issues/1176, the current default settings (just PyPDF + `multimodal=True`) currently force one to hit a `NotImplementedError`, so the default settings are improved

   ```none
   NotImplementedError: Media support without just a screenshot of the full page is not yet implemented.  
   ```

- Adds tests for `FileNotFoundError` to each reader
- Renames PyMuPDF's `image_dpi` arg to `dpi`, since (1) this parameter impacts tables too and (2) it will now match other readers' DPI parameter name (e.g. Docling)


jamesbraza:
Let's hold off on this for now, I am not sure full-page parsing is the right default either

I upstreamed most of this to https://github.com/Future-House/paper-qa/pull/1205

PR #1203, opened 2025-11-14
Math formula support

jamesbraza:
I wanted to see if/how https://build.nvidia.com/nvidia/nemotron-parse respects LaTeX. Also in general, it's nice to have formula support in our prompts and ecosystem, they show up as images:

screenshot of duplicate media template PDF


src/paperqa/prompts.py
```diff
@@ -129,7 +129,8 @@
     " if any images present in a multimodal message were used,"
     " and if no images were present it should be false."
     "\n\nThe excerpt may or may not contain relevant information."
-    " If not, leave `summary` empty, and make `relevance_score` be 0."
+    " If not, leave `summary` empty, make `relevance_score` be 0,"
+    " and set `used_images` as false."
```
whitead: why?

jamesbraza: Actually you're totally right, this prompt change is a bad idea, as images can be used to determine irrelevance

jamesbraza: Alright reverted it with explanatory comment. Thanks

packages/paper-qa-docling/tests/test_paperqa_docling.py
```diff
@@ -182,15 +182,15 @@ def test_media_deduplication() -> None:
     all_media = [m for _, media in parsed_text.content.values() for m in media]  # type: ignore[misc]
 
     all_images = [m for m in all_media if m.info.get("type") == "picture"]
-    assert len(all_images) == 5, "Expected each image to be read"
+    assert len(all_images) == 11, "Expected each image and formula to be read"
```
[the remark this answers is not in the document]
jamesbraza: Sure I just changed the assertion and added a comment to document why 11 clearly. Thanks

jamesbraza:
> looks awesome!

Yeah I was looking at the enrichments, trying to see if it would recognize the formula. Here's what GPT-4o says:

---

The media is a mathematical formula written in LaTeX format. It appears to represent a recursive or iterative expression involving variables \(x\), \(n\), and \(a\). The formula is as follows:

\[
x + n + a = \sqrt{ax + (n + a)^2 + x\sqrt{a(x + n) + (n + a)^2 + (x + n)\sqrt{\cdots}}}
\]

### Key Elements:
1. **Variables**:
   - \(x\), \(n\), and \(a\) are the primary variables in the equation.
   - These variables are combined in various ways, including addition, multiplication, and squaring.

2. **Structure**:
   - The equation is recursive, with nested square root terms that repeat indefinitely.
   - Each nested term includes a combination of the variables \(x\), \(n\), and \(a\), along with their squared or multiplied forms.

3. **Mathematical Operations**:
   - Addition, multiplication, squaring, and square root operations are used extensively.
   - The recursive nature suggests a potentially infinite series or iterative computation.

4. **Purpose**:
   - The formula may represent a mathematical model, a series expansion, or a solution to a specific problem involving recursion.

### Observations:
- The formula is complex and appears to involve a self-referential or iterative process.
- The ellipsis (\(\cdots\)) at the end indicates that the recursion continues indefinitely.

This formula is likely part of a mathematical or scientific discussion, possibly related to series, recursion, or iterative methods. The surrounding text does not provide additional context for the formula, so its specific application or meaning is unclear.

PR #1204, opened 2025-11-14
Retrying 503's from OpenAlex

jamesbraza:
It seems these should be retried too


whitead (approved):
can you just proactively retry all 5xxs?

jamesbraza:
> can you just proactively retry all 5xxs?

Meh there's 5xx codes (e.g. 501 Not Implemented, 505 HTTP Version Not Supported) that I know OpenAlex doesn't throw in normal operation, so I don't want to retry these.

Also, Mike pointed out in Slack that the 503's probably arose from nearly DDOS'ing OpenAlex (load balancer had no more instances to serve a request), given our high concurrency of our metadata audit script.

With retrying of 503's it seems OpenAlex can recover, but in general I realized this is probably not a retryable error

PR #1205, opened 2025-11-14
Minor reader bugfixes and fortified regression testing

jamesbraza:
This PR improves consistency across readers:
- Tests equation parsing of Docling and PyMuPDF
- Renames PyMuPDF's `image_dpi` arg to `dpi`, since (1) this parameter impacts tables too and (2) it will now match other readers' DPI parameter name (e.g. Docling)
- Updates PyMuPDF reader's default DPI to unspecified, to align with PyMuPDF's default behavior
- Includes `full_page` in autogenerated index name hash (as full page parsing should not have a collision)
- Adds many smaller assertions: image coordinates and shapes, `FileNotFoundError`, expected screenshot type

### DIFF
--- a/packages/paper-qa-pymupdf/src/paperqa_pymupdf/reader.py
+++ b/packages/paper-qa-pymupdf/src/paperqa_pymupdf/reader.py
@@ -70,7 +70,9 @@ def parse_pdf_to_pages(  # noqa: PLR0912
             Can be a single value to apply to both X and Y directions,
             or a two-tuple to specify X and Y directions separately.
             The default was chosen to perform well on image extraction from LitQA2 PDFs.
-        dpi: Optional DPI (dots per inch) for image resolution.
+        dpi: Optional DPI (dots per inch) for image resolution,
+            if left unspecified PyMuPDF's default resolution from
+            pymupdf.Page.get_pixmap will be applied.
         page_range: Optional start_page or two-tuple of inclusive (start_page, end_page)
             to parse only specific pages, where pages are one-indexed.
             Leaving as the default of None will parse all pages.
--- a/packages/paper-qa-docling/src/paperqa_docling/reader.py
+++ b/packages/paper-qa-docling/src/paperqa_docling/reader.py
@@ -54,8 +54,8 @@ def parse_pdf_to_pages(  # noqa: PLR0912
         parse_media: Flag to also parse media (e.g. images, tables).
         pipeline_cls: Optional custom pipeline class for document conversion.
             Default is Docling's standard PDF pipeline.
-        dpi: Optional DPI (dots per inch) for image resolution.
-            Default PDF resolution is 72 DPI, so dpi of 144 would render at 2x scale.
+        dpi: Optional DPI (dots per inch) for image resolution,
+            if left unspecified Docling's default 1.0 scale will be employed.
         custom_pipeline_options: Optional keyword arguments to use to construct the
             PDF pipeline's options.
         page_range: Optional start_page or two-tuple of inclusive (start_page, end_page)


packages/paper-qa-pymupdf/src/paperqa_pymupdf/reader.py
```diff
@@ -53,7 +53,7 @@ def parse_pdf_to_pages(  # noqa: PLR0912
     parse_media: bool = True,
     full_page: bool = False,
     image_cluster_tolerance: float | tuple[float, float] = 25,
-    image_dpi: float | None = 150,
+    dpi: float | None = None,
```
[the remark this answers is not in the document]
jamesbraza: No I decided to have it default to `None` to align with the reader library itself

packages/paper-qa-pymupdf/src/paperqa_pymupdf/reader.py
```diff
@@ -70,7 +70,7 @@ def parse_pdf_to_pages(  # noqa: PLR0912
             Can be a single value to apply to both X and Y directions,
             or a two-tuple to specify X and Y directions separately.
             The default was chosen to perform well on image extraction from LitQA2 PDFs.
-        image_dpi: Dots per inch for images captured from the PDF.
+        dpi: Optional DPI (dots per inch) for image resolution.
```
nadolskit: nit: do you mind documenting the consequences of what happens when `None` is used? assuming the providing function `get_pixmap` just provides an internal value

packages/paper-qa-pymupdf/tests/test_paperqa_pymupdf.py
```diff
@@ -258,3 +274,16 @@ def custom_to_markdown(self, clean=False, fill_empty=True) -> str:
...
+
+def test_equation_parsing() -> None:
+    parsed_text = parse_pdf_to_pages(STUB_DATA_DIR / "duplicate_media.pdf")
+    assert isinstance(parsed_text.content, dict)
+    assert isinstance(parsed_text.content["1"], tuple)
+    p1_text, p1_media = parsed_text.content["1"]
+    # SEE: https://regex101.com/r/pyOHLq/1
+    assert re.search(
+        r"[_*]*E[_*]* ?= ?[_*]*mc[_*]*(?:<sup>)?[ ^]?[2²] ?(?:<\/sup>)?", p1_text
+    ), "Expected inline equation in page 1 text"
```
sidnarayanan, line 287: This doesn't cover LaTeX macros, see the last case here: https://regex101.com/r/1p2wpa/1

Maybe try this one? https://stackoverflow.com/questions/14182879/regex-to-match-latex-equations 
https://regex101.com/r/wP2aV6/25

jamesbraza: So for context, this assertion is a superset of what our the readers (PyPDF, PyMuPDF, Docling, Nemotron Parse) will return per this inline PDF content: 

screenshot of duplicate media PDF

That `\sqrt{m^2 c^4 + p^2 c^2}` case or the LaTeX macro for block formula case won't come up. And if they did come up, it would be a PDF misread of the inline LaTeX.

In general this assertion is a 'sanity check' that we're reading the formula. If new cases arrive, we can adjust then as needed.

Let me know what you're thinking, but I don't think we need to cover LaTeX macros

packages/paper-qa-pymupdf/src/paperqa_pymupdf/reader.py
jamesbraza: Yeah just did so, thank for the request

packages/paper-qa-pymupdf/tests/test_paperqa_pymupdf.py
sidnarayanan: Ah ok, this is just for rendered PDFs, got it. Sounds good!

PR #1209, opened 2025-11-17
Pulling in latest `fhlmi` and `prek`

jamesbraza:
Getting CI to pass again after https://github.com/Future-House/ldp/pull/369 and https://github.com/j178/prek/issues/1104

Also working around https://github.com/mwouts/jupytext/issues/1467 better after more updates


jabra (approved):
LGTM

PR #1213, opened 2025-11-20
Removed `image_enrichment_pages` extra in `read_doc` overloads

jamesbraza:
This was missed during a refactor I did halfway through https://github.com/Future-House/paper-qa/pull/1143


jamesbraza:
Made https://github.com/python/mypy/issues/20278 so hopefully `mypy` can catch this mistake in the future

PR #1215, opened 2025-11-20
Define Session.extra_info

sidnarayanan:
I wanted the ability to add extra fields to PQASession

---

> [!NOTE]
> Adds a user-defined `extra_info` field to `PQASession` with serialization support and tests.
> 
> - **Core/Types**:
>   - Add `PQASession.extra_info: dict[str, JsonValue]` for arbitrary user metadata.
> - **Tests**:
>   - Add `test_pqa_session_serialization` to validate (de)serialization with and without `extra_info`.
> - **Misc**:
>   - Minor string fix in a test assertion message.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1a6ddcf1660a33028921629d31fa3875baedac57. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/paperqa/types.py
+++ b/src/paperqa/types.py
@@ -357,6 +357,12 @@ class PQASession(BaseModel):
         ),
     )
 
+    extra_info: dict[str, JsonValue] = Field(
+        default_factory=dict,
+        description="A field for any extra user-defined information. "
+        "Can be used for things like metadata.",
+    )
+
     def __str__(self) -> str:
         """Return the answer as a string."""
         return self.formatted_answer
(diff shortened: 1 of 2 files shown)


src/paperqa/types.py
```diff
@@ -357,6 +357,12 @@ class PQASession(BaseModel):
         ),
     )
 
+    extra_info: dict[str, JsonValue] = Field(
```
jamesbraza, line 360: Should we just do `extra="allow"` for Pydantic?

We already support extension via subclassing since we have `extra="ignore"`

sidnarayanan: Yeah on further thought I think I will just make a subclass for what I want to do 👍

Issue #1273, opened 2026-01-22
enrichment image format not accepted


satra:
for several, but not all pdfs that i am indexing, i get the following error

> Skipping enrichment for media index 0 on page 1 with metadata {'type': 'picture', 'width': 517, 'height': 108, 'info_hashable': '{"height": 108, "type": "picture", "width": 517}', 'page_num': 1} because it was rejected by the LLM provider for 'gpt-4o-2024-11-20'. Full error message: litellm.BadRequestError: OpenAIException - You uploaded an unsupported image. Please make sure your image has of one the following formats: ['png', 'jpeg', 'gif', 'webp'].. Received Model Group=gpt-4o-2024-11-20

is there a way to address this issue?

jamesbraza:
Huh I haven't seen this. Are you able to share a PDF DOI or title, or the PDF itself that leads to this error?

Any chance you can share a minimal reproducer?

satra:
@jamesbraza - this is an example pdf with that warning: https://doi.org/10.1093/sleep/zsae048 (the pdf should be downloadable at the link). i'm also getting litellm timeouts when processing this, but that may be an unrelated issue.

jamesbraza:
Hi @satra, with `paper-qa[pypdf]==2026.01.05`, the below code works for me, including enrichment:

```python
from paperqa import Docs, Settings

docs = Docs()
settings = Settings(
    parsing={"parse_pdf": "paperqa_pypdf.parse_pdf_to_pages", "multimodal": True}
)
docname = await docs.aadd(
    "zsae048.pdf",
    title=(
        "Data-driven, generalizable prediction of adolescent sleep disturbances"
        " in the multisite Adolescent Brain Cognitive Development Study"
    ),
    doi="10.1093/sleep/zsae048",
    settings=settings,
)
(details,) = (d for d in docs.docs.values() if d.docname == docname)
texts = [t for t in docs.texts if t.doc == details]
media = [m for t in texts for m in t.media]
```

Can you give me a Python script that repros your issue?

satra:
@jamesbraza - here is the code: https://github.com/sensein/paperqa-test/blob/main/index_pdfs.py

i'm doing it on a locally downloaded pdf (i have both openai and semantic scholar keys set).

<details> <summary> script output </summary>

```
❯ python ../../paperqa-test/index_pdfs.py --paper-dir pubs
[...]

PR #1298, opened 2026-02-20
Fixed PyPDF reader's failure to handle non-PNG data without `pdfplumber`

jamesbraza:
Closes https://github.com/Future-House/paper-qa/issues/1273

### DIFF
--- a/packages/paper-qa-pypdf/src/paperqa_pypdf/reader.py
+++ b/packages/paper-qa-pypdf/src/paperqa_pypdf/reader.py
@@ -321,7 +321,7 @@ def parse_pdf_to_pages(  # noqa: PLR0912
                     for img_idx, img_obj in enumerate(page.images):
                         pil_image = cast("Image.Image", img_obj.image)
                         width, height = pil_image.size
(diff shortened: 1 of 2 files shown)


packages/paper-qa-pypdf/src/paperqa_pypdf/reader.py
```diff
@@ -319,7 +319,17 @@ def parse_pdf_to_pages(  # noqa: PLR0912
                     # PyPDF will blow up here with a nice message
                     media_list = []
                     for img_idx, img_obj in enumerate(page.images):
-                        width, height = cast("Image.Image", img_obj.image).size
+                        pil_image = cast("Image.Image", img_obj.image)
+                        width, height = pil_image.size
+                        if img_obj.name.lower().endswith(".png"):
```
whitead: There is not a better attribute to check?

jamesbraza: You're right, we can check `pil_image.format`. Thanks for this suggestion, went with it

PR #1302, opened 2026-02-25
Fix acronym-led citation docname ingest failures

AmT42:
## Summary
Fixes ingest failures when `citation_to_docname` sees acronym/symbol-led citations (for example, `CD47/SIRPα ...`).

## Root cause
The parser only accepted TitleCase tokens (`[A-Z][a-z]+`), so some valid citations could raise and fail ingestion.

## Changes
- Keep existing preferred match: `([A-Z][a-z]+)`
- Add acronym fallback: `([A-Z0-9]{2,})`
- Add deterministic final fallback: `Doc{hexdigest(citation)[:8]}`
- Keep year extraction unchanged

## Tests
- `test_citation_to_docname_acronym_title` -> `CD47/SIRPα ..., 2022` => `CD472022`
- `test_citation_to_docname_non_text_fallback_is_deterministic` -> stable `Doc...` output

---

> [!NOTE]
> **Low Risk**
> Small, well-tested change limited to docname derivation; main risk is potential docname changes for edge-case citations that previously raised.
> 
> **Overview**
> Fixes `citation_to_docname` to avoid ingest failures when citations don’t contain a TitleCase author token. It now falls back to extracting an acronym/number token (e.g., `CD47...`) and, if that fails, generates a **deterministic** `Doc<hash8>` prefix while keeping year extraction unchanged.
> 
> Adds unit tests covering acronym-led citations and ensuring the final malformed-citation fallback is stable across calls.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6016ff53dd27ce3c3794a5e69d43ee2f7cb0af87. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,4 +1,5 @@
 Andrew White <andrew@futurehouse.org> <white.d.andrew@gmail.com>
+Ahmet Celebi <59479833+AmT42@users.noreply.github.com> At4 <59479833+AmT42@users.noreply.github.com>
 Anush008 <anushshetty90@gmail.com> Anush <anushshetty90@gmail.com>
 Dmitrii Magas <eamagea123@gmail.com> eamag
 Geemi Wellawatte <geemi@futurehouse.org> <gwellawatte@gmail.com>


tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
+from paperqa.utils import citation_to_docname
+
+
+def test_citation_to_docname_acronym_title() -> None:
+    citation = "CD47/SIRPα axis: bridging innate and adaptive immunity, 2022"
+    assert citation_to_docname(citation) == "CD472022"
+
+
+def test_citation_to_docname_non_text_fallback_is_deterministic() -> None:
+    citation = "___, n.d."
```
jamesbraza, line 14: Can you add a comment above `citation` here explaining the rationale behind `"___, n.d."`

Mainly, just state if this is a contrived input to test an edge case, or if it's actually seen in the real world

tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
...
+    first = citation_to_docname(citation)
+    second = citation_to_docname(citation)
+
+    assert first == second
```
jamesbraza:
```suggestion
    assert first == second, (
        "Expected docname creation to be repeatable when given the same input,"
        " even if when hashes are involved"
    )
```

Let's just document the rationale here

AmT42:
Hey   I added the rationale comments to the test. Let me know if this is clear or I can adjust

jamesbraza:
Hi @AmT42 yes it's LGTM, but can you get `pre-commit` CI to pass

jamesbraza:
Hi @AmT42 it's still not passing, you need to add yourself to `.mailmap` I guess

PR #1307, opened 2026-02-27
Supporting signed GCS links in `ParsedMedia`

jamesbraza:
This PR moves `ParsedMedia` to support URLs such as signed GCS links by:
- Adding a `url` field to `ParsedMedia`
- Integrating it into `Docs.aget_evidence` workflow, with tests

### DIFF
--- a/src/paperqa/types.py
+++ b/src/paperqa/types.py
@@ -602,7 +602,7 @@ class ParsedMedia(BaseModel):
     @model_validator(mode="after")
     def _check_data_xor_url(self) -> Self:
         if not self.data and not self.url:
-            raise ValueError("At least one of 'data' (non-empty) or 'url' is required.")
+            raise ValueError("Setting one of 'data' (non-empty) or 'url' is required.")
         if self.data and self.url:
             raise ValueError(
                 "Set 'data' or 'url', not both. Having both creates ambiguous"
@@ -657,14 +657,15 @@ class ParsedMedia(BaseModel):
             # We only compare bytes or URLs, we consider mixes incompatible.
             # If you need compatibility, resolve the URL to bytes or generate a URL
             return False
-        common = (
-            self.index == other.index
-            and self.text == other.text
-            and self._get_info_hashable() == other._get_info_hashable()
-        )
+        if (
+            self.index != other.index
+            or self.text != other.text
+            or self._get_info_hashable() != other._get_info_hashable()
+        ):
+            return False
         if self.data:
-            return common and self.data == other.data
-        return common and self.url == other.url
+            return self.data == other.data
+        return self.url == other.url
 
     def to_image_url(self) -> str:
         """Get a URL suitable for LLM image content.
@@ -672,7 +673,7 @@ class ParsedMedia(BaseModel):
         Returns a signed/public HTTP URL when available, otherwise falls back
         to an RFC 2397 base64 data URL built from the raw bytes.
         """
-        if self.url and not self.data:
+        if self.url:
             return self.url
         image_type = cast(str, self.info.get("suffix", "png")).removeprefix(".")
         if image_type == "jpg":  # SEE: https://stackoverflow.com/a/54488403
--- a/tests/test_paperqa.py
+++ b/tests/test_paperqa.py
@@ -1914,10 +1914,10 @@ def test_media_to_image_url(subtests: SubTests) -> None:
 
 
 def test_parsed_media_data_or_url() -> None:
-    with pytest.raises(ValidationError, match="At least one of"):
+    with pytest.raises(ValidationError, match="one of"):
         ParsedMedia(index=0, data=b"")
 
-    with pytest.raises(ValidationError, match="At least one of"):
+    with pytest.raises(ValidationError, match="one of"):
         ParsedMedia(index=0)
 
     with pytest.raises(ValidationError, match="not both"):


src/paperqa/types.py
```diff
@@ -625,26 +653,59 @@ def to_id(self) -> UUID:
...
+            return self.url
         image_type = cast(str, self.info.get("suffix", "png")).removeprefix(".")
         if image_type == "jpg":  # SEE: https://stackoverflow.com/a/54488403
             image_type = "jpeg"
         return encode_image_as_url(image_type, self.data)
 
     def save(self, path: str | os.PathLike) -> None:
         """Save the image to the input file path."""
+        if not self.data:
+            raise ValueError("There's no need to save media with image data as a URL.")
```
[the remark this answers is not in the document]
jamesbraza: Sure, adjusted the wording and added an explanatory comment

src/paperqa/core.py
```diff
@@ -254,13 +254,13 @@ async def _map_fxn_summary(  # noqa: PLR0912
...
-                        Message.create_message(
-                            text=message_prompt,
-                            images=(
-                                [i.to_image_url() for i in unique_media]
-                                if unique_media
-                                else None
-                            ),
+                        (
+                            create_multimodal_message(
+                                message_prompt,
```
jabra: nit: can you add explicit function parameters for readability

src/paperqa/types.py
```diff
@@ -575,7 +575,12 @@ class ParsedMedia(BaseModel):
         PlainSerializer(bytes_to_string),
         BeforeValidator(lambda x: x if isinstance(x, bytes) else string_to_bytes(x)),
     ] = Field(
-        min_length=1, description="Raw image, ideally directly savable to a PNG image."
+        default=b"",
```
jabra, line 578: this forces truthiness checks everywhere, you might want to keep it similar to url with data: bytes | None = None

src/paperqa/types.py
```diff
@@ -594,17 +599,40 @@ class ParsedMedia(BaseModel):
         ),
     )
 
+    @model_validator(mode="after")
+    def _check_data_xor_url(self) -> Self:
+        if not self.data and not self.url:
```
jabra, line 604: if you data is nullable above that you can check nullability instead of truthiness
if (self.data is None) == (self.url is None)

src/paperqa/types.py
```diff
@@ -594,17 +599,40 @@ class ParsedMedia(BaseModel):
...
+
     def _get_info_hashable(self) -> Hashable:
         if info_hashable := self.info.get("info_hashable"):
             return cast(Hashable, info_hashable)
         # We know info_hashable_hash key isn't present, so no need to filter it
         return json.dumps(self.info, sort_keys=True)
 
     def __hash__(self) -> int:
-        return hash((self.index, self.data, self.text, self._get_info_hashable()))
+        data_or_url = self.data or self.url
```
jabra, line 622: same here

src/paperqa/types.py
```diff
@@ -625,26 +653,65 @@ def to_id(self) -> UUID:
     def __eq__(self, other) -> bool:
```
jabra, line 653: same and can simplify and be more explicit

```
def __eq__(self, other: object) -> bool:
    if not isinstance(other, ParsedMedia):
        return NotImplemented

    # exactly one of data/url must be set (XOR invariant)
    if (self.data is None) != (other.data is None):
        return False

    if (
        self.index != other.index
        or self.text != other.text
        or self._get_info_hashable() != other._get_info_hashable()
    ):
        return False

    if self.data is not None:
        return self.data == other.data

    return self.url == other.url
```

src/paperqa/types.py
```diff
@@ -625,26 +653,65 @@ def to_id(self) -> UUID:
...
+        return common and self.url == other.url
 
     def to_image_url(self) -> str:
-        """Convert the image data to an RFC 2397 data URL format."""
+        """Get a URL suitable for LLM image content.
+
+        Returns a signed/public HTTP URL when available, otherwise falls back
+        to an RFC 2397 base64 data URL built from the raw bytes.
+        """
+        if self.url and not self.data:
```
jabra: you are guarantee to have only one because of the validator 
you can simplify to 
```
if self.url is not None:
    return self.url

if self.data is None:
    raise RuntimeError("ParsedMedia invariant broken: no data or url.")
```

jabra (commented):
left a comments

src/paperqa/core.py
jamesbraza: Yeah just went with it

src/paperqa/types.py
jamesbraza: Imo a truthiness check is not much different than a nullability check.

Going with `bytes | None` creates typing statefulness I don't want to introduce. I actually piloted this just now and what happens is all over the code base you see type errors that you have to either `cast` or check for

src/paperqa/types.py
jamesbraza: Per my above comment, adding nullability added too much headache

src/paperqa/types.py
jamesbraza: Even with nullability `self.data or self.url` can still work

src/paperqa/types.py
jamesbraza: Sure nice suggestion, went with it just now. I agree having a `common` local was bad

src/paperqa/types.py
jamesbraza: Yeah good point, went with a variant of this just now

PR #1333, opened 2026-06-04
Add NemotronParseBBox.to_original_coordinates for raw nemotron-parse output

jamesbraza:
## What

`nemotron-parse` emits bounding boxes normalized to its letterboxed **2048×1648** canvas (aspect-preserving resize + centered white padding). NVIDIA's hosted NIM maps these back to the input image **server-side** — via [`postprocessing.py::transform_bbox_to_original`](https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/postprocessing.py) — before returning, so PaperQA's existing decode (`coord × input_dims`) is correct against NIM. Raw vLLM / Hugging Face output is **not** transformed, so a consumer of raw model output draws every box shifted and compressed — worst at low DPI, where the canvas padding is largest.

This ports that inverse transform so raw-output consumers can recover original-image coordinates:

- `NEMOTRON_PARSE_TARGET_HEIGHT = 2048`, `NEMOTRON_PARSE_TARGET_WIDTH = 1648` — sourced (with a link) from the model's [`preprocessor_config.json`](https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/preprocessor_config.json)
- `transform_canvas_bbox_to_original(...)` module function
- `NemotronParseBBox.to_original_coordinates(height, width)` method, mirroring the existing `to_page_coordinates`

## Why it lives here

Our self-hosted vLLM Modal proxy returns raw canvas coordinates rather than replicating NIM's server-side mapping; this is the canonical, tested transform it will call. PaperQA's NIM decode path is intentionally unchanged — applying this on NIM coordinates would double-transform.

## Tests

`test_api.py` adds:
- **Parity** against NVIDIA's reference `transform_bbox_to_original` across the no-resize, downsized, and width-resized regimes
- **Round-trip** recovery (forward-letterbox a page rectangle → invert → recover, 0 px error)
- **Low-DPI** heavy-padding case, where the correction is large vs. naive page scaling

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/paper-qa-nemotron/src/paperqa_nemotron/api.py
+++ b/packages/paper-qa-nemotron/src/paperqa_nemotron/api.py
@@ -91,6 +91,69 @@ NemotronParseToolName: TypeAlias = Literal[
     "markdown_bbox", "markdown_no_bbox", "detection_only"
 ]
 
+# nemotron-parse resizes each input image (preserving aspect ratio) to fit within this
+# fixed height x width canvas, then center-pads the remainder with white, and emits
+# bounding boxes normalized to that padded canvas. These are the model's native input
+# dimensions; see "size" / "final_size" in:
+# https://huggingface.co/nvidia/NVIDIA-Nemotron-Parse-v1.1/blob/main/preprocessor_config.json
+NEMOTRON_PARSE_TARGET_HEIGHT = 2048  # px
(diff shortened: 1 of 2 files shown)


jamesbraza:
Claude why

PR #1337, opened 2026-06-12
Remove duplicated "words" in evidence-summary JSON system prompts

chris234567:
The JSON evidence-summary system prompts append a literal "words" after the {summary_length} placeholder, but {summary_length} is filled from AnswerSettings.evidence_summary_length, whose default is "about 100 words". With default settings the rendered system prompt therefore reads "...relevant information from the text - about 100 words words.".

The non-JSON summary_prompt already avoids this (it writes "Relevant Information Summary ({summary_length}):" without appending "words").

Drop the trailing literal " words" in both summary_json_system_prompt and summary_json_multimodal_system_prompt so the placeholder supplies its own units. Cosmetic prompt cleanup; no schema/API change.


whitead (approved):
Nice

PR #1343, opened 2026-07-14
bring in line with llm_config changes in LDP

agolajko:
PaperQA's existing configuration paths do not consistently reach LMI's typed model chain after the fhlmi/ldp 1.0 migration. This change accepts ordered `models` configurations, preserves legacy `model_list` configurations, and forwards the configured chain to LDP agents so per-model Responses API flags, retries, timeouts, and fallbacks take effect.

- Route answer, summary, agent, and enrichment model construction through the same configuration helper; reject explicitly empty model chains through LMI validation.
- Forward `agent.agent_llm_config` to SimpleAgent, ReActAgent, and MemoryAgent. Precedence is explicit `agent_config["llm_config"]`, legacy `agent_config["llm_model"]`, then the normal agent model configuration/defaults. Repeated construction preserves settings.
- Handle LMI's wrapped provider errors without silently dropping evidence after exhausted rate limits, server errors, or authentication failures.
- Refresh the affected recordings and retain one OpenRouter-backed DeepSeek reasoning test with legacy-config coverage.

### Configuration example

```python
from paperqa import Settings

settings = Settings(
    agent={
        "agent_type": "ldp.agent.SimpleAgent",
        "agent_llm": "gpt-4o-mini",
        "agent_llm_config": {
            "models": [
                {
                    "name": "gpt-4o-mini",
                    "responses_api": True,
                    "max_retries": 1,
                    "timeout": 60,
                    "extra_params": {"temperature": 0.0},
                },
                {
                    "name": "gpt-4o",
                    "extra_params": {"temperature": 0.0},
                },
            ]
        },
    }
)
```

Pass these settings to `agent_query`. The LDP agent first uses the Responses API for `gpt-4o-mini`; failures eligible under LMI's fallback policy advance to `gpt-4o` via Chat Completions. The same `models` shape is accepted by `llm_config`, `summary_llm_config`, and `parsing.enrichment_llm_config`.

Top-level `rate_limit` and `request_limit` are not part of typed `LLMConfig` and are not forwarded into LDP's internally constructed models by this change.

### Upstream follow-up

[Future-House/ldp#482](https://github.com/Future-House/ldp/pull/482) fixes direct `LiteLLMModel.select_tool` behavior. The LDP-agent example above uses the normal LMI call path and does not depend on that fix. PaperQA's local Aviary ToolSelector/fake-agent workaround still dispatches only the primary model through Chat Completions; removing it and raising the fhlmi minimum version remain follow-up work after the upstream fix is released.

### Validation

Focused local tests cover all three LDP agent types, explicit/legacy override precedence, settings immutability, rejection of empty chains through all four model getters, and provider-mocked Responses tool selection with fallback to Chat Completions. Related reasoning, lifecycle, and PDF-metadata recordings are checked in playback-only mode with dummy keys. Repository hooks include formatting, lint, and mypy. No live provider calls are needed for these checks.

Tracking: https://linear.app/edisonscientific/issue/AAI-440/update-paperqa-server-to-support-the-new-lmi-config


tests/test_paperqa.py
```diff
@@ -1389,6 +1390,23 @@ async def test_custom_llm_custom_media(stub_data_dir: Path) -> None:
     class StubLLMModel(LLMModel):
```
sidnarayanan: Maybe this could inherit from `LiteLLMModel` and avoid the DIY fallback, but not entirely sure

agolajko: Solved

